PR #3, opened 2017-02-09
Initial libretro port

fr500:
this should be not be too intrusive. I still have to hook up input so it's open mostly for feedback.


Nikolaj64:
https://gist.github.com/Nikolaj64/5cdc9d8c7372a7239ab4f29bbc155d20

Implemented actual input support (no to uch though) and changed the color depth to reflect the changes made in the latest commits of melonDS.  
The code could probably be a lot better, but it works.

Would have made a pull request on libretro/melonDS, but IDK how?

fr500:
@Nikolaj64 thanks a lot!, you can just fork from libretro but I'll integrate your input changes

PR #5, opened 2017-02-11
Allow passing ROM file via CLI argument

fredericmeyer:
Added a small piece of code so that ROM path can optionally be passed as CLI parameter (or via drag and drop of course)


StapleButter:
UI overhaul made this obsolete, so uh

PR #14, opened 2017-04-02
add CMake build system

s1nka:
just for fun add cmake for linux

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ bin
 obj
 *.depend
 *.layout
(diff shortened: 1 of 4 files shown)


StapleButter:
Thanks. Would it also work under Windows, though?

s1nka:
I try add windows-build for cmake. But windows-user can use CodeBlocks.

StapleButter:
I'll take inspiration from this for a cross-platform build system, probably.

Issue #39, opened 2017-05-14
Zubo (EUR) can't move around at all


SilasLaspada:
You can't move in Zubo (EUR).  The character is supposed to move to where the touchscreen is clicked, but it just turns in that direction and doesn't actually move. The console is spammed with "!! PROJ MATRIX STACK UNDERFLOW" followed by "unknown GPU read16 04001010" and "unknown GPU read16 04001012".

SilasLaspada:
I found the source of this issue, and it is really interesting. It turns out that this game let's you blow away stuff like cobwebs using the microphone. It lets you do this at any time. Now, the real problem is that you can't move while blowing. From what I can tell (please correct me if I'm wrong) MelonDS doesn't actually utilize a microphone yet, so it just tells the games that the mic is used constantly (to let you progress through puzzles, I assume). This means that in Zubo, it thinks that you're blowing into the mic the whole time, and thus doesn't let you move. I assume this can't be properly fixed until MelonDS properly picks up sounds from a microphone. But hey, at least we know why it doesn't work! :P

StapleButter:
https://github.com/StapleButter/melonDS/blob/master/src/SPI.cpp#L411

microphone reads return 0xFFF, which might make the game think that there's noise (I guess the silence level is supposed to be around 0x800).

good find there.

Brian151:
Maybe you could implement this is a way similar to desmume. I forget all the specifics, as I only ever had to use it once, but Desmume has an auto-mic system that can be turned on/off, and has a few different configurations.

SilasLaspada:
I can confirm that changing that value to 0x800 fixes the issue. Of course that means that you can't do anything that needs a microphone, but it's a start. Do you think you could make a hotkey to switch between on/off for the mic like Brian151 said? That would probably the simplest solution that isn't a hack.

SilasLaspada:
"Fixed" as of latest commit. You can't do any puzzles that require a mic, but at least you can move around.

PR #44, opened 2017-05-31
Basic microphone on/off toggle

SilasLaspada:
This is an extremely basic option that works from the emulation menu and the config file. It toggles the microphone read from 0x800 (silence) to 0xFFF (enough noise for games that need the mic to work). It's basically copied-pasted and edited from the code for the "Threaded 3D renderer" option with an if-else statement. Fixes issue #39.

### DIFF
--- a/src/Config.cpp
+++ b/src/Config.cpp
@@ -83,7 +83,7 @@ ConfigEntry ConfigFile[] =
 
     {"Threaded3D", 0, &Threaded3D, 1, NULL, 0},
 	
-	{"Microphone", 0, &Microphone, 1, NULL, 0},
+    {"Microphone", 0, &Microphone, 1, NULL, 0},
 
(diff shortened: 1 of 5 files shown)


StapleButter:
Well, thanks for the help. However, I'm not sure whether I'll take this. Two main points:

* If I were to do without proper microphone support, I'd rather have a key to feed noise to the microphone when pressed. That would do for most microphone-enabled games.

* In that switch statement in SPI.cpp, the default case is for all TSC channels that aren't touchscreen X and Y (0x50 and 0x10 resp.). The microphone channel is 0x60. Other channels shouldn't return microphone input.

SilasLaspada:
Ok, I understand. I was also thinking that a key press would be much more convenient. I could probably fix the issue with the channels, but I don't know how to do the hotkeys, and it would probably be better to put in support for a real mic anyway.

PR #103, opened 2017-09-08
Fix const correctness warnings

xfix:


StapleButter:
would merge it, but now it will have to be done separately

PR #104, opened 2017-09-09
Support touch screen input (like, using touch screen for touch screen)

xfix:
This allows for using an actual touch screen to input on touch screen. I may be the only person who cares about that actually, but oh well. Tested on a Linux tablet, but I think it should work on Windows tablets (didn't test however due to lack of hardware).

Could be cleaner, pretty sure about that, it's just that programming directly on a tablet even with connected keyboard is annoying.

### DIFF
--- a/src/wx/main.cpp
+++ b/src/wx/main.cpp
@@ -622,6 +622,31 @@ void EmuThread::ProcessEvents()
             }
             break;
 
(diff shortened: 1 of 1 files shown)


xfix:
Seems like it started to work (emulator stopped hijacking touch events) in current master.

PR #215, opened 2017-12-07
Implemented "Blow into microphone" key.

evg-zhabotinsky:
Mic reads max. volume white noise while that key is held down.
Should work for games that just require you to blow into microphone.

Handled the same as all other 12 keys up until `NDS::SetKeyMask()`, where it sets `NDS::MicBlowing` flag, which overrides values read from mic (0x800 as of now) with random noise (values 0x000 to 0xFFF) when set.

Only tested with "Phoenix Wright - Ace Attorney" (blowing on fingerprinting powder).

### DIFF
--- a/src/Config.cpp
+++ b/src/Config.cpp
@@ -25,8 +25,8 @@
 namespace Config
 {
 
-int KeyMapping[12];
-int JoyMapping[12];
+int KeyMapping[13];
+int JoyMapping[13];
 
 int WindowWidth;
 int WindowHeight;
@@ -67,6 +67,7 @@ ConfigEntry ConfigFile[] =
     {"Key_L",      0, &KeyMapping[9],   42, NULL, 0},
     {"Key_X",      0, &KeyMapping[10],  17, NULL, 0},
     {"Key_Y",      0, &KeyMapping[11],  30, NULL, 0},
+    {"Key_Mic_blow", 0, &KeyMapping[12],  41, NULL, 0},
 
     {"Joy_A",      0, &JoyMapping[0],  -1, NULL, 0},
     {"Joy_B",      0, &JoyMapping[1],  -1, NULL, 0},
@@ -80,6 +81,7 @@ ConfigEntry ConfigFile[] =
     {"Joy_L",      0, &JoyMapping[9],  -1, NULL, 0},
     {"Joy_X",      0, &JoyMapping[10], -1, NULL, 0},
     {"Joy_Y",      0, &JoyMapping[11], -1, NULL, 0},
+    {"Joy_Mic_blow", 0, &JoyMapping[12], -1, NULL, 0},
 
     {"WindowWidth",  0, &WindowWidth,  256, NULL, 0},
     {"WindowHeight", 0, &WindowHeight, 384, NULL, 0},
--- a/src/Config.h
+++ b/src/Config.h
@@ -27,8 +27,8 @@ namespace Config
 void Load();
 void Save();
 
-extern int KeyMapping[12];
-extern int JoyMapping[12];
+extern int KeyMapping[13];
+extern int JoyMapping[13];
 
 extern int WindowWidth;
 extern int WindowHeight;
--- a/src/NDS.cpp
+++ b/src/NDS.cpp
@@ -108,6 +108,8 @@ u32 KeyInput;
 u16 KeyCnt;
 u16 RCnt;
 
+bool MicBlowing;
+
 bool Running;
 
 
@@ -522,6 +524,8 @@ void SetKeyMask(u32 mask)
 
     KeyInput &= 0xFFFCFC00;
     KeyInput |= key_lo | (key_hi << 16);
+
+    MicBlowing = !(mask & 0x1000);
 }
 
 
--- a/src/NDS.h
+++ b/src/NDS.h
@@ -103,6 +103,8 @@ extern u8 ARM7BIOS[0x4000];
 
 extern u8 MainRAM[0x400000];
 
+extern bool MicBlowing;
+
 bool Init();
 void DeInit();
 void Reset();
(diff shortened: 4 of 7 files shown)


hyarsan:
What's wrong with adding actual microphone support?

evg-zhabotinsky:
* The two features are not mutually exclusive, and neither covers all use cases of the other.
* One might want to be discreet, and for most games just noise is enough, so using a button for that makes sense. That's what I wanted, and thus what I implemented.
* Implementing actual microphone support would require more SDL knowledge than I have now, copy-pasting from surrounding code would not be enough. Also, that's a feature I'll hardly ever use, wouldn't want to annoy all my neighbours.

hyarsan:
I wasn't saying **you** had to do it... I thought melonDS was supposed to be accuracy based?

evg-zhabotinsky:
I stand by my assertion that this feature is useful, and not only as a stopgap.
Also, it does not affect realism or precision of simulation unless you press the button.

And if you were simply wondering what the technical challenges are: 1. find somebody who would implement the feature; 2. I suspect the audio codec has more complex logic than a single 12-bit unsigned input from the microphone, even if hardly any games care.

BreadFish64:
@hyarsan it's also open source, so if he doesn't want to add a feature, he's not obligated to do so

hyarsan:
@BreadFish64 I never told anyone to implement it, I just wanted to know why he didn't mention anything about working microphone? I also wanted to know what your thoughts were...

LogPod:
Umm wut just happened

hyarsan:
@LogPod He deleted his post, don't worry.

Brian151:
this really needs to be merged-in... 

while the number is small, there are a handful of games that eventually become unplayable due to lack of this feature.

There isn't much justifiable reason for adding actual mic support. Also, it is entirely possible to not have a mic on the computer, itself. As already mentioned, that's a lot of extra, probably decently complicated code for not a whole lot of gain. Accuracy is nice and all, but something like this shouldn't be held-back for any reason, and that includes accuracy. That especially holds true if there are games that aren't working partially (or basically entirely) because they require the mic. And in all honesty, it would be fairly easy to implement support for both random noise and actual mic input. Basically just an if or switch statement or two.

hyarsan:
I don't see what's wrong with passing sound files into the DS mic :p

Brian151:
does work, and some games require it.

StapleButter:
This would do as a quick hack, but I need to seriously consider all this.

* we need a proper 'hotkey' system for this kind of things, where we will eventually add things like 'close lid', 'toggle fullscreen' or whatever
* I want to eventually add proper mic support, too

Before implementing any hotkey system, I'd need to fix the input dialog so it doesn't lag like hell.

The issue with actual mic support is that mic input on the DS is as bare-metal as you can get. There is no hardware sampling or buffering or anything. Reading mic input from the TSC basically returns the current mic voltage. Games can then choose to sample it at whatever rate is convenient for them, by setting up a timer.

This is a problem when modern audio APIs require you to select a sampling rate and do all the work, periodically providing you a chunk of sampled audio. Doing it right would require reading microphone input at a steady rate, and resampling based on the amount of cycles between TSC reads. There would inevitably be latency, which could be lowered at the expense of performance, so a compromise would have to be done (kind of the same as with audio output, actually).

hyarsan:
See? melonDS is all about accuracy, so mic would be a good idea.

evg-zhabotinsky:
1. So you mean to say that Mic hardware interface is just that single 12-bit SPI register? That would make things a lot easier than I thought. API documentation for DS confused me. As for resampling, I don't think that DS mic takes intervals between reads from this register into account, so you can ignore them too and simply return a sample from, say, one and a half buffer length of real time ago while keeping 3 last whole buffers of input.
2. Current input system isn't final? Well, it could use a simple method of adding new gamepad+keyboard mappings to small integers, but ultimately all the keys that go to virtual hardware will still share the same path anyway. So until this specific hotkey has a Mic input thread to go to, I see no reason to separate it from "normal" virtual hardware keys. Also, it's never late to revert bits of commit when a better solution is finally implemented, and reworked input system will make it even easier.

Brian151:
@hyarsan 
There are legit people without mics, or who can't get the mics to work, and I'd be one of them.
There also are people who simply cannot be shouting stuff into mics for one reason or another, so using the mic basically excludes them from playing some games (although the devs were foolish to do this in the first place, I've heard enough that LEGIT hardware doesn't always get it right)

It should be optional to use an actual mic.

hyarsan:
@Brian151 I never said you had to use a mic, just use a sound file.

Brian151:
@hyarsan 
alright
But the way you wrote your recent comment it kinda implies that

poudink:
@hyarsan late I know but adding the option to use a key doesn't decrease the accuracy in any way. you aren't forced to use it and let's be real, if you had the choice between pressing a key and feeding a sound file to the program each time you want to use the mic function, which one would you choose.

hyarsan:
I see what you mean, fine

hyarsan:
What is with the 3 dislikes

Brian151:
disagreement was felt. what i can't believe, though is just how much disagreement there is around such a simple topic...

lungshenli:
am I stupid or is this not in melonDS as of right now?

Brian151:
last i checked, its definitey not @lungshenli

StapleButter:
well I guess we can close that. I'm sorry that it didn't make it in, feels like I'm shitting on external help and makes me feel a bit bad.

however now we have a proper microphone input system, supporting actual mic input, WAV files, and white noise (basically what you coded). we also have a proper hotkey system instead of hacking into the keypad input system.

evg-zhabotinsky:
> feels like I'm shitting on external help and makes me feel a bit bad.

Not a useful attitude. Since the proper mic and hotkey support is here now, it only makes sense to discard my hack in favor of a cleaner approach that uses the now existing infrastructure. I've tested it just now, seems to be working like a charm!

I don't quite understand why the keypad and the hotkey system have to be separate, though. If you think about it, the lid switch is an actual hardware button! If I guessed correctly that the hotkey system is meant for easy addition of new input keys, why not use it for the keypad too? The less code gets the job done, the better!

PR #230, opened 2018-02-22
Add linux libpcap names to the pcap library list

dogtopus:
Not sure why they aren't there already, but add them won't hurt ;)


StapleButter:
yeah, sure

PR #245, opened 2018-05-08
Add searching for SDL2 via pkg-config, fixes build on Arch Linux

hcorion:
For some strange reason melonDS stopped building on Arch Linux for me, with SDL2 undefined reference errors.


hcorion:
Closing this as the issue was fixed in #248

PR #248, opened 2018-05-23
Two fixes

Dirbaio:


hcorion:
The Fix build on Arch Linux commit is already covered in #245

StapleButter:
I like this, so let's do it

PR #254, opened 2018-10-08
fix some compilation errors

khang06:
now it actually compiles for me

also went and fixed some warnings


MojoJojoDojo:
@khang06 Can you please link a compiled version for windows?

StapleButter:
well I think this has no more reason to be

thank you tho, I appreciate the intent behind this, and sorry that it didn't make it in (but I was working on that anyway)

PR #269, opened 2018-11-05
Add manifest for building as a flatpak

cpba:
It's a bit rough but it includes a desktop file and everything needed 
for the emulator to run.
It enables pulseaudio for sound, network for wifi emulation, device=all 
for gamepad support and filesystem=home to enable the emulator to write 
the save files to the same folder as the rom.

Because melonds uses sdl2 and gtk3, it should run on wayland, but I'm 
not including the socket because I haven't personally tested it to work.

### DIFF
--- a/flatpak/net.kuribo64.melonds.desktop
+++ b/flatpak/net.kuribo64.melonds.desktop
@@ -1,0 +1,8 @@
+[Desktop Entry]
+Name=melonDS
+Comment=Nintendo DS emulator
+Exec=melonDS
+Type=Application
+Categories=Game;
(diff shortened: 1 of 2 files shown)


cpba:
My brother had some problems running the precompiled version on his old laptop, so I made a flatpak to help him get around them. I figured somebody else could find it useful so I thought I'd send a pull request.

I realise it's a bit rough, but if you decide you want to publish melonds on flathub or something similar I can work on touching it up a bit to get it accepted.

StapleButter:
sure, why not

PR #285, opened 2018-11-21
Add option to limit framerate from UI

abcdjdj:
Signed-off-by: Madhav Kanbur <abcdjdj@gmail.com>


abcdjdj:
@StapleButter I think it would be great to expose this option in the ui.

StapleButter:
'ts a good idea

we can also have a 'turbo' hotkey that controls it when pressed, I guess. we'll see later.

PR #357, opened 2019-01-06
Fixed checking for build type

John-Gee:
Hello,

I may be wrong but on my machine with cmake 3.13.2 the CMakeList did not recognize -DCMAKE_BUILD_TYPE=Debug and it seems to be because EQUAL is for numbers and STREQUAL is for strings.

I also refactored it into an elseif, because it looked weird to me but I can undo that of course if not to your taste.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -183,10 +183,9 @@ if(NOT CMAKE_BUILD_TYPE)
   set(CMAKE_BUILD_TYPE Release)
 endif()
 
-if(CMAKE_BUILD_TYPE EQUAL Debug)
+if(CMAKE_BUILD_TYPE STREQUAL Debug)
 set(CMAKE_CXX_FLAGS "-g")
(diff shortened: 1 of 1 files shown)


abcdjdj:
I just realized that before this commit, none of the compiler optimizations were actually getting enabled. This commit gave me a nice performance boost. I'm sure that a lot of other people using low end hardware will notice a difference in performance with this commit. Thanks a ton for spotting this :+1:

John-Gee:
You're very welcome!

You may still be able to get more performance by tweaking the CXX flags for your machine, maybe with march=native or something like that.

abcdjdj:
Yeah, I do use -march=native. I found the Gentoo wiki to be very informative as far as optimizations are concerned. I'll just leave the link over here in case anyone reading this thread is interested - https://wiki.gentoo.org/wiki/GCC_optimization

PR #362, opened 2019-01-08
Use events for checking buttons

John-Gee:
So I don't know why it does not work without events, like you I'd expect update to do its job, but somehow it doesn't.
I've looked a bit around and it seems events is the way people go about, no idea of the pros/cons though.

I've tested this in the input menu and it allows me to set the buttons no problem.

This is just a prototype of using events to see how you like it.
If you want to go that way, it probably would require a bit more tidying.

### DIFF
--- a/src/libui_sdl/DlgInputConfig.cpp
+++ b/src/libui_sdl/DlgInputConfig.cpp
@@ -188,19 +188,18 @@ Uint32 JoyPoll(Uint32 interval, void* param)
     int id = dlg->pollid & 0xFF;
     if (id > 12) return 0;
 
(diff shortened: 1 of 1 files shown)


Arisotura:
this has inspired me another idea tho: when beginning the mapping process, ignore inputs that were pressed/whatever at the time the button is clicked, only sensing inputs that go from 'not pressed' to 'pressed' after the button click.

Arisotura:
if that doesn't do the job, I don't know what else I can do, other than blaming a shitty driver

John-Gee:
That's kind of what this does, since it only takes new presses.

Issue #203, opened 2017-10-28
Zip roms support


Catarax:
A support to roms zipped would be great, thanks!

Felipefpl:
not only .zip but also .rar and .7z that is more popular than rar.

WaluigiWare64:
melonDS supports reading numerous archive formats in latest master.

PR #364, opened 2019-01-12
Add zip and 7z rom support

abcdjdj:
Makes use of libarchive. Tested only on linux. Should work on windows too I guess.

Addresses #203 

Cheers!


abcdjdj:
Done. Thanks for pointing it out. Btw I did try testing it on windows but all my builds (even without this PR) crash on this [https://github.com/StapleButter/melonDS/blob/master/src/libui_sdl/libui/windows/areautil.cpp#L35](). All I did was download the latest SDL runtime binaries and add the include/linker directories in CodeBlocks. Any ideas why? @StapleButter

src/NDSCart.cpp
```diff
@@ -866,39 +868,125 @@ bool ReadROMParams(u32 gamecode, u32* params)
...
 
+bool CheckArchiveExtensions(const char *ext)
+{
+    const char *extensions[] =
+    {
+        "zip", ".7z", "lz4"
+    };
+
+    for(int i = 0; i < sizeof(extensions)/sizeof(char *); ++i)
+        if(!strcmp(ext, extensions[i]))
```
ccawley2011:
```suggestion
        if(!strcasecmp(ext, extensions[i]))
```

src/libui_sdl/main.cpp
```diff
@@ -1328,7 +1331,7 @@ void OnDropFile(uiWindow* window, char* file, void* blarg)
     char* ext = &file[strlen(file)-3];
     int prevstatus = EmuRunning;
 
-    if (!strcasecmp(ext, "nds") || !strcasecmp(ext, "srl"))
+    if (!strcasecmp(ext, "nds") || !strcasecmp(ext, "srl") || !strcasecmp(ext, "zip") || !strcasecmp(ext, ".7z"))
```
ccawley2011: It may be a good idea to use `CheckArchiveExtensions` here to simplify adding further archive formats in the future.

src/libui_sdl/main.cpp
```diff
@@ -2032,7 +2035,7 @@ int main(int argc, char** argv)
         char* file = argv[1];
         char* ext = &file[strlen(file)-3];
 
-        if (!strcasecmp(ext, "nds") || !strcasecmp(ext, "srl"))
+        if (!strcasecmp(ext, "nds") || !strcasecmp(ext, "srl") || !strcasecmp(ext, "zip") || !strcasecmp(ext, ".7z"))
```
ccawley2011: As above.

src/libui_sdl/main.cpp
abcdjdj: Done

src/libui_sdl/main.cpp
abcdjdj: Fixed

Ace4896:
Would like to test this on Windows as well, but I had a go with MinGW-w64 GCC-8.1.0 (`x86_64-win32-seh`), SDL 2.0.9 and libarchive 3.3.3. After adding the necessary DLLs to the same directory, mine just crashes a few moments after running the exe.

I can't get any useful output though, but I'm guessing it's related to the backtrace you posted (which is related to the GUI library instead).

GermanCoding:
> Um guys, can someone please help me figure out why my build keeps crashing so that I can test this PR on windows? I'm using mingw64 (x86_64-8.1.0-posix-seh-rt_v6-rev0) and SDL 2.0.9.

I'm currently running into the exact same issue, with the same backtrace in debug configuration - using current master branch as codebase. So far I've played around a little and figured out that it pretty much crashes everywhere when interacting with window stuff in user32.dll. No idea why tho.

GermanCoding:
I finally got it to work by switching compilers. Moving from the MinGW-w64 project to the mingw-64 compiler from MSYS2 (http://www.msys2.org/) fixed issues with crashing executables. Reverse-engineering the official melonDS release, it seems that @Arisotura also uses MSYS2. Apparently they have some fixes, or do things different (their MinGW build is also newer than the other project, currently 8.3.0).

Oh and by the way, a great tutorial on how to setup MSYS2 [with C::B] is here: https://stackoverflow.com/questions/36972968/how-can-use-mingw-w64-and-msys2-with-any-ide-like-eclipse-or-codeblocks

(I know all of this is not related at all to the pull request, I'm just posting it here because @abcdjdj and @Ace4896 were asking)

Ace4896:
Ah, thanks @GermanCoding, finally managed to compile on Windows and test this on a couple of games. All seems to be working well.

If anyone else wants to test, some additional packages needed can be installed with:

`pacman -S mingw-w64-x86_64-curl mingw-w64-x86_64-gtk3 mingw-w64-x86_64-libarchive mingw-w64-x86_64-SDL2`

abcdjdj:
Alright. LGTM, waiting for your input @Arisotura

Arisotura:
welp. sorry for having let this rot.

there are a few things to resolve there, I'll be looking into it.

* needs to use Platform::OpenFile() instead of melon_fopen() which was recently removed, but that'll be taken care of real quick
* I'm unsure about calling libarchive from the core, that doesn't seem good for portability. A better way would be coding archive support into Platform::OpenFile() or into a similar alternate function. But maybe this will need wrapping the whole file API and not just fopen().

Arisotura:
nevermind the first part

however, does this just pick the first .nds file it finds in the archive? seems so. that certainly works as long as you archive ROMs individually, may be problematic for fullset people tho.

I might want to look into how other emulators do archive support

Ace4896:
I checked how Desmume handles it (at least with 7z archives, but it's probably the same for the other types), and it shows a list of .nds files if there's more than one. If it couldn't find any .nds files, all files inside are listed instead.

In this archive that I tried, I had two DS games, one in the root of the archive and the other in the "Test" subdirectory inside the archive:

DeSmuME_0 9 11_x64_bGaOYCaPZd

Maybe something like this would work?

Arisotura:
where to put savefiles/savestates/etc is another issue, yeah

abcdjdj:
So to summarize it : 
1. Move libarchive stuff from core to platform
2. Display all nds files in the zip and let the user pick the file
3. Set SRamPath after picking nds file from zip

Hope I didn't miss anything. I've got a shit load of exams coming right up but I'll try to work on this pr in my free time. Cheers :D

abcdjdj:
@Arisotura Is it okay to use a Combobox for displaying the list of roms in the archive? I tried using a Checkbox, but for long lists we'll need to use some kind of scrollview which I can't figure out (really wish that libui had better documentation).

Rom Picker 1
Rom Picker 2

Arisotura:
I'd need to add some support for modal dialogs in libui anyway

and imo this should be a simple list or a listview with file details, that'd be the way to go


anyway, I had an idea regarding this. I'd feel bad about having your work be vain, so

I'll need to rewrite melonDS's ROM loading system, so it uses a flexible abstraction layer instead of directly using fopen/fread/etc. and so that we can derive proper paths for savefiles, savestates, etc

system on top of which you could base your work, so we can have it be self-contained and just exclude it from compilation if its dependencies aren't available on the target platform.

abcdjdj:
Sounds great! Please don't feel bad, I'm sure that we can still re-use the libarchive code at the very least. So I guess I'll just go ahead and close this PR for now. Have a great day!

Arisotura:
so hey hey

I'm thinking of the design for this

so let me know about your thughts and all

basically:

* have a ROMSource class defining the base for a ROM source (open/close, read...)
* have said ROMSource class also provide methods for deriving paths for save files, etc
* there will be a class like ROMSource_FS implementing that base and just using fopen/fread/etc
* other implementations (like yours) could be sourcing from an archive file; they could be disabled via #ifdef if needed (for example if the target platform doesn't support the dependencies)
* save files and all always go to the filesystem (C:\derp\zorp.nds ->C:\derp\zorp.sav, C:\derp\archive.zip|zorp.nds -> C:\derp\zorp.sav)

I'll take your opinions on this

abcdjdj:
> so hey hey
> 
> I'm thinking of the design for this
> 
> so let me know about your thughts and all
> 
> basically:
> 
>     * have a ROMSource class defining the base for a ROM source (open/close, read...)
> 
>     * have said ROMSource class also provide methods for deriving paths for save files, etc
> 
>     * there will be a class like ROMSource_FS implementing that base and just using fopen/fread/etc
> 
>     * other implementations (like yours) could be sourcing from an archive file; they could be disabled via #ifdef if needed (for example if the target platform doesn't support the dependencies)
> 
>     * save files and all always go to the filesystem (C:\derp\zorp.nds ->C:\derp\zorp.sav, C:\derp\archive.zip|zorp.nds -> C:\derp\zorp.sav)
> 
> 
> I'll take your opinions on this

I like the idea of making archive support optional via ifdefs. Making a base class + 2 derived classes would really help cleaning up the file i/o code in NDSCart.  LGTM :+1: 

Also, how's the libui stuff coming along?

Anuskuss:
Just my 2 cents: Why bother streaming from the zip when we could just load the ROM into RAM? I think that should be the case for non-zipped ROMs as well.

abcdjdj:
We aren't streaming from the file in the existing implementation of non zipped ROMs. Neither does my libarchive code. It's all done in RAM.

Anuskuss:
@abcdjdj I read that wrong then, my apologies.

Arisotura:
coming back on this; sorry for letting it rot for so long

anyway, we're making a new Qt UI, and we have ideas for redesigning filesystem access. once we're done, you'll be able to adapt your code to this, such that there is no libarchive dependency in the core.

abcdjdj:
> coming back on this; sorry for letting it rot for so long
> 
> anyway, we're making a new Qt UI, and we have ideas for redesigning filesystem access. once we're done, you'll be able to adapt your code to this, such that there is no libarchive dependency in the core.

No worries at all. I'm looking forward to adapting this PR for the new QT frontend :)

Amnesia1000:
I don't know if it will work for you, but retroarch uses the compressed files in this way from what I have seen, once the game is chosen, it is unzipped in the same folder where it is, running normally, once the emulator is closed the game is deleted ( previously it was unzipped in a temporary folder), this use for me is great, since I have several hack or translated games with IPS files, which since I don't like to patch them, it takes them from the same folder.
A game selector is not required (assuming you have only one game per compressed file). Although I can understand that someone can have more than one game per zip.

PR #366, opened 2019-01-15
Some changes for compiling on MSVC

Ace4896:
This is my initial attempt at getting this to compile on MSVC. As of now, there's a bunch of warnings in the compile output, but at the very least, it can compile and run. Tested using the `Visual Studio 15 2017 Win64` generator, compiling with `cmake --build .` and finally copying SDL's dll file into the build directory.

Some things to change though:

- I haven't handled the different compile flags for GCC and MSVC; currently, the Debug and Release configurations use flags intended for GCC, so the MSVC compiler complains about and ignores these options. There's a bunch of warnings about using the `/EHsc` flag with the compiler as well.
- There could have been a better way to handle `pcap` includes (I ended up just including the `src` directory).
- I think I've broken MinGW due to the manifest changes (which require CMake 3.4 and appears to MSVC only). Reverting these however causes some linking errors once the program compiles. Not too sure how to handle this.


Ace4896:
I've added another commit which should get rid of the `/EHsc` errors and use the correct compiler flags. I also put the old resource file back, which should help with MinGW.

Ace4896:
I had a go at compiling with CMake + MinGW and I can't link to the precompiled SDL2 libraries. I'll probably take a look at this another time, since it doesn't seem to be related to the manifest changes I did.

Ace4896:
Pretty sure this isn't the right way to do this, if I figure something out I'll submit another PR.

PR #372, opened 2019-01-24
Add Mac Cmake Build Support 

IamRifki:
Since the support for Building on mac with the latest commit is broken, for some reason. I Macgyvered one myself, It probably isn't the best and should be modified.


IamRifki:
It seems that the code doesn't work on other machines, nvm

PR #382, opened 2019-03-01
Fixed build, some sources were forgotten in CMakeLists.

baconwaifu:
Build broke due to linking errors, not being able to find `LAN::DeInit()` when it was very clearly there. attempting to manually point it at the object file revealed that it never got compiled, due to not being included in CMakeLists. This fixes that, as well as informing `LAN.cpp` the *proper* place for `Wifi.h`, and adding a missing include for the byte order functions. (`arpa/inet.h`)


baconwaifu:
Have also begun to push some wifi fixes for *nix systems.

Arisotura:
this was taken care of, so... yeah

PR #396, opened 2019-04-23
Add frameskip option

Hydr8gon:
As the title says. I'm probably keeping this in the Switch version, so I might as well throw a PR out there. Do with it what you will. If you want it I'm open to change whatever, like the UI implementation, for example.

### DIFF
--- a/src/Config.cpp
+++ b/src/Config.cpp
@@ -29,10 +29,12 @@ namespace Config
 const char* kConfigFile = "melonDS.ini";
 
 int Threaded3D;
+int Frameskip;
 
 ConfigEntry ConfigFile[] =
 {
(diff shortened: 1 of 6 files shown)


Arisotura:
I need to look into it

there are potential issues with frameskip on the DS, especially when video capture is used

Hydr8gon:
Yeah, it breaks stuff that uses display capture. Also it should probably be noted that there might be some synchronization issues when using threaded 3D rendering with this (I tried to skip 3D rendering too for maximum benefit, but just did it quickly).

Hydr8gon:
Just gonna close this for now, not sure if it's worth updating to work for 0.8 anyways.

PR #398, opened 2019-05-01
Revamp CMake build system

Nadia-h:
This is almost an entire rewrite of the CMake build system that
- Splits the build script up into different files to make it easier to build for example a libretro core or some other new frontend
- Uses libui's existing build system instead of needlessly doing it ourselves
- Finally builds correctly on MSYS2 on Windows with no changes
- Copies `romlist.bin` to the build directory

I've also added build instructions for Windows to the readme and written a small script to copy the needed files to run it to a directory that can be distributed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ make
 #### MSYS2 and CMake
 
 1. Install [MSYS2](https://www.msys2.org/)
-2. Open the **MSYS2 MingW 64-bit** terminal
+2. Open the **MSYS2 MinGW 64-bit** terminal
 3. Update the packages using `pacman -Syu` and reopen the terminal if it asks you to
 4. Install dependencies: `pacman -S mingw-w64-x86_64-{toolchain,SDL2,cmake} make git`
 5. Run the following commands
--- a/README.md
+++ b/README.md
@@ -55,9 +55,25 @@ make
 ### Windows:
 
  * use CodeBlocks
- * or receive golden cookies if you get Cmake to work
 
-Build system is not set in stone.
+#### MSYS2 and CMake
+
+1. Install [MSYS2](https://www.msys2.org/)
+2. Open the **MSYS2 MingW 64-bit** terminal
+3. Update the packages using `pacman -Syu` and reopen the terminal if it asks you to
+4. Install dependencies: `pacman -S mingw-w64-x86_64-{toolchain,SDL2,cmake} make git`
+5. Run the following commands
+   ```bash
+   git clone https://github.com/Arisotura/melonDS.git
+   cd melonDS
+   mkdir build
+   cd build
+   cmake .. -G "MSYS Makefiles"
+   make -j5
+   ../msys-dist.sh
+   ```
+
+If everything went well, melonDS and the libraries it needs should now be in the `dist` folder.
 
 ## TODO LIST
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,4 +32,3 @@ endif()
 configure_file(
 	${CMAKE_SOURCE_DIR}/romlist.bin
 	${CMAKE_BINARY_DIR}/romlist.bin COPYONLY)
-install(TARGETS ${PROJECT_NAME} RUNTIME DESTINATION ${CMAKE_INSTALL_PREFIX}/bin)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
-cmake_minimum_required(VERSION 3.13)
+cmake_minimum_required(VERSION 3.10.2)
 
-cmake_policy(VERSION 3.13)
+cmake_policy(VERSION 3.10.2)
 
 set(CMAKE_CXX_STANDARD 11)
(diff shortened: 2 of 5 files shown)


Ace4896:
Just tested this out on Windows with MSYS2, this really made the compilation process much easier for me (at least compared to setting up the compiler manually with CodeBlocks). The script you made for copying the DLLs was pretty handy too, saved me from searching for them myself.

I'll try this out on Linux as well just to make sure.

README.md
```diff
@@ -55,9 +55,25 @@ make
 ### Windows:
 
  * use CodeBlocks
- * or receive golden cookies if you get Cmake to work
 
-Build system is not set in stone.
+#### MSYS2 and CMake
+
+1. Install [MSYS2](https://www.msys2.org/)
+2. Open the **MSYS2 MingW 64-bit** terminal
```
Ace4896: Really minor nit, but it should be MinGW

Nadia-h:
Lowered the required version to the one included in Ubuntu 18.04 LTS, hopefully that should be good enough.
Should've probably tested on something other than Arch and MSYS2...

Ace4896:
Just tried it out on KDE Neon and tested Windows w/ MSYS2 again, seems to be OK for both of them. Haven't tried anything other than the main libui frontend though.

I did also try using the VS 2017 generator; it compiled but failed to link (libui-related things from what I remember). But then again, it didn't work in the old cmake system either, also for the same reason. I'll try to look into it if I have time, since it could just be an odd MSVC install I have.

Nadia-h:
Tried it with VS2019 and it can't find the SDL2 headers, I have no clue how to go about fixing that.

Ace4896:
It's a bit annoying to fix since there's no official `FindSDL2.cmake` available (after looking around, everyone seems to be making their own scripts to deal with this), but I did find [this repo](https://github.com/tcbrindle/sdl2-cmake-scripts) that had one under the BSD license. I had a go at trying to add this a while back and was going to make a PR about this, but I got busy and ended up forgetting... oops.

The other (temporary) option which I used to try this earlier was to make an `sdl2-config.cmake` file in the base directory where the SDL2 `lib` and `include` directories were, and have it contain the following ([source](https://trenki2.github.io/blog/2017/06/02/using-sdl2-with-cmake/)):

```cmake
set(SDL2_INCLUDE_DIR "${CMAKE_CURRENT_LIST_DIR}/include")

# Support both 32 and 64 bit builds
if (${CMAKE_SIZEOF_VOID_P} MATCHES 8)
  set(SDL2_LIBRARIES "${CMAKE_CURRENT_LIST_DIR}/lib/x64/SDL2.lib;${CMAKE_CURRENT_LIST_DIR}/lib/x64/SDL2main.lib")
else ()
  set(SDL2_LIBRARIES "${CMAKE_CURRENT_LIST_DIR}/lib/x86/SDL2.lib;${CMAKE_CURRENT_LIST_DIR}/lib/x86/SDL2main.lib")
endif ()

string(STRIP "${SDL2_LIBRARIES}" SDL2_LIBRARIES)
```

In both cases, I had to set the `SDL2_DIR` variable so cmake could use this.

Nadia-h:
I find this pretty weird though, running CMake within MSYS2 it seems it can find SDL2 and generate the build files just fine, but then Visual Studio can't find it? I have no idea what's going on here.

Ace4896:
From what I remember, MSYS2 can find it just fine since it sets some environment variables to include `~/bin` and `~/include` (maybe `~/lib` as well), where `~` is `<msys2-install-path>/mingw64` for the 64-bit version. After using `pacman`, the installed SDL2 stuff ends up in these directories.

I'll look into using [`vcpkg`](https://github.com/Microsoft/vcpkg) since it manages packages like MSYS2 and it might make it easier to build with MSVC. Maybe we don't need to make a `FindSDL2.cmake` thing after all.

Pik-a-boo:
Someone knows how to compile MelonDS with PGO?
How to add the PGO flag and stuff.

NarryG:
> From what I remember, MSYS2 can find it just fine since it sets some environment variables to include `~/bin` and `~/include` (maybe `~/lib` as well), where `~` is `<msys2-install-path>/mingw64` for the 64-bit version. After using `pacman`, the installed SDL2 stuff ends up in these directories.
> 
> I'll look into using [`vcpkg`](https://github.com/Microsoft/vcpkg) since it manages packages like MSYS2 and it might make it easier to build with MSVC. Maybe we don't need to make a `FindSDL2.cmake` thing after all.  

Did you ever manage to figure out that vcpkg stuff to get it building undert MSVC? I'm in a super edge-case situation where I want to mod some C# libraries onto Melon-DS and I built everything around C++/CLI, which means MSVC is a requirement

Ace4896:
Sorry, I didn't end up figuring it out. I've forgotten what I did since I had lots of exams, so I can't report any progress...

Now that my exams are done, I'll try again with a fresh Windows VM, though I can't guarantee if I can get it to work (C / C++ stuff isn't really my forte).

Issue #347, opened 2019-01-02
[Feature Request] Hotkey for toggling framerate limiter


Ace4896:
Would be nice to have a hotkey for this option (both keyboard and joystick).

StapleButter:
noted

Anuskuss:
I would also like these actions as hotkeys:
- Exit emulator
- Switch screen (1/2/both) and switch layout (h/v)
- Save/load state

P.S. speaking of layouts, it would be nice if melonDS would remember the layout I used on a per-game basis. Just an idea 😃

Arisotura:
ideas noted

PR #401, opened 2019-05-12
Add Fast Forward (Hold and Toggle) Hotkeys

Ace4896:
See #347 (not closing since there were other suggestions there).

Adds hotkeys for fast forward (disabling FPS limit) which are synced with the `Config -> Limit framerate` checkbox in the menu. I was only able to test the keyboard hotkeys since I don't have a controller on me right now, but I *think* at least the toggle button should work for joystick.

I've set the keyboard hold hotkey to TAB by default, but I'm not sure what to set the toggle hotkey to (right now, it's set to `-1` in the code, but it doesn't read as `None` in the hotkey configurator).

EDIT: I forgot to mention, but the hold / toggle hotkeys work independently. So if you use toggle the FPS limiter off first, the hold hotkey shouldn't do anything until you toggle the FPS back on (either by hotkey or the menu option).


Arisotura:
I need to add a way to set keys to 'none'. maybe pressing the Delete key?

autofire372:
I can't map the triggers on my DS4 (through DS4Windows) to either of these hotkeys. I guess melonDS doesn't recognize controller axes, only buttons...

Ace4896:
Not exactly ideal, but you could work around this by having your DS4 controller buttons map to keyboard keys (iirc DS4Windows had this option in a profile's settings).

I'd take try this myself but I don't have my DS4 on me...

Anuskuss:
> I need to add a way to set keys to 'none'. maybe pressing the Delete key?

No need to reopen this, but I think the `ESC` key would've been more appropriate.

PR #416, opened 2019-05-28
Linux: Fix issue where melonDS.ini is created with \ in name

### LINKED ISSUE #413: [Linux] melonDS.ini created with \ in name if no existing config is found
If I run the app via a terminal (like `./melonDS`) and change some settings, the config file is given the wrong name:

```bash
jon@jon-acer-aspire-a715-71g:~/Documents/Portable Apps/MelonDS> ./melonDS
melonDS 0.7.4
http://melonds.kuribo64.net/
jon@jon-acer-aspire-a715-71g:~/Documents/Portable Apps/MelonDS> ls -a1
.
..
bios7.bin
bios9.bin
firmware.bin
firmware.bin.bak
melonDS
.\melonDS.ini
romlist.bin
timewarp.mln
```

For context, I ran into this issue while I was attempting to bundle melonDS as an AppImage. It created the config file in the parent directory rather than the app's directory, so I decided to check what would happen normally. I'm still figuring this out though, so I'll make a PR if I manage to get it working properly.

Ace4896:
Should fix #413.

I found out that this issue wasn't because I ran it in a terminal, but whenever there was no config in any of the valid locations and it creates a new one in the emulator directory.

### DIFF
--- a/src/libui_sdl/Platform.cpp
+++ b/src/libui_sdl/Platform.cpp
@@ -167,7 +167,13 @@ FILE* OpenLocalFile(const char* path, const char* mode)
         int len = emudirlen + 1 + pathlen + 1;
         emudirpath = new char[len];
         strncpy(&emudirpath[0], EmuDirectory, emudirlen);
+
+#ifdef __WIN32__
         emudirpath[emudirlen] = '\\';
+#else
+        emudirpath[emudirlen] = '/';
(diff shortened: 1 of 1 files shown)


Ace4896:
See 27d21e06df8d24ab6c4e53ba5305af533b09f6f2.

Issue #415, opened 2019-05-28
Fails to build on msys2 (glActiveTexture/glBlendColor "not declared in this scope")


autofire372:
```
C:/msys64/home/Alec/melonDS/src/GPU3D_OpenGL.cpp: In function 'bool GPU3D::GLRenderer::Init()':
C:/msys64/home/Alec/melonDS/src/GPU3D_OpenGL.cpp:351:5: error: 'glActiveTexture' was not declared in this scope
     glActiveTexture(GL_TEXTURE0);
     ^~~~~~~~~~~~~~~
C:/msys64/home/Alec/melonDS/src/GPU3D_OpenGL.cpp:351:5: note: suggested alternative: 'glBindTexture'
     glActiveTexture(GL_TEXTURE0);
     ^~~~~~~~~~~~~~~
     glBindTexture
C:/msys64/home/Alec/melonDS/src/GPU3D_OpenGL.cpp: In function 'void GPU3D::GLRenderer::RenderSceneChunk(int, int)':
C:/msys64/home/Alec/melonDS/src/GPU3D_OpenGL.cpp:876:13: error: 'glBlendColor' was not declared in this scope
             glBlendColor((float)b/31.0, (float)g/31.0, (float)r/31.0, (float)a/31.0);
             ^~~~~~~~~~~~
C:/msys64/home/Alec/melonDS/src/GPU3D_OpenGL.cpp:876:13: note: suggested alternative: 'glClearColor'
             glBlendColor((float)b/31.0, (float)g/31.0, (float)r/31.0, (float)a/31.0);
             ^~~~~~~~~~~~
             glClearColor
C:/msys64/home/Alec/melonDS/src/GPU3D_OpenGL.cpp:885:9: error: 'glActiveTexture' was not declared in this scope
         glActiveTexture(GL_TEXTURE0);
         ^~~~~~~~~~~~~~~
C:/msys64/home/Alec/melonDS/src/GPU3D_OpenGL.cpp:885:9: note: suggested alternative: 'glBindTexture'
         glActiveTexture(GL_TEXTURE0);
         ^~~~~~~~~~~~~~~
         glBindTexture
C:/msys64/home/Alec/melonDS/src/GPU3D_OpenGL.cpp: In function 'void GPU3D::GLRenderer::RenderFrame()':
C:/msys64/home/Alec/melonDS/src/GPU3D_OpenGL.cpp:960:5: error: 'glActiveTexture' was not declared in this scope
[...]

PR #420, opened 2019-06-01
Update dependencies in README

Ace4896:
See #415.

I'm pretty sure `mesa-common-dev` is the package needed for Ubuntu, Debian etc., but if another one is needed, please let me know.

I also changed the make instruction so it uses `nproc --all` to use all available threads for compiling.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ As for the rest, the interface should be pretty straightforward. If you have a q
 * Install dependencies:
 
 ```sh
-sudo apt-get install libpcap0.8-dev libcurl4-gnutls-dev libsdl2-dev gtk+-3.0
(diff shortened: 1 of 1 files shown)


Arisotura:
I wonder, though, why is mesa-common-dev required? I haven't seemed to need it here

Ace4896:
Oh, maybe it's already installed by default or when graphics driver packages are installed? I put it there just in case it wasn't already installed, but I can remove it if you want.

Ace4896:
Ah never mind, I saw that `libsdl2-dev` installs a bunch of Mesa-related things already with all the necessary OpenGL headers, no need for `mesa-common-dev`.

Arisotura:
okie dokie

Issue #272, opened 2018-11-07
Make it compile on Windows with MSVC


KrossX:
Here's a few changes and a crappy makefile I ended up making. It works with the msvc 14.13 (VS2017) compiler I used it with.  I was gonna try and see if I could make the fullscreen mode but didn't want to get codeblocks or mingw. 

I noticed no speed difference, although I don't have a fancy CPU with AVX to try and compile it with. Less dlls though, which is nice I think.

As for the changes. The change in `Config.cpp` makes both defines point to the same version. Otherwise I would get an error due to mismatch. The change in `libui_sdl/Platform.cpp` makes it so that `PCAP_OPENFLAG_PROMISCUOUS` is defined, I don't know if it has to be a zero or one though, otherwise there would be an error since it's referenced later on the file.

[makefile_and_diff.zip](https://github.com/StapleButter/melonDS/files/2555769/makefile_and_diff.zip)

StapleButter:
I guess this will need updating due to changes in the code this affects.

PCAP_OPENFLAG_PROMISCUOUS is defined to 1.

I need to rework that code and how pcap is integrated anyway.

KrossX:
Here's an updated set. It outputs a few warnings:
```

src/Config.cpp(129): warning C4477: 'fprintf' : format string '%s' requires an argument of type 'char *', but variadic argument 2 has type 'void *'
src/NDS.cpp(1211): warning C4477: 'sprintf' : format string '%lu' requires an argument of type 'unsigned long', but variadic argument 1 has type 'u64'
src/NDS.cpp(1211): note: consider using '%llu' in the format string
src/NDS.cpp(1211): note: consider using '%Iu' in the format string
src/NDS.cpp(1211): note: consider using '%I64u' in the format string
src/NDS.cpp(1212): warning C4477: 'sprintf' : format string '%lu' requires an argument of type 'unsigned long', but variadic argument 1 has type 'u64'
src/NDS.cpp(1212): note: consider using '%llu' in the format string
src/NDS.cpp(1212): note: consider using '%Iu' in the format string
src/NDS.cpp(1212): note: consider using '%I64u' in the format string
src/NDS.cpp(3224): warning C4552: '&': result of expression not used
src/libui_sdl/libui/windows/colordialog.cpp(859): warning C4312: 'type cast': conversion from 'int' to 'HMENU' of greater size
```
Only a small change was required to the source:

```diff
[...]

PR #467, opened 2019-06-22
Compilation on Windows with MSVC makefile

KrossX:
A few changes and a simple NMAKE makefile for compilation on Windows. A better alternative to issue #272 as it's easier to see if it's outdated.

### DIFF
--- a/makefile
+++ b/makefile
@@ -1,0 +1,137 @@
+#NMAKE
+
+CC_FLAGS = /c /Ox /EHsc /GL /Dstrncasecmp=_strnicmp /Dstrcasecmp=_stricmp /DWIN32 /D__WIN32__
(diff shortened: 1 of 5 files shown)


MojoJojoDojo:
Needs detailed instructions for usage.

KrossX:
With the dev command prompt and extra includes/libs path set, you just run `nmake` in that directory.

RSDuck:
why can't cmake be used for this? What if a new file is added, now both the cmake and the nmake script need to be maintained. Preventing those situation was the reason build script generators like cmake were invented.

KrossX:
Not meant to be merged actually and I've been maintaining it, see the referenced issue. I made it for myself, as I don't like to use cmake, and shared in case it might be of use to someone else.

Issue #471, opened 2019-06-24
Closing a config menu with esc crashes on Windows


NarryG:
If you open any config menu then press esc, the entire program crashes.

Arisotura:
can't reproduce

NarryG:
Happening consistent for me on both the official build as well as my own custom build.
Here's some actual debug output. It is under my own custom build which I made under MSVC (which is why it's an image of VS), but it happens on 0.8.1 official as well.

[img](https://narry.land/0krvjD.png)

This may actually be a bug in LibUI based on what the exception says? Interestingly enough, it only crashes when I'm not attached to the debugger. When I'm on the debugger, it hits the breakpoint then it's fine afterwards.

NarryG:
How old is the libUI implementation? Peeking over on their repo, it looks like this may be an issue that was already solved.
https://github.com/andlabs/libui/issues/55#issuecomment-242757219

Arisotura:
try the latest revision?

NarryG:
Works

PR #472, opened 2019-06-25
Fix crash on esc when in menus on Windows

NarryG:
Fixes the crash mentioned in #471 by merging [ee87a9](https://github.com/andlabs/libui/commit/ee87a9db2383fd9333b3bb41644c3f06c9ae21be) from libui


Arisotura:
o, didn't notice there was a pull request for this, I already backported the fix

PR #490, opened 2019-07-14
x64 JIT recompiler

RSDuck:
feel free to request any changes.


MojoJojoDojo:
Tried this on Linux.
I got crashes in a couple of games (Pokemon BW2 and Mega Man Star Force 3) that I dont get without JIT.
I hope i compiled it correctly.

RSDuck:
I tried Pokemon White 2 myself (though admitedly only early game) and I didn't experience a crash.
Can you try compiling a debug build, there might be an assertion which failed. A save state would also be great and a log.

MojoJojoDojo:
Hmm it does not crash while in debug.
Here are save states that i created in release : 
[saves.zip](https://github.com/Arisotura/melonDS/files/3392451/saves.zip)

RSDuck:
can you reproduce the crash using the save file?

MojoJojoDojo:
I guess i compiled it wrong?
I added -DJIT_ENABLED=ON after cmake.
Sorry if i did.

RSDuck:
> I guess i compiled it wrong?
I added -DJIT_ENABLED=ON after cmake.
Sorry if i did.

You did everything right, if the JIT can be enabled in the emulator settings. The pity thing about those dynarecs is that they're quite nondeterministic, which can make reproducing bug quite challenging. Anyway, keep reporting those issues, one fix somewhere else might also fix this. I'm currently hunting down a bug in how wants to be a millionaire.

MojoJojoDojo:
Forgot to tell you how to reproduce the crashes.
On Star Force 3 it happens after i get into battle.
On Pokemon its when I exit/enter between places - from a small place like a pokemon center to a big one for example.
I will report here if i got more issues.
Thanks.

RSDuck:
now I got it reproduced, thanks

RSDuck:
I think, Pokemon is fixed now. It was quite a severe bug, hopefully this fixes Mega Man too.

MojoJojoDojo:
Thanks now everything correctly.
BTW I dont know if it's related or not but even with JIT  Mortal Kombat is still very slow (half speed - 30 fps).
It's even kinda slowish on other emulators.
I wonder if this game was just badly programmed or something.

RSDuck:
> BTW I dont know if it's related or not but even with JIT Mortal Kombat is still very slow (half speed - 30 fps).
It's even kinda slowish on other emulators.
I wonder if this game was just badly programmed or something.

yes, that game is known to cause problems.

cascardian:
Awesome work!

I have a question: Is there a reliable way to make the JIT detect the best block size automatically and continuously?

RSDuck:
> I have a question: Is there a reliable way to make the JIT detect the best block size automatically and continuously?

Unfortunately only by having some kind of database, so no. Guessing from experience with the JIT of Desmume the best performance/compatibility comes at around 8-12 instructions per block.

MojoJojoDojo:
If you boot games non-directly (from the menu)  the game would crash the second time onwards.

RSDuck:
the last commit should fix the bios menu. There's a weird thing where it takes a bit longer to load the menu when the block size is high(somewhere about >16), there's probably nothing which can be done. Afaik Desmume even automatically reduces the block size to 12 or something while executing the bios, though in this case the bios eventually even boots, so we're still better off.

MojoJojoDojo:
Now I need to see how this affects online play with console players using Altwfc/Wiimmfi.
I assume this being non-deterministic will cause desync issues when playing online.

MojoJojoDojo:
Well I played online with someone so far so good no desyncs or bugs or anything.
I hope this gonna get merged soon because i still use 0.8.1 with the branch lol.

RSDuck:
I'm glad to hear that. It'll be likely included in the next release, together with some other things which are still being worked on.

poudink:
very excited to try this out. my PC is trash and since my graphic card is AMD the hardware renderer didn't give me quite the boost I'd hoped for. maybe this will finally make things run more consistently for me.

RSDuck:
just for documentary purposes, this:

> the last commit should fix the bios menu. There's a weird thing where it takes a bit longer to load the menu when the block size is high(somewhere about >16), there's probably nothing which can be done. Afaik Desmume even automatically reduces the block size to 12 or something while executing the bios, though in this case the bios eventually even boots, so we're still better off.

turned out to be wrong. Instead it was a tiny bug sitting much deeper inside melonDS, which previously didn't come into effect, and also explains why the weird bug only happened the first time the bios is booted. It was fixed with this commit: https://github.com/Arisotura/melonDS/pull/490/commits/657e8a1123e7123df292ffd72073c3c0dfb1fcbb

PR #505, opened 2019-08-15
Add warning message if romlist.bin is not found

Zettymaster:
This will hopefully help to avoid guys like me loosing saved due to missing romlist.bin.
Will probably come in handy when someone uses melonDS later and makes a mistake.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,5 @@ obj
 melon_grc.c
 melon_grc.h
 cmake-build
+cmake-build-debug
+.idea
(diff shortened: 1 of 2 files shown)


RSDuck:
I don't want to let this sit for ever. It looks fine to me, let's merge it @Arisotura 

As a note for next time, while working in your own branch, please don't merge the changes from upstream. Instead rebase your branch, or let it be, as long as no conflicts occur. This won't have any impact, now since all of your commits are going to be squashed into one in the master branch. Still it's just cleaner that way.

Zettymaster:
Ok, will do next time.

PR #506, opened 2019-08-18
[WIP] Slot-2/GBAMP emulation

asiekierka:
Emulated cartridges:

- [x] GBA Movie Player (guesswork)
- [ ] DS Memory Expansion Pak (preliminary)
- [x] SuperCard CF (guesswork)

TODO:

- [x] add Slot-2 configuration menu
- [ ] implement DS Memory Pak emulation (Internet Browser)
- [x] implement emulation for an "ext RAM + DLDI" CompactFlash cart of choice
- [ ] expose DLDI patching as a config option? (right now it's "always on" on this branch, uses a "dldi.bin" file)


src/slot2/GBACartMemoryPak.cpp
```diff
@@ -0,0 +1,54 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include "../Platform.h"
+#include "GBACartMemoryPak.h"
+
+// TODO: This is insufficient. Testing with real hardware is necessary.
+
+GBACartMemoryPak::GBACartMemoryPak() {
+    this->Memory = (u16*) malloc(0x800000);
```
RSDuck: atleast from what I've picked up, we use ``new`` 😄

src/NDSCart.cpp
```diff
@@ -399,7 +399,7 @@ void Write(u8 val, u32 hold)
     switch (CurCmd)
     {
     case 0x00:
-        // Pok�mon carts have an IR transceiver thing, and send this
+        // Pok�mon carts have an IR transceiver thing, and send this
```
RSDuck: could you check your editor settings if something's wrong with the encoding?

src/slot2/GBACart.cpp
```diff
@@ -0,0 +1,70 @@
...
+    with melonDS. If not, see http://www.gnu.org/licenses/.
+*/
+
+#include <stdio.h>
+#include <string.h>
+#include "../Platform.h"
+#include "GBACart.h"
+#include "GBACartNone.h"
+
+#define RAM_TO_CART(addr) (((addr) >> 1) & 0x00FFFFFF)
```
RSDuck: consider using an inline function

src/slot2/GBACartSuperCardCF.h
```diff
@@ -0,0 +1,35 @@
...
+
+#include "GBACart.h"
+#include "IDEDrive.h"
+
+class GBACartSuperCardCF: public GBACart {
+public:
+    GBACartSuperCardCF(const char *drivePath);
+    ~GBACartSuperCardCF();
+    bool IsValid() { return this->Valid; }
+    virtual u16 RomReadWord(u32 addr);
```
RSDuck: 16-bit word is a bit ambigious with a 32-bit processor. Maybe rename it to just RomRead/`RomRead16`?

src/libui_sdl/main.cpp
```diff
@@ -1512,6 +1516,41 @@ void OnAreaResize(uiAreaHandler* handler, uiArea* area, int width, int height)
 }
 
 
+void LoadSlot2Cart()
```
RSDuck: I would move the cart selection into the core. Otherwise this code needs to be duplicated for each frontend

src/slot2/GBACart.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+namespace GBACartHelper {
+    static GBACart *cart = &GBACartNone::singleton();
+
+    void Use(GBACart *inputCart) {
```
RSDuck: follow up to my suggestion of moving the cart selection into the core. Then we don't need to move raw pointer through so many places of the code (frontend -> NDS.cpp -> here) and instead they stay here contained. Then we also don't need this singleton thing, since we can control the instantiation from a single place.

asiekierka:
I don't mind "pedantic" suggestions - anything to improve the code quality! All of your suggestions seem reasonable.

src/slot2/GBACart.cpp
asiekierka: The main reason it's done this way is because each cart can take different arguments (for example: a CF/SD image path, a cartridge ROM path, et cetera). Do you have a better proposal on how to handle this? I have some ideas (I'd pass a struct containing all possible configuration options), but would like to hear yours if they differ.

src/slot2/GBACart.cpp
asiekierka: I'm not sure - I always preferred using defines for this kind of simple mathematical kludgery.

src/slot2/GBACart.cpp
RSDuck: I see, the gui needs to communicate with the individual cart types. Maybe we could have a function per carttype named UseCartXYOrSo where all of the object construction and error checking is done, so only the switch, which passes the data from the control elements to the appropriate function remains in the frontend.

That all said, we still need to save the config somehow. Maybe it would be the smartest to add all the settings as config entries. Then only function which updates the cartridge configuration would be needed. It would take the config data by itself to construct the appriate object with the right settings. The gui dialog would then only have to change the right config data and call the "update" method.

asiekierka: I've gone for the struct + config entries approach, if you could have a look and let me know what you think I'd appreciate that.

RSDuck: I was a bit confused, but after going through your commit, I saw you've already implemented it. 👍 It solves all problems I mentioned, looks good to me

asiekierka:
Not getting merged in its current shape anyway; closing.

PR #508, opened 2019-08-24
Fix division edge case

xperia64:
The upper 32-bits of the quotient should not be 1 when the numerator is positive and the denominator is 0.

I have verified this on hardware.


Arisotura:
thanks!

PR #509, opened 2019-08-25
Fix cmake resource file compilation issues

Zapeth:
Should fix https://github.com/Arisotura/melonDS/issues/307 ~~and https://github.com/Arisotura/melonDS/issues/507~~
Nevermind, just saw that the output of https://github.com/Arisotura/melonDS/issues/507 should only be generated for Linux builds anyway so this shouldn't have any effect on that.

(untested with an MSVC setup, might require an additional `if(NOT MSVC)` if it causes issues there)

### DIFF
--- a/src/libui_sdl/CMakeLists.txt
+++ b/src/libui_sdl/CMakeLists.txt
@@ -14,6 +14,10 @@ SET(SOURCES_LIBUI
 	OSD.cpp
 )
 
+if (WIN32)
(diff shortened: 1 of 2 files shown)


Zapeth:
Took a second look at https://github.com/Arisotura/melonDS/issues/507 and was able to reproduce and fix it on my system, so I figured I'd include it here.

PR #522, opened 2019-09-29
fixed error when using clang

Zettymaster:
I noticed that when using clang, it fails compilation because of supplying an unsigned int to a normal int.

This is pretty much a simple case of adding a static_cast<int> to the affected parameters.


Zettymaster:
I just noticed, that when compiling for release, i get a linker error. `/src/libui_sdl/libui/liblibui-temporary.a(areaevents.c.o) in archive is not an object`

Zettymaster:
Turns out that error was just my pacman somehow not installing llvm even though clang is installed -.-

PR #527, opened 2019-10-15
DSHawk

SuuperW:
I'm working on NintendoDS support in BizHawk. I don't know if you wish to include this or any part of it in your repo.

for reference, discussion in the BizHawk repo: https://github.com/TASVideos/BizHawk/issues/1688

Oh, the é from Pokémon in NDSCart was changed by my editor for some reason, apparently it doesn't like that character.


poudink:
so this is for tasing support right?

SuuperW:
Yes, the purpose of this is to make TASing NDS games easier.

poudink:
Sounds like a great addition, then.

Arisotura:
as much as I like this idea, I have a couple concerns about this.

a) how much modification to the melonDS core does this require?

b) I don't do speedruns, so I personally don't have interest into this feature. I think it's neat if someone else contributes such a feature, but I hope there will be someone to atleast keep it maintained, someone who knows how it works and how it might break with melonDS updates.

vadosnaprimer:
If designed properly, changing original core behavior won't be needed. TAS related features can be made optional through a #define so that the one who builds it decides what gets added or changed.

SuuperW:
@Arisotura 
a) Very little, at least so far. I've had to make changes so that ROMs and savestates can use byte arrays rather than files (no change to existing behavior), and I have added a lag frame counter (minor update for savestate format). I am currently looking to port some code from DeSmuME to make bios/firmware files optional.
-Commit 4d3a3f8 for ROM from byte array. Includes some refactoring to avoid duplicate code between methods for loading from byte array and loading from file.
-Commit 30e330a for byte array savestates. Again most of the changes to Savestate.cpp are refactoring.
-Commits 50d610c and 74201d7 for lag frames.

b) I can't promise to maintain this long-term.

vadosnaprimer:
> I can't promise to maintain this long-term.

You don't have to explicitly maintain it, you can just make it maintainable long-term for whoever else wants to work on it.

src/libui_sdl/MelonAPI.cpp
```diff
@@ -86,6 +86,10 @@ DLL bool Init()
 
 	ResetCounters();
 
+    // rdireect console output so BizHawk can see it
```
vadosnaprimer: typo

src/ARMInterpreter.cpp
```diff
@@ -220,26 +221,46 @@ void A_MRC(ARM* cpu)
...
+    {
+        u32 oldcpsr = cpu->CPSR;
+        cpu->CPSR &= ~0xBF;
+        cpu->CPSR |= 0x93;
+        cpu->UpdateMode(oldcpsr, cpu->CPSR);
+
+        cpu->R_SVC[2] = oldcpsr;
+        cpu->R[14] = cpu->R[15] - 2;
+        cpu->JumpTo(cpu->ExceptionBase + 0x08);
+    }
```
vadosnaprimer: Can copy-paste be avoided?

src/NDS.cpp
```diff
@@ -410,12 +414,16 @@ void Reset()
...
-        for (i = 0; i < 16; i++)
-            ((u32*)ARM7BIOS)[i] = 0xE7FFDEFF;
+        s32 len = _binary_fakeBios7_bin_end - _binary_fakeBios7_bin_start;
+        if (len > sizeof(ARM7BIOS)) len = sizeof(ARM7BIOS);
+        memset(ARM7BIOS, 0, sizeof(ARM7BIOS));
+        memcpy(ARM7BIOS, _binary_fakeBios7_bin_start, len);
     }
     else
     {
+        ARM7->useHLE_bios = false;
```
vadosnaprimer: Can copy-paste be avoided?

src/ARMInterpreter.cpp
SuuperW: If you mean A_SVC and T_SVC, they were already separate and neither the original or thew new HLE code blocks are identical. (Line 249 has no >>1; Line 261 ends with -2.) I didn't figure it was worth refactoring at this point, and messing with the code structure pattern.

RSDuck:
adding bios HLE seem like something which deserves a separate pull request to me.

src/MemoryStream.cpp
```diff
@@ -0,0 +1,160 @@
...
+	DeInit();
+}
+
+// Make chunks point to data; does not copy data.
+void MemoryStream::Init(u8* data, s32 len)
+{
+	if (chunks) DeInit();
+
+	if (len > CHUNK_SIZE * MAX_CHUNKS)
+		throw "data array too big";
```
RSDuck: this would introduce exceptions into the code base, maybe we can come up with something different

src/ARM.h
```diff
@@ -132,6 +132,9 @@ class ARM
     NDS::MemRegion CodeMem;
 
     static u32 ConditionTable[16];
+
+    bool useHLE_bios;
+    u8 intrWaitARM_state;
```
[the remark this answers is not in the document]
SuuperW: Sorry, I copy/pasted those names without thinking about that. Will change.

InfamousKnight:
Does anyone know if save states are complete for melonds? Because that would prevent lots of desync issues if complete.

SuuperW:
This branch is failing the tests.
It passes when building the frontend, though, and only fails at the msys-dist.sh script. I do not understand what's happening there, and it isn't really important for my purposes. Considering all this, I will leave this issue for someone else.

I can undo commit a0baa7f to get the tests to pass, if this is wanted.

rzumer:
Your branch seems to implement several independent features right now. If you could split off single features into their own branches and open one PR per feature, it would be easier to review and merge them.

SuuperW:
Here's a start for splitting features into separate PRs. PR #602

PR #550, opened 2019-12-08
Support GBA game cartridges for DS connectivity

### LINKED ISSUE #390: Solar sensor support
This is a feature request for solar sensor emulation. As far as I know this feature is only available via GBA connectivity between Boktai DS (Lunar Knights) and the three GBA Boktai games. By inserting those cartridges, the embedded solar sensor is enabled and and transmits the amount of sunlight detected to the DS game. I am not aware of any DS emulator that supports this feature, but multiple GBA emulators can handle the solar sensor signals, which may be useful.

rzumer:
This adds basic slot-2 support, enabling ~partial~ complete? DS-GBA [game connectivity](https://nintendo.fandom.com/wiki/List_of_Nintendo_DS_games_with_GBA_connectivity) (for commercial titles). It may conflict somewhat with #506, but I tried to keep the design as simple as possible.

Implemented features are ROM and SRAM direct reads/writes, Flash memory reads/writes (for Pokémon), and GPIO reads/writes with solar sensor support (for Boktai DS/Lunar Knights). For the latter, the level of sunlight is controlled via the keypad left and right keys, and the level is displayed in the OSD when modified (only when a Boktai GBA cartridge is inserted).

Closes #390.

Further design notes:
- GBACart.{h,cpp} is mostly copied from the NDS equivalents, but the read/write implementations differ, and the solar sensor implementation is original. Most of the added code for Flash state and read/write functions is inspired from DeSmuME, and the solar sensor implementation from mGBA.
- GBA SRAM writes are committed immediately. In my tests I didn't see any reduction in performance from doing that compared to writing on exit, though there is general performance degradation for a short period when queuing a lot of writes (e.g. Pokémon migration). Note: DeSmuME does not seem to commit writes to the GBA SRAM at all.
- The solar sensor and its hotkeys (hard-bound to left and right keypad keys) are only processed if the inserted GBA cartridge has a gamecode that corresponds to a Boktai game.
- Save states are ~left TODO~ implemented, with the ROM header stored in the state to avoid depending on a file path.
- NDS.cpp has one new function `LoadGBAROM()` to load a GBA cart.
- The UI (main.cpp) will detect GBA files by their extension and use the correct loading path for them without adding a new menu item. The only change in the API is adding an `int slot` to a few functions.
- GBA carts are automatically reloaded in the UI after loading a DS cart, because `NDSCart::LoadROM()` calls `NDS::Reset()` (which clears all the cartridges). Shutting down the system will eject GBA carts, however.
- The CLI will accept an additional input file name to load a GBA ROM on boot (the order matters).

I also changed the value returned from GBA memory reads on the deactivated CPU from 0xFF to 0x00 based on GBATEK's documentation. Still no proper open bus handling, but all 0xFF does correspond to what a DSi produces when probed in those memory regions, so I think we should be fine.

Finally, I added my name in the copyright notice on `GBACart.{h,cpp}` and set the year range to 2019 only, but if you want the notice to be exactly the same in every file I don't mind and can revert it.


rzumer:
Note: writing to GBA SRAM is not yet implemented (I forgot that was a thing, notably in the Pokémon connectivity). Will try to get it done ASAP.

rzumer:
Open question: GBATEK claims that for DSi, "reading GBA ROM areas does return FFFFh halfwords instead of the usual open bus values though." Does it also apply to the deselected CPU? If someone can test the behavior on hardware that it would be nice to know, even though it probably does not matter in practice.

rzumer:
Added solar sensor emulation for Boktai DS/Lunar Knights connectivity with the GBA Boktai games.

RSDuck:
the left and the right arrow keys?

rzumer:
[Keypad](https://en.wikipedia.org/wiki/Keypad).

RSDuck:
laptops don't have those

rzumer:
@RSDuck Done.

rzumer:
Added `DoSavestate()` implementations for all the GBA state, but I have not tested it. Two things:
- the GBA ROM is saved in the state, which could make the size quite large, but on the other hand reduces the file dependencies and risk of breakage if people move their things around. GBA files don't get too large, so I think it's a fair tradeoff considering that GBA connectivity isn't used in many situations.
- I haven't tested it, so I'm not sure if the bool* casts to u8* are OK yet.

RSDuck:
hm, the save state thing is tricky indeed, but putting the ROM file into the save state sounds like asking for trouble in my ears. Maybe put the file path and a hash (the CRC functions already present can be used) into the save state. Then when loading a save state try to load the rom at the specified path, if it's not possible open a file dialog to locate it and verify it via the hash.

rzumer:
I don't want to introduce a circular dependency between the UI and the cartridge model. Storing path + hash is possible but like I said, is brittle. If the file moves should the state be loaded anyway? If so, and the state occurs during a series of cartridge reads or writes, it will break. If not, save states will break if the file is moved or renamed, or if the state is shared.

RSDuck:
you can change the Error field in [SaveState.h](https://github.com/Arisotura/melonDS/blob/master/src/Savestate.h#L34) to an int or an enum. This way you can flag a specific error if the file can't be loaded. You check for this error in the [gui code](https://github.com/Arisotura/melonDS/blob/master/src/libui_sdl/main.cpp#L1805) and if it occurs you open the file dialog. After selecting a file it's checksum is compared to the loaded one, if it doesn't match a dialog window pops up where you can choose between ignore, select again and cancel. Libui doesn't support such dialog boxes, but fortunately it's very easy to implement those. I already did something similar with an Ok - Cancel box for the jit, [it's really not hard](https://github.com/Arisotura/melonDS/tree/generic_jit/src/libui_sdl/libui). If the rom is swapped out or removed it's like immediately taking out the cartridge after the savestate was loaded. That's something which can happen on real hw too.

I really want to avoid 64 mb Pokemon Platinum save states containing whole Emerald roms.

rzumer:
I opted to implement my proposed alternative from my last comment. Besides save contents, only the 192 header bytes of the cartridge are stored in the state, which allows titles to be identified (they show up in firmware and are correctly detected in the title I tested, Boktai DS).

There is a more detailed explanation of the rationale in the code comments, but basically, I think it is less of a headache for everyone if a save states works with no added file dependencies, so that it can easily be shared or reused without the full GBA ROM.

However, GBA ROMs can still be loaded manually prior to a state load, and if the cartridge is the same as the one in the save state, its contents post-header will not be wiped.

RSDuck:
ah, I was surprised I hadn't read your suggestion, but then I noticed you edited your last post 😄

I think this is a good solution, the only thing I would ask is a message box or some other way to signal that only the header could be loaded.

rzumer:
I fixed all crashes found, changed the GBA cartridge reset system so it's possible to maintain the cartridge ROM header and SRAM when resetting without a tracked path, and added "(GBA ROM header only)" at the tail of the state load OSD message only if a GBA cartridge is included in the save state and has a different CRC from the currently loaded one (if any).

Arisotura:
let's merge this. ahahhhh

PR #554, opened 2019-12-11
Add GitHub workflow configurations for continuous integration with Linux and Windows

### LINKED ISSUE #301: Hate to ask this but.. is there a CI Server
I dont see a CI Server, but it might be a smart move since im using a windows machine and CMake is such a jerk to me. Not your codebase but in general. Or atleast add a compilation guide

### LINKED ISSUE #387: Autobuilds
Maybe melonds could have autobuilds.
Desmume uses https://www.appveyor.com/ and it's free for open source projects

### LINKED ISSUE #540: GitHub Actions support
Add GitHub Actions support for automated builds.

rzumer:
This enables continuous integration with GitHub Actions, allowing developers to see if their commits/PRs are breaking the build.
Example run [here](https://github.com/rzumer/melonDS/commit/7027813cb268c8b168a2c6ac7e871761f02a4435/checks?check_suite_id=353617464).

Both Linux and Windows are included, and the compiled binaries are exported with shared libraries and `romlist.bin`.

Closes #540, closes #301 and closes #387.

### DIFF
--- a/.github/workflows/build-ubuntu.yml
+++ b/.github/workflows/build-ubuntu.yml
@@ -17,11 +17,11 @@ jobs:
       shell: bash
       working-directory: ${{runner.workspace}}
       run: | # Fetch a new version of CMake, because the default is too old.
-           wget -nv https://github.com/Kitware/CMake/releases/download/v$CMAKE_VERSION/cmake-$CMAKE_VERSION-Linux-x86_64.tar.gz \
-           && tar -zxf cmake-$CMAKE_VERSION-Linux-x86_64.tar.gz \
-           && sudo rm -f /etc/apt/sources.list.d/dotnetdev.list /etc/apt/sources.list.d/microsoft-prod.list \
-           && sudo apt-get update \
-           && sudo apt-get install gtk+-3.0 libcurl4-gnutls-dev libpcap0.8-dev libsdl2-dev
+        wget -nv https://github.com/Kitware/CMake/releases/download/v$CMAKE_VERSION/cmake-$CMAKE_VERSION-Linux-x86_64.tar.gz \
+        && tar -zxf cmake-$CMAKE_VERSION-Linux-x86_64.tar.gz \
+        && sudo rm -f /etc/apt/sources.list.d/dotnetdev.list /etc/apt/sources.list.d/microsoft-prod.list \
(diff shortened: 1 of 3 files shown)


MojoJojoDojo:
I am not sure @Arisotura gonna merge it.
He doesn't want to maintain it.

RSDuck:
It's only a single file, if we want to remove it again, it won't be that hard. If Arisotura, doesn't want to maintain it I would keep an eye on it too.

rzumer:
> Btw I read this PR from another project : [xqemu/xqemu#266](https://github.com/xqemu/xqemu/pull/266)
> I think there is a way to get MSYS2 working than.

Cool, I will look into it.

RSDuck:
can you also enable building for pull requests?

poudink:
isn't arisotura a windows user? a linux-only developer feature isn't going to be very useful to her in that case.

rzumer:
Added a Windows configuration. It fetches MSYS2 from XQEMU because the official MSYS2 distribution packages cause CI to fail due to permission errors which I haven't managed to solve. It does not export artifacts at this point.

rzumer:
Now exporting build artifacts (dependencies not included).

RSDuck:
> isn't arisotura a windows user? a linux-only developer feature isn't going to be very useful to her in that case.

I'm using Windows too. But 99% of the code of melonDS is shared amongst platforms, even if we wouldn't have Windows CI builds (thanks for that), we would still benefit (almost always) by knowing whether a pull requests builds or not.

RSDuck:
it seems like you're currently only putting the exe file as artefacts, can you add the romlist.bin and for Windows the appropriate dlls (see https://github.com/Arisotura/melonDS/blob/master/msys-dist.sh) too and put everything into a zip file or so?

rzumer:
@RSDuck Done now, sorry if someone got flooded by emails because of my force-pushes.

MojoJojoDojo:
I think you should clone the XQEMU repo and not use it as is lol.

rzumer:
I'm not interested in maintaining a redundant package. If @Arisotura wants to do that in parallel feel free, but energy would be better spent in finding a way to run off the official distribution packages, and I fought with it enough for now.

PR #557, opened 2019-12-19
BUGFIX: Fixed misspelled 'float' from original 'floor'

NewJoarpa:
Fixed misspelled 'float' from original 'floor' in 'src\libui_sdl\main.cpp' line 590.
The code couldn't compile before this bugfix.

Tool used: CodeBlocks.


RSDuck:
I experienced this myself, though floor is right. The thing is that floor is from math.h but with some compilers it slips through is available eventhough math.h isn't directly included.

In any way the casting a float to float is in the same way redudant as flooring a positive float which is going to be casted to int.

NewJoarpa:
Got it.

PR #579, opened 2020-02-21
Made randomizing the MAC address optional.

WouterPeeters9:
Fixes Arisotura/melonDS#447

This is my first time working in C++, so I don't know if I've broken any code laws.
I can confirm that this fixes the anti-time travel bug in Pokémon Platinum though.

### DIFF
--- a/src/Config.cpp
+++ b/src/Config.cpp
@@ -34,6 +34,8 @@ int Threaded3D;
 int GL_ScaleFactor;
 int GL_Antialias;
 
+int RandomizeMAC;
+
 ConfigEntry ConfigFile[] =
 {
     {"3DRenderer", 0, &_3DRenderer, 1, NULL, 0},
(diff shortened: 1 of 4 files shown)


RSDuck:
looks good, though I wonder whether it is a good idea to hardcode the MAC address if it's not randomised @Arisotura

nathanielcwm:
What about using the systems MAC address?

JamesTheAwesomeDude:
Some ideas:

* Use the host's MAC ("lowest-indexed" card would probably hold water)
* Use whatever the ROM has burned-into it (allow changing/setting it?)

InfamousKnight:
Is there anything else that needs to be done to get this merged?

RSDuck:
we'll implement it properly ourselves in the new QT frontend.

Arisotura:
I redid it but for the new UI, sorry that your PR didn't make it in in time

PR #582, opened 2020-02-29
Statically link libraries on Windows

nadiaholmquist:


rzumer:
Please check the CI workflow, we need to exclude the libraries from the distributed artifact on Windows if they're no longer needed.

nadiaholmquist:
Looks like it still works fine and includes no additional DLLs, I suppose it could be removed but it doesn't cause any problems as-is.

WaluigiWare64:
We build statically with Windows currently, and the libui gui is gone.

PR #592, opened 2020-03-29
Firmware chooser dialog (support for custom BIOS/FW paths)

StaticSaga:
This PR adds the ability to change the FS paths of the bios7.bin, bios9.bin and firmware.bin files, graphically or by editing the config.
I think that the files could actually be managed in a separate struct, but for now (there are only three files) I do not think that this feature could be useful.
There is a very simple file chooser in the Config->Firmware dumps, that also warns the user that DS/DS Lite emulation requires a DS firmware.bin and not a DSi one. It builds on Windows and on Linux, but the CI has problems with the Windows build, I don't know why.
 
Here are the screenshots from a Linux system.
FirmwareChooser2
FirmwareChooser

### DIFF
--- a/src/NDS.cpp
+++ b/src/NDS.cpp
@@ -149,6 +149,21 @@ bool Running;
 bool RunningGame;
 
 
+char* Bios7Path;
+char* Bios9Path;
+char* FirmwarePath;
+
+void SetFilePaths(char* bios7, char* bios9, char* firmware) {
+    Bios7Path = bios7;
+    Bios9Path = bios9;
+    FirmwarePath = firmware;
+}
+
+char* GetFirmwarePath() {
+    return FirmwarePath;
+}
+
+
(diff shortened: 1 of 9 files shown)


RSDuck:
this looks good, the only problem is that we're about to switch from libui to another gui library

StaticSaga:
Have you already chosen the library? (is it related to the FLTK branch or was that just a PoC?)
If so, I can help.

RSDuck:
yes, we tried FLTK as an option, but ultimately we let it be, because it's lack of proper layouting (pixels coordinates are everywhere).

We're currently still fighting with the decision to give in and to use QT (which wanted to avoid). If you want to discuss more easily, hop onto IRC (see the readme).

StaticSaga:
Thank you for your support. Closing this PR for now.

PR #594, opened 2020-04-01
Accurize "screen sizing" UI text

JamesTheAwesomeDude:
The description is currently imprecise/confusing/misleading. "Auto" (I believe) implies it's been curated. But there any many cases in popular games where it's clearly not chosen the "correct" screen.

Would it be reasonable to label this option _descriptively_, rather than _optimistically_?

> 'auto' mode emphasizes whichever screen is set to the main video engine (the one with 3D). hotkeys for switching between layout modes are planned.

– https://github.com/Arisotura/melonDS/issues/489#issuecomment-511202251

### DIFF
--- a/src/libui_sdl/main.cpp
+++ b/src/libui_sdl/main.cpp
@@ -2580,7 +2580,7 @@ void CreateMainWindowMenu()
         uiMenuItemOnClicked(MenuItem_ScreenSizing[1], OnSetScreenSizing, (void*)&kScreenSizing[1]);
         MenuItem_ScreenSizing[2] = uiMenuAppendCheckItem(submenu, "Emphasize bottom");
         uiMenuItemOnClicked(MenuItem_ScreenSizing[2], OnSetScreenSizing, (void*)&kScreenSizing[2]);
-        MenuItem_ScreenSizing[3] = uiMenuAppendCheckItem(submenu, "Auto");
+        MenuItem_ScreenSizing[3] = uiMenuAppendCheckItem(submenu, "Emphasize 3D engine");
         uiMenuItemOnClicked(MenuItem_ScreenSizing[3], OnSetScreenSizing, (void*)&kScreenSizing[3]);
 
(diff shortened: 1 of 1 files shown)


RSDuck:
I never perceived auto to mean curated. It's a common name meaning just letting the computer decide, how good it's decision is, can be debatable. Secondly I question whether "Emphasis 3D engine" makes more sense to most users.

JamesTheAwesomeDude:
> I question whether "Emphasis 3D engine" makes more sense to most users.

Relative to "Emphasis Auto", I believe it's a direct improvement. It's unclear, in any case, what "Auto" _means_ without reading that description—it provides no information about the "decision" process happening.

-----

I suppose the crux is: Is this option intended/planned to become "smarter"? Or is the following indefinite? (If the latter, then the renaming seems appropriate.)

image

Arisotura:
'emphasize 3D engine' would be confusing, because some games don't use the 3D engine at all, while others can choose to use it for 2D-like graphics, etc. the thing with the DS is that the possibilities for getting shit displayed onscreen are many.

JamesTheAwesomeDude:
> 'emphasize 3D engine' would be confusing

This is true—however, I still argue that it'd be _less_ confusing that the UI's status quo: which promises nothing, and delivers something entirely unexpected.

I tried experimenting with a bunch of other phrasings before sending the PR: “Primary Video”, “Primary Video Engine”, “Primary/3D Engine”, “Primary/3D” Video”, “Primary Engine Screen”—but, of these, “3D Engine” emerged as (to my judgement) the most concise and accurate.

If this setting doesn't have any planned improvements in the mid-to-near future—if it's going to _remain_ as a setting which emphasizes the screen attached to the 3D engine—I would recommend the following as the least confusing UI item I can come up with:

image

(Again, I want to emphasize that “Auto” is really quite confusing, since it doesn't align with what the user would want in almost as many cases as not. And, importantly, it doesn't even give the user a **hint** as to what it actually does!)

PR #601, opened 2020-04-13
Fix build with latest MSYS2 SDL2 package

nadiaholmquist:


rzumer:
Upstream issue: https://github.com/msys2/MINGW-packages/issues/6380

PR #602, opened 2020-04-13
Embed ROM list, HLE bios

SuuperW:
These commits were originally part of my DSHawk branch, and contain required changes for BizHawk integration. PR #527

Limited support for running games without bios or firmware files. (This code builds on code for embedding resource files, which is why embedding the ROM list is included in this PR.)
It still checks for these files on start-up, but this can easily be commented out at main:2753 for testing.
I've left the check as-is, because there are inaccuracies (causing some games to fail completely) and use of bios/firmware files is still strongly recommended. Also because that check is bypassed when running via BizHawk, which is my aim. Decisions about how to re-work dialogs and recommend use of these files to MelonDS users is left for someone else.


Arisotura:
re: HLE BIOS

I don't like the idea of having C++ hooks for those, I'd rather provide an alternate BIOS like DraStic does.

also, regarding romlist.bin, the idea of having it be external was that, if needed, it could be updated without having to push a new release.

SuuperW:
I wouldn't know how to provide an alternate BIOS and it seems like something that would take a lot of work to make. Is this something that's on your TODO list already?

Would it be good to have the current romlist embedded and still allow the use of a separate file if the loaded ROM isn't found there? My main reason for embedding it is to simplify integration in BizHawk.

Arisotura:
the alternate BIOS wouldn't be that much harder to code than a HLE one, and it would imply less code complexity on melonDS's side (as it can be loaded just like an original BIOS). it would even be more accurate and allow to reimplement certain things cleanly (like the decompression algorithms that use callbacks).

however, thinking back on it, embedding romlist.bin could be a good idea (maybe while keeping support for an external one? dunno).

SuuperW:
I didn't code the HLE bios, it's 99% copied from DeSmuME. Hence why making an alternate BIOS would be a lot more work.
(I do agree that it's a nicer solution, though.)

PR #607, opened 2020-04-23
Dirty hack to toggle fullscreen with F11 in gtk & windows

lucianposton:
Dirty hack to make F11 toggle fullscreen in the windows and gtk builds, based on https://github.com/Arisotura/melonDS/issues/237#issuecomment-436489499. I just wanted to throw this up on github in case anyone else compiling the gtk build wanted a hardcoded f11 fullscreen toggle. Feel free to close.

See: https://github.com/Arisotura/melonDS/issues/237


RSDuck:
thanks for the PR, though as it's based on the old libui GUI I close it.

PR #608, opened 2020-04-23
Add an option for 240px screen gap

lucianposton:
Adds an option for 240px screen gap.

Why 240px? It's the number of pixels you'd want if you have 2 1920p monitors side-by-side, and you want one DS screen centered & fullscreen on one monitor while the 2nd DS screen is aligned on the near edge of the 2nd monitors. :)
DS is 4:3 aspect ratio. 240=(1920-1080*(4/3))/2


RSDuck:
thank you for your contributions, unfortunately we don't accept any changes to the current GUI, because we're currently replacing it by a new GUI using QT instead of libui

PR #609, opened 2020-04-23
Add hotkey to switch screen emphasis

lucianposton:
Just throwing this up on github in case others were looking for a way to add a hotkey to switch screen emphasis between bottom/top until melonDS's QT ui is finished.


Arisotura:
I like this idea; might also have one for cycling through screen rotation

however, we switched to a Qt UI, which is why I didn't merge this; if you port this to the Qt UI, I will merge it

PR #610, opened 2020-04-23
Hide mouse when not in use

lucianposton:
Another QoL hack that hides the cursor when it is not in use (only implemented the gtk part though). I found the cursor to get in the way after use when switching between different DS window emphasis.


Arisotura:
'ts a nice idea, but this will need to be adapted to the new Qt frontend

RSDuck:
we moved to QT

PR #611, opened 2020-04-23
Larger unemphasized screen, when possible

lucianposton:
Increases the size of the smaller, unemphasized screen, when possible, making its text readable on 1920x1080.

Before:
img-2020-04-23T22:56:30 236Z

After:
img-2020-04-23T22:54:49 927Z

### DIFF
--- a/src/libui_sdl/main.cpp
+++ b/src/libui_sdl/main.cpp
@@ -1398,6 +1398,7 @@ void SetupScreenRects(int width, int height)
         // side-by-side
 
         int heightreq;
+        int emph_smaller_width;
         int startX = 0;
 
         width -= gap;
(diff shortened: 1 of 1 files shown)


poudink:
I still see a black bar between the second screen and the first screen. couldn't you increase its size further, so that it uses up all of the space available?
Something like this.
BiggerUnemphasizedScreen

lucianposton:
It does exactly that. My example screenshot just had the gap enabled.

lucianposton:
screenshot with gap toggled off, fwiw:
img-2020-05-02T01:59:13 215Z

poudink:
oh okay, I see. that's pretty great then.

PR #616, opened 2020-04-30
emulate time

SuuperW:
Adds an option to use either the host system's time or emulated time with a specified time at boot.
Does not include any UI to set this option; defaults to current behavior of using host system's time.


src/NDS.cpp
```diff
@@ -823,70 +826,72 @@ u32 RunFrame()
...
     FrameStartTimestamp = SysTimestamp;
 
-    if (!Running) return 263; // dorp
-    if (CPUStop & 0x40000000) return 263;
-
-    GPU::StartFrame();
-
-    while (Running && GPU::TotalScanlines==0)
+    bool didAnything = false;
+    if (Running && !(CPUStop & 0x40000000))
```
RSDuck, line 830: how is this different than the original behaviour? If I'm not wrong you could just increment FramesSinceBoot at the beginning of RunFrame.

src/NDS.cpp
```diff
@@ -66,6 +66,7 @@ ARMv5* ARM9;
 ARMv4* ARM7;
 
 u32 NumFrames;
+u32 FramesSinceBoot;
```
RSDuck, line 69: FramesSinceBoot seems reduandant to me, you use NumFrames too.

Btw both seems not be reset when loading a new game. I think for NumFrames it was an oversight from Arisotura.

src/NDS.cpp
SuuperW: One reason for doing it that way was that it should be 0 on the first frame of emulation. If I simply incremented it at the start of the frame, the RTC would think it's 1/60th of a second later than it is. This could be solved by initializing it to -1 instead of 0, if FramesSinceBoot were a signed value.

Also this change is for the other counters as well, NumFrames and NumLagFrames. I guess these could have a similar fix with initializing to -1. But it seems simpler to me to do it this way rather than make those changes, and by having the counter increments at the end it's more obvious when reading the code that they are counting completed frames rather than started frames.

src/NDS.cpp
SuuperW: It is redundant in this code, but not in the DSHawk branch. There, NumFrames is what's reported to the client, and can be reset by the client; any frame can be arbitrarily marked as "frame 0". But this should not affect the RTC.

Neither are reset when loading a ROM, since a power cycle does not mean resetting the session. (And currently using the power button [I have more code for this on the DSHawk branch.] to power cycle just re-loads the ROM.) There must be some way to reboot the system without resetting any of these frame counters.

RSDuck: ah I agree, I'll communicate that with Arisotura, we'll see what we can do with the current frontend refactor

PR #618, opened 2020-05-07
high resolution software renderer

SuuperW:
I think the title says enough.


RSDuck:
Do you know if there is a performance hit at 1x resolution?

SuuperW:
There's sure to be some performance hit, but it was not enough for me to notice casually, and I did not do any testing for that.
E.g., in GPU2D.cpp there are various for loops which will execute 1 or 0 times at 1x resolution.

HiresPosition contains different values than FinalPosition did, since of course it's hires. But every place where HiresPosition is used, it is multiplied by ResMultiplier, then shifted right by HD_SHIFT (7). Which I expect results in identical behavior at 1x. But again, I did not actually test it beyond casual use.

I can do some testing here later.

SuuperW:
I changed the zero-dot W check to use native resolution and made a couple changes for YTop/Bottom to compare with current resolution.
I then compared screenshots that should be identical, one with HD_SHIFT 7, and one with HD_SHIFT 0. There's a small difference with some of the small polygons.
Changing the sort key to use native resolution resolved this difference.
I also verified that HiresPosition >> HD_SHIFT equals the original FinalPosition.

I also checked FPS on master and on HiRes_Soft branches, and found a ~10% drop. I will look for some ways to improve this.

RSDuck:
Take a look at the final [composition shader](https://github.com/Arisotura/melonDS/blob/master/src/libui_sdl/main_shaders.h#L99) of the OpenGL renderer. Basically everything is rendered at 1x with a dummy value where the 3D is inserted as the final pass, which is not only less invasive than your current approach but probably more performant too.

To be 100% sure maybe keep the old FinalPosition intact and add instead add a second path in GPU3D_Soft.cpp (separated via a template argument) so that at 1x the same code as currently is run as now.

SuuperW:
If I'm understanding correctly, you're suggesting that the 2D graphics still be rendered at 1x, then scaled up afterwards and the 3D buffer inserted at that point.
I tested this earlier, and found it significantly slower than my current approach.

I am considering making separate code paths for the most slowed-down parts of the hi-res code. E.g. GPU2D::DrawPixel_Normal can be split into _Normal and _HiRes versions. (Or use a template.)

I personally dislike the idea of making separate code paths for hires as a "just in case". I'd rather make some code to compare lossless video recordings of a few test cases, if that would be satisfactory. (Such recordings can be easily obtained through BizHawk.) Or, does Arisotura already have an easy enough way to test accuracy? (There must have been tests at some point, to know how accurate the current implementation is.)

poudink:
huh, I thought arisotura said implementing something like this in the software renderer would be a lot of work/very slow/could bork accuracy/all of the above so I'm surprised to see this

RSDuck:
> If I'm understanding correctly, you're suggesting that the 2D graphics still be rendered at 1x, then scaled up afterwards and the 3D buffer inserted at that point.
I tested this earlier, and found it significantly slower than my current approach.

interesting, I didn't expect that! My aim was mostly to keep complexity down, but revisiting it, there aren't so many changes. It might even clear the path for high resolution rendering of affine BGs and sprites.

> I am considering making separate code paths for the most slowed-down parts of the hi-res code. E.g. GPU2D::DrawPixel_Normal can be split into _Normal and _HiRes versions. (Or use a template.)

yeah, if we have already have an indirect function in an inner loop, we might as well profit from it...

> I personally dislike the idea of making separate code paths for hires as a "just in case". I'd rather make some code to compare lossless video recordings of a few test cases, if that would be satisfactory. (Such recordings can be easily obtained through BizHawk.) Or, does Arisotura already have an easy enough way to test accuracy? (There must have been tests at some point, to know how accurate the current implementation is.)

I'll ask Arisotura, she's put quite a lot of work into getting these values right. She probably has some test cases to verify this, though your suggestion wouldn't hurt either.

> huh, I thought arisotura said implementing something like this in the software renderer would be a lot of work/very slow/could bork accuracy/all of the above so I'm surprised to see this

well most of the points still stand:

> very slow

2x-4x will probably work decently with this and a beefy processor, but at some resolutions GPUs can't really be matched anymore, though I haven't tried it.

> could bork accuracy

if we're not careful that could happen

MojoJojoDojo:
> high resolution rendering of affine BGs and sprites.
That would be nice even though only 1-2 games use it extensively on the DS lol.

SuuperW:
I have no idea what affine BGs and sprites are, so don't look to me to implement that. It seems to me that making other graphics HD would be considerably more work, though, since a large portion of that goes through the emulated VRAM.

And yeah, rendering speed at 2x is reasonable for my purposes. But, especially at actual HD resolutions, the OpenGL renderer is much faster.

poudink:
I think they mean high resolution scaling/rotation of 2D elements for things like HD mode 7 effects in the few games that use it.

SuuperW:
Performance at 1x is much closer to master now.

poudink:
What really is needed for this to be merged? I can try recording two lossless videos and comparing them as was suggested earlier to see if this causes any difference in rendering at 1x, if needed.

RSDuck:
> What really is needed for this to be merged? I can try recording two lossless videos and comparing them as was suggested earlier to see if this causes any difference in rendering at 1x, if needed. It seems to me like the main issue is the lack of proper testing.

this adds quite some complexity we need to maintain, while it doesn't add that much. We rather have a clean separation where enhancements don't clog up the accuracy foccused part of the codebase.

RianCaio:
Soo there anything new to this?

WaluigiWare64:
This is also quite old and has a lot of merge conflicts to be resolved.

PR #651, opened 2020-06-16
Slirp

RSDuck:
Just to let build


poudink:
I've been following this branch but I can't tell what it's really supposed to be. What is it?

RSDuck:
it implements the DSi wifi hardware and uses libslirp for indirect mode wifi (which makes it a lot more reliable).

PR #678, opened 2020-07-22
Add support for loading ROMs from a variety of compressed files

WaluigiWare64:
This pull request adds the ability to load ROMs from an archive file using the `libarchive` library.
What it does:
- Checks is the file selected has an extension of `zip`, `7z`, `rar`, `tar`, `tar.gz`, `tar.xz` or `tar.bz2`
- If it does, then it gets a list of all the files inside the archive by calling `Archive::ListArchive()`. If there is more than one, then it displays a dialog box to select the ROM
- Then, it extracts the file by calling `Archive::ExtractFileFromArchive()`, which returns where the extracted file is located
- It then changes the `filename` variable to the location of the extracted ROM.

I'm still learning C++ so any feedback would be appreciated :D

### To-do list
- [x] Link statically
- [x] Switch to libarchive
- [x] Add dialog box to select ROM
- [ ] Load ROM directly from RAM instead of extracting to disk


src/frontend/qt_sdl/main.cpp
```diff
@@ -1399,16 +1403,52 @@ QString MainWindow::loadErrorStr(int error)
...
+    //Open the ZIP archive
+    int err = 0;
+    zip *z = zip_open(zipName, 0, &err);
+    
+	struct zip_stat st;
+    zip_stat_init(&st);
+    zip_stat_index(z, 0, 0, &st); //Get information about the file at index 0
+	
+    //Allocate memory for its uncompressed contents
+    char *contents = new char[st.size];
```
RSDuck, line 1416: we make the star on the left side

contents needs to be deleted

prefer u8 over char for a binary stream

src/frontend/qt_sdl/main.cpp
```diff
@@ -1399,16 +1403,52 @@ QString MainWindow::loadErrorStr(int error)
     }
 }
 
+std::string extractROM(char* zipName, std::string zipDir){
+    //Open the ZIP archive
+    int err = 0;
+    zip *z = zip_open(zipName, 0, &err);
+    
+	struct zip_stat st;
```
RSDuck, line 1411: inconsistent identation

src/frontend/qt_sdl/main.cpp
```diff
@@ -1399,16 +1403,52 @@ QString MainWindow::loadErrorStr(int error)
...
+    zip_stat_init(&st);
+    zip_stat_index(z, 0, 0, &st); //Get information about the file at index 0
+	
+    //Allocate memory for its uncompressed contents
+    char *contents = new char[st.size];
+	
+    //Read the compressed file
+    zip_file *f = zip_fopen_index(z, 0, 0); //Open file at index 0
+	zip_fread(f, contents, st.size);
```
RSDuck, line 1420: indentation

src/frontend/qt_sdl/main.cpp
```diff
@@ -1399,16 +1403,52 @@ QString MainWindow::loadErrorStr(int error)
...
+    zip_fclose(f);
+
+    zip_close(z);
+	
+	//Write the file (binary mode)
+    std::ofstream(zipDir + "/" + st.name, std::ofstream::binary).write(contents, st.size);
```
RSDuck, line 1426: throwing it onto the disk and then reading it again is not optimal, especially if it's just directly next to the zip file and not some temp folder.

We currently don't have any interface from the core to load a rom from a blob. We already discussed this internally, it's probably the best to get @Arisotura involved as well

src/frontend/qt_sdl/main.cpp
WaluigiWare64: Weird. The indentation is fine on Notepad++

src/frontend/qt_sdl/main.cpp
WaluigiWare64: fixed.

src/frontend/qt_sdl/main.cpp
WaluigiWare64: fixed.

src/frontend/qt_sdl/main.cpp
WaluigiWare64: Alright, i've made it delete contents.
Also, when I change it's type to u8, it gives me errors like `error: invalid conversion from 'u8*' {aka 'unsigned char*'} to 'const char*'`

RSDuck: then cast it to char*

WaluigiWare64: ok, done it.

WaluigiWare64:
Alright, so I've fixed some bugs and statically linked libzip.
Is there anything else I need to do @RSDuck?

RSDuck:
my main criticism still stands, extracting the ROM as a file instead of into memory is imho not the best way to handle this

WaluigiWare64:
Alright, I'll see what I can do.

src/frontend/qt_sdl/main.cpp
Arisotura: yup. I have ideas for this (providing a ROMSource interface that the core will use to access the ROM, or whatever name is good for you; then frontends will implement classes derived from that interface, to implement loading from a file, from a ZIP, etc...)

archives also pose the problem that we do need to let the user select a file from within the archive

WaluigiWare64: > providing a ROMSource interface that the core will use to access the ROM

wouldn't this mean you would have to change a lot of the code. most of the file work is done using fseek, fread which take file pointers.

> archives also pose the problem that we do need to let the user select a file from within the archive

shouldn't be too hard. i could just read all the files inside the zip, if there's one, launch it, or if there's more, open a dialog box. 

> and also, where do save files and all go?

what about putting them in a folder with the same name of the zip file. for example, Mario Kart DS.nds inside DS_ROMS.zip would be placed in DS_ROMS/Mario Kart DS.sav.

ImSpecial:
Can you also add .7z as well to this PR? Not sure how niche .7z is, but it has better compressions compared to plane old .zip and I use that for archivals when possible.

WaluigiWare64:
I'll see.
That'll require switching over to `libarchive`.
I also need to fix some of the issues with this PR.

ImSpecial:
Thanks, sorry to bud in here and ask, but when I saw this PR it seemed relevant since you're already are in the part of the code that handles this sort of stuff and in the right mindset as well, and well, not worth a new issue/request post for .7z support since this here exists. But thanks for looking into it.

src/frontend/qt_sdl/main.cpp
```diff
@@ -1412,15 +1416,56 @@ QString MainWindow::loadErrorStr(int error)
...
+    int err = 0;
+    zip *z = zip_open(zipName, 0, &err);
+    
+    struct zip_stat st;
+    zip_stat_init(&st);
+    zip_stat_index(z, 0, 0, &st); //Get information about the file at index 0
+    char newName[255];
+    strcpy(newName, st.name); //fix for Linux invalid encoding filename	
+    //Allocate memory for its uncompressed contents
+    u8 *contents = new u8[st.size];
```
v1993, line 1430: I'd suggest using `auto contents = std::make_unique<u8[]>(st.size)` to avoid potential for memory leaks. Use `contents.get()` instead of `contents` in following code and also include `<memory>`.

v1993 (changes requested):

WaluigiWare64:
I'm working on this PR again, and am switching to `libarchive` for support of many more archive formats.

src/CMakeLists.txt
```diff
@@ -94,6 +94,7 @@ if (ENABLE_JIT)
 	endif()
 endif()
 
+
```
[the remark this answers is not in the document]
WaluigiWare64: nope.
must have forgot to remove the extra line.

src/frontend/qt_sdl/main.cpp
WaluigiWare64: Thanks, have included this in the libarchive extract function.

nathanielcwm:
@WaluigiWare64 idk how useful this may be but #364 was a previous attempt at this using libarchive.

WaluigiWare64:
Yep I've seen that PR, although it seems to have no activity anymore.
Also I'm nearly done, probably will push the commit today.

nathanielcwm:
Might be good to squash your old commits that aren't using libarchive.

src/CMakeLists.txt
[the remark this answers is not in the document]
WaluigiWare64: Done

RSDuck:
> Might be good to squash your old commits that aren't using libarchive.

we squash everything in the end anyway

PR #729, opened 2020-08-25
Create a melonDS Updater

WaluigiWare64:
This adds support for updating melonDS.
It uses the GitHub API to get the latest release for the current platform. If it is newer than the current version (need to find a way to get that), it downloads the latest release. It uses `libcurl` and `nlohmann-json`.
This doesn't work with Windows currently becuase MSYS2's `libcurl` package is [broken and cannot link statically](https://github.com/msys2/MINGW-packages/issues/3538).
### To-do list
- [ ] Get the GitHub access token from users
- [ ] Extract melonDS.exe from the zip file downloaded and extract it (replacing the old melonDS)
- [ ] Figure out a way to get the current artifact version.
- [ ] Add support for getting stable updates. 
- [ ] Find a workaround for MSYS2's broken libcurl package.


rzumer:
> Get the GitHub access token from users
It should be seamless IMO. If GitHub can't offer that then better host the current version + release packages on the official site.

src/frontend/Updater.hpp
```diff
@@ -0,0 +1,109 @@
...
+    CURLcode res;
+    FILE *fp;
+  
+    std::string readBuffer;
+
+    curl = curl_easy_init();
+    if(curl) 
+    {
+        curl_easy_setopt(curl, CURLOPT_URL, webURL);
+        curl_easy_setopt(curl, CURLOPT_USERAGENT, "curl/7.77.1");
```
rzumer, line 46: Is `7.77.1` a hardcoded version string?

src/frontend/qt_sdl/main.cpp
```diff
@@ -1872,6 +1878,21 @@ void MainWindow::onChangeAudioSync(bool checked)
...
+    emuThread->emuPause();
+    int updateCheck = Updater::checkForUpdates("dummy ver");
+    if (!updateCheck)
+    {
+        QMessageBox::information(this, "melonDS", "You are already on the latest version.");
+        emuThread->emuUnpause();
+    }
+    else 
+    {
+        Updater::installUpdate()
```
rzumer: Should have a confirmation before actually installing, maybe with a link to release notes if that's possible.

src/frontend/Updater.hpp
WaluigiWare64: yep it's a version string.
the GitHub API wouldn't work if I didn't have a useragent.

src/frontend/qt_sdl/main.cpp
WaluigiWare64: will do.

WaluigiWare64:
Found a way to get commit messages and other info, and make sure that the build is from this repository, not any forks. Will do this tomorrow.

* cURL https://api.github.com/repos/Arisotura/melonDS/actions/runs (Workflows API)
* Parse it
* Interate over `json[workflow_runs]`
* see if `obj[head_repositry][name]` is `Arisotura/melonDS` if not, skip.
* cURL `obj[artifacts_url]`, parse, and see if `name` is current platform. If not, skip.
* If `name` is current platform, get `obj[head_commit][message]`. This is the commit message.
* Now prompt user to update, if yes, use the old `obj[artifacts_url]` and cURL the file.

WaluigiWare64:
Also have found a workaround script for building `libcurl` ourselves.
I converted [this script](https://github.com/nu8/gulf/blob/master/c-https/2-curl.ps1) to a Bash script and can confirm it works.

src/frontend/Updater.hpp
rzumer: Then use the actual version string in libcurl headers or something, hardcoding is not good practice.

src/frontend/qt_sdl/main.cpp
rzumer: You can just link to the latest release page, i.e. https://github.com/Arisotura/melonDS/releases/latest.

WaluigiWare64: will do

src/frontend/Updater.hpp
WaluigiWare64: done.

src/frontend/qt_sdl/main.cpp
WaluigiWare64: got the commit info, so done.
now for stable releases.

Arisotura:
I have two main concerns over this:

1. make it optional and opt-in. I abhor forced updates.
2. there is a security concern with this, we need to ensure it's not possible for an attacker to feed bogus binaries through the updater.

WaluigiWare64:
> I have two main concerns over this:
> 
>     1. make it optional and opt-in. I abhor forced updates.
> 
>     2. there is a security concern with this, we need to ensure it's not possible for an attacker to feed bogus binaries through the updater.

It only checks for updates when the menu item is clicked. It downloads the melonDS.zip for the user to extract later. Currently this only downloads from GitHub Actions, but I will add stable releases and Azure soon.

WaluigiWare64:
only thing that's needed now is to get the current commit. perhaps #1117 could be useful.

src/frontend/qt_sdl/UpdateUtil.cpp
```diff
@@ -0,0 +1,165 @@
...
+#include <QEventLoop>
+
+#include <QFile>
+
+#include <QString>
+
+#include <QVersionNumber>
+
+
+QByteArray request(QString& errString, QUrl webURL, QByteArray token="")
```
RSDuck, line 36: PascalCase (multiple cases, too lazy to mark them all)

src/frontend/qt_sdl/UpdateUtil.cpp
```diff
@@ -0,0 +1,165 @@
...
+    if (reply->error() != QNetworkReply::NoError)
+        errString = reply->errorString();
+    else
+        errString = "";
+    return reply->readAll();
+}
+
+namespace Updater
+{
+    QUrl downloadURL;
```
RSDuck, line 57: no indentation in namespace

RSDuck (commented):
looks mostly good, though having to extract it by oneself is kind of a bad UX.

Yoshi9800:
Ok guys put this into master build now is done put into emulator

KostaSaizo7:
As an average user, I am not really a fan of this PR in its current state. Here's a comparison with Dolphin's updater from a user perspective, in order to update to the latest dev version:

**_Dolphin_**

1. After opening Dophin, the below window pops up:
image

2. The user clicks the "Install update" button and after that the updater does everything it is needed (downloads the latest dev version, extracts and replaces the old files and launches Dolphin again).

So, it is **just one click.**

**_MelonDS_**

1. After opening MelonDS, the user navigates to About>Update Channel and select "Development".

2. About>Check for updates, the below window pops up:
image

3. The average user **won't know what this means**. So the user either gives up or googles it (I googled). So, to create the token you need a Github account (the average user won't have one), then go to Github Settings>Developer settings>Personal access tokens>Generate new token, then give a name, then change the expiration to "no date" (so it will not expire in the future), then select the "public_repo" scope, then "Generate token" and finally copy the token to the updater. 

4. After all that, the below window pops up:
image
Click "Yes"
image

5. So now in the melonDS directory there are these:
image
The user needs to extract the exe from the zip and replace the old one. Finally, we made it!

PS: Do you want to update  melonDS again in the future? You should have kept the token! MelonDS does not save it and github only shows it once when you generate it. So If you didn't kept it, you must generate a new one.

TLDR: It is easier to just download from Github the latest dev version than using the updater. **The user should be able to update with just a few clicks**, not creating accounts, generating tokens and extracting zips.

I really hope I am not being too harsh on this, I am just trying to help :)

WaluigiWare64:
> PS: Do you want to update melonDS again in the future? You should have kept the token! MelonDS does not save it and github only shows it once when you generate it. So If you didn't kept it, you must generate a new one.

I was planning on adding that, but I have been busy the past few days.

KostaSaizo7:
Error when trying to update to stable
image
It creates a 0kb melonDS_0.9.3_win64.7z file in the directory

EndangeredNayla:
nightly.link would bypass the need for a token

PR #732, opened 2020-08-29
add build dependency for Ubuntu: `cmake`

gojun077:
add cmake to list of pkgs to install for building melonDS on Ubuntu...


WaluigiWare64:
Isn't this the same as #724?

gojun077:
my bad, I didn't see that someone else made the exact same PR a few days ago...

PR #734, opened 2020-08-31
Fix "Improved polygon splitting" option in GUI

v1993:
This fixes checkbox not reflecting current state of option on dialogue opening, "Cancel" not reverting change and option not being disabled when SW renderer is used.


v1993:
@Arisotura mind taking a look at this, please? It's a fairly small fix that removes confusion about option status.

PR #757, opened 2020-09-14
Random minor fixes

v1993:
Various fixes for issues found by LGTM static analyzer: https://lgtm.com/projects/g/Arisotura/melonDS/alerts/?mode=tree. Sadly, it uses out-of-date Ubuntu as for now and causes build to fail since introduction of libslirp as dependency, so analysis is not exactly fresh, but I've addressed some of still relevant issues.


v1993:
P.S.: Why not use `#pragma once` instead of header guard? It's implemented by virtually all compilers and reduces problems with header name clashing.

RSDuck:
> P.S.: Why not use #pragma once instead of header guard? It's implemented by virtually all compilers and reduces problems with header name clashing.

because then we would need to change it everywhere 😄

PR #765, opened 2020-09-23
Fix tab order in Qt dialogs

PoroCYon:
So that pressing the tab key will actually move the cursor to the next UI element, instead of having a seemingly random order.


PoroCYon:
Not sure what the issue is with the aarch64 CI.

PR #768, opened 2020-09-25
#pragma once on all melonDS headers

kyandora:
Addresses https://github.com/Arisotura/melonDS/pull/757#issuecomment-692364390.


WaluigiWare64:
Is there an advantage to using `#pragma once` vs normal include guards?

poudink:
the linked comment says it "reduces problems with header name clashing"

PR #769, opened 2020-09-26
Updated flatpak manifest

gal20:
Please consider publishing it to flathub.

### DIFF
--- a/flatpak/net.kuribo64.melonDS.yml
+++ b/flatpak/net.kuribo64.melonDS.yml
@@ -20,9 +20,9 @@ modules:
   - name: libslirp
     buildsystem: meson
     sources:
-      - type: archive
-        url: https://elmarco.fedorapeople.org/libslirp-4.3.1.tar.xz
-        sha256: 388b4b08a8cc0996cc5155cb027a097dc1a7f2cfe84b1121496608ab5366cc48
+      - type: git
+        url: https://gitlab.freedesktop.org/slirp/libslirp
+        tag: "v4.3.1"
 
   - name: melonds
     buildsystem: cmake-ninja
--- a/flatpak/net.kuribo64.melonDS.yml
+++ b/flatpak/net.kuribo64.melonDS.yml
@@ -21,10 +21,8 @@ modules:
     buildsystem: meson
     sources:
       - type: archive
-        url: https://gitlab.freedesktop.org/slirp/libslirp/-/archive/v4.3.1/libslirp-v4.3.1.tar.gz
-        sha256: 500473e4f957f302cd3a9737b6c9f7512ee17d394f646f4223cf103df249cedc
-      - type: patch
-        path: libslirp-meson.patch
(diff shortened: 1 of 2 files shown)


flatpak/net.kuribo64.melonDS.yml
```diff
@@ -11,12 +11,18 @@ finish-args:
...
 modules:
-  - name: libpcap
+  
+  - name: libslirp
+    buildsystem: meson
     sources:
       - type: archive
-        url: http://www.tcpdump.org/release/libpcap-1.9.0.tar.gz
-        sha256: 2edb88808e5913fdaa8e9c1fcaf272e19b2485338742b5074b9fe44d68f37019
+        url: https://gitlab.freedesktop.org/slirp/libslirp/-/archive/v4.2.0/libslirp-v4.2.0.tar.gz
```
WaluigiWare64, line 24: `libslirp` has a newer version available - 4.3.1

gal20: Newer versions fail to build because of the [`version : run_command`](https://gitlab.freedesktop.org/slirp/libslirp/-/blob/v4.3.1/meson.build#L2) in the meson.build file. I tried to replace it using `sed` to get around it, but it didn't work.

gal20: Solved it by creating a patch file, it now works with the latest version of libslirp.

WaluigiWare64: Could you try using this - https://elmarco.fedorapeople.org/libslirp-4.3.1.tar.xz - without the patch?
For some reason the official archive fails for me too but this one works.

nadiaholmquist: Another option is to fetch it from git with a tag for the version instead, that solves the problem of the version being undefined when building.

gal20: Yeah, it works. This is probably because the .tarball-version file isn't generated for some reason, but is already present in the alternative archive.

gal20: @nadiaholmquist thanks! This is a much better solution.

nadiaholmquist:
I'm not quite sure how packaging a flatpak works, but maybe it would be worth it to look into doing flatpak CI? That way we have a generic Linux build people can use without them complaining that they are missing libslirp or whatever.

gal20:
AFAIK you can simply submit the manifest to flathub and they will build it on their servers. Here's their [app submission guide](https://github.com/flathub/flathub/wiki/App-Submission#how-to-submit-an-app).

WaluigiWare64:
Does this require any more work?

gal20:
@WaluigiWare64 No. The flatpak manifest isn't required in this repository at all, since @nadiaholmquist created the pull request to flathub.

WaluigiWare64:
Since melonDS was added to flathub quite a while ago, is this PR still necessary?

gal20:
No. The file can also be removed from this repository, since it doesn't serve any purpose here.

PR #771, opened 2020-09-27
Add support for macOS

WaluigiWare64:
WIP - would appreciate any help.

### To-do list
- [x] Fix ARMJIT_Memory.cpp - malloc.h isn't a header on macOS and fix issues with uc_mcontext
- [x] Fix mapping JIT memory
- [x] Fix SIGBUS crash with fastmem - handle both SIGSEGV and SIGBUS?

- [x] Change OpenGLSupport.h to include macOS headers
- [x] Find a workaround to the `AF_PACKET` which is not on macOS and instead use `AF_LINK`
- [x] Fix keys unable to be binded

#### When it works on macOS
- [x] Build macOS `.app` bundle which includes icon and metadata
- [x] Add instructions for building on macOS
- [x] Add libslirp to brew (Homebrew/homebrew-core#63412)
- [x] Add macOS CI


src/ARMJIT_Memory.cpp
```diff
@@ -628,8 +640,12 @@ void Init()
     FastMem7Start = mmap(NULL, AddrSpaceSize, PROT_NONE, MAP_ANON | MAP_PRIVATE, -1, 0);
 
     MemoryBase = (u8*)mmap(NULL, MemoryTotalSize, PROT_NONE, MAP_ANON | MAP_PRIVATE, -1, 0);
+    #ifdef __APPLE__
+        MemoryFile = shm_open("melondsfastmem", O_RDWR|O_CREAT, 0600);
```
RSDuck, line 644: I think you need to add something like the current pid to make shared mem object unique. Otherwise all melonDS processes will share this => chaos.

WaluigiWare64: Do I add that to the name?

RSDuck: yes

WaluigiWare64: done.

src/ARMJIT_Memory.cpp
```diff
@@ -628,8 +640,12 @@ void Init()
     FastMem7Start = mmap(NULL, AddrSpaceSize, PROT_NONE, MAP_ANON | MAP_PRIVATE, -1, 0);
 
     MemoryBase = (u8*)mmap(NULL, MemoryTotalSize, PROT_NONE, MAP_ANON | MAP_PRIVATE, -1, 0);
+    #ifdef __APPLE__
+        MemoryFile = shm_open("melondsfastmem" + getpid(), O_RDWR|O_CREAT, 0600);
```
RSDuck, line 644: this won't compile. You can't add an integer to a string literal. You need to use something like sprintf

WaluigiWare64: for some reason it doesn't give me any errors with this.

WaluigiWare64:
@RSDuck got it to build but while linking it gives me this
```
[ 73%] Linking CXX executable ../../../melonDS
ld: warning: option -s is obsolete and being ignored
Undefined symbols for architecture x86_64:
  "_ARM_Dispatch", referenced from:
      __ZN5ARMv510ExecuteJITEv in libcore.a(ARM.cpp.o)
      __ZN5ARMv410ExecuteJITEv in libcore.a(ARM.cpp.o)
  "_ARM_Ret", referenced from:
      __ZN6ARMJIT8Compiler27Comp_SpecialBranchBehaviourEb in libcore.a(ARMJIT_Compiler.cpp.o)
      __ZN6ARMJIT8Compiler12CompileBlockEP3ARMbPNS_12FetchedInstrEi in libcore.a(ARMJIT_Compiler.cpp.o)
  "__ZN3NDS10ImportSRAMEPhj", referenced from:
      __ZN8Frontend10ImportSRAMEPKc in Util_ROM.cpp.o
  "__ZN7QThread16createThreadImplEOSt6futureIvE", referenced from:
      __ZN8Platform13Thread_CreateEPFvvE in Platform.cpp.o
  "__ZN9EmuThread22windowFullscreenToggleEv", referenced from:
      __ZN9EmuThread3runEv in main.cpp.o
ld: symbol(s) not found for architecture x86_64
collect2: error: ld returned 1 exit status
make[2]: *** [melonDS] Error 1
make[1]: *** [src/frontend/qt_sdl/CMakeFiles/melonDS.dir/all] Error 2
make: *** [all] Error 2
```
Do you know why?

RSDuck:
the first two are symbols defined in assembler code. Though I don't know what it doesn't like about the other ones.

src/ARMJIT_Memory.cpp
RSDuck: ah yes, this is C++ stupidness. The string literal is from type const char*, so adding an integer to it moves the pointer forward. So say the process id is 2 the string will be "londsfastmem".

WaluigiWare64:
I have no idea how to fix this.

src/ARMJIT_Memory.cpp
WaluigiWare64: Good to know.
Will fix later.

src/ARMJIT_Memory.cpp
```diff
@@ -645,7 +645,7 @@ void Init()
     #ifdef __APPLE__
         char* fastmem_pid_name = new char[snprintf(NULL, 0, "%i", getpid())];
         sprintf(fastmem_pid_name, "%d", getpid());
-        strcat(fastmem_pid_name, "melondsfastmem);
+        strcat(fastmem_pid_name, "melondsfastmem");
```
RSDuck, line 648: the strcat is in both cases unecessary. You just need to change the formatting string to "melondsfastmem%d". Also for consistency change it in both cases to %d

WaluigiWare64: Done.

src/ARMJIT_Memory.cpp
WaluigiWare64: Fixed.

RSDuck:
maybe try disabling the JIT first in the build options?

WaluigiWare64:
I think it's something to do with this line:
`ld: warning: option -s is obsolete and being ignored`
Though I'll try your suggestion.

WaluigiWare64:
Just tried disabling the JIT. Two of the errors have gone.
```
ld: warning: option -s is obsolete and being ignored
Undefined symbols for architecture x86_64:
  "__ZN7QThread16createThreadImplEOSt6futureIvE", referenced from:
      __ZN8Platform13Thread_CreateEPFvvE in Platform.cpp.o
```

RSDuck:
are you building with clang or gcc?

WaluigiWare64:
I'm using gcc.
Also I've just tried using the Apple Clang.
I had to disable the JIT, but this one compiles and opens.
I'm currently trying to see if it'll run any games.

WaluigiWare64:
Now with `clang` we have
```
ld: warning: option -s is obsolete and being ignored
Undefined symbols for architecture x86_64:
  "_ARM_Dispatch", referenced from:
      ARMv5::ExecuteJIT() in libcore.a(ARM.cpp.o)
      ARMv4::ExecuteJIT() in libcore.a(ARM.cpp.o)
  "_ARM_Ret", referenced from:
      ARMJIT::Compiler::Comp_SpecialBranchBehaviour(bool) in libcore.a(ARMJIT_Compiler.cpp.o)
      ARMJIT::Compiler::CompileBlock(ARM*, bool, ARMJIT::FetchedInstr*, int) in libcore.a(ARMJIT_Compiler.cpp.o)
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [melonDS] Error 1
make[1]: *** [src/frontend/qt_sdl/CMakeFiles/melonDS.dir/all] Error 2
make: *** [all] Error 2
```
However, it builds without the JIT.

WaluigiWare64:
I've fixed the issue, for macOS the symbols needed to be prepended with an underscore. Now to squash those commits.
Also, could someone test if OpenGL works with the artifact generated? I can't run it in my VM as it has no 3D acceleration and instead tries to render the OpenGL in software.

src/ARMJIT_Memory.cpp
```diff
@@ -628,8 +640,14 @@ void Init()
     FastMem7Start = mmap(NULL, AddrSpaceSize, PROT_NONE, MAP_ANON | MAP_PRIVATE, -1, 0);
 
     MemoryBase = (u8*)mmap(NULL, MemoryTotalSize, PROT_NONE, MAP_ANON | MAP_PRIVATE, -1, 0);
+    #ifdef __APPLE__
+        char* fastmem_pid_name = new char[snprintf(NULL, 0, "melondsfastmem%d", getpid())];
```
RSDuck, line 644: apparently the value returned by snprintf doesn't take the null terminator into account, so it needs to be increased by one

src/ARMJIT_Memory.cpp
```diff
@@ -663,6 +681,10 @@ void DeInit()
     svcUnmapProcessCodeMemory(envGetOwnProcessHandle(), (u64)MemoryBaseCodeMem, (u64)MemoryBase, MemoryTotalSize);
     virtmemFree(MemoryBaseCodeMem, MemoryTotalSize);
     free(MemoryBase);
+#elif defined(__APPLE__)
+    char* fastmem_pid_name = new char[snprintf(NULL, 0, "melondsfastmem%d", getpid())];
```
RSDuck, line 685: same here

RSDuck:
I pushed the sigbus handling to master, so once the merge conflicts are cleaned up fastmem could work

src/ARMJIT_Memory.cpp
```diff
@@ -701,7 +712,13 @@ void Init()
     FastMem7Start = MemoryBase + AddrSpaceSize;
     MemoryBase = MemoryBase + AddrSpaceSize*2;
 
-    MemoryFile = memfd_create("melondsfastmem", 0);
+    #ifdef __APPLE__
+        char* fastmem_pid_name = new char[snprintf(NULL, 0, "melondsfastmem%d", getpid()) + 1];
```
RSDuck, line 716: just noticed, please make this camelCase instead of snake_case

WaluigiWare64: right, will do.

src/ARMJIT_Memory.cpp
```diff
@@ -701,7 +712,13 @@ void Init()
     FastMem7Start = MemoryBase + AddrSpaceSize;
     MemoryBase = MemoryBase + AddrSpaceSize*2;
 
-    MemoryFile = memfd_create("melondsfastmem", 0);
+    #ifdef __APPLE__
+        char* fastmemPidName = new char[snprintf(NULL, 0, "melondsfastmem%d", getpid()) + 1];
```
RSDuck, line 716: you're leaking memory here

PR #772, opened 2020-09-27
Add Hybrid Layout

### LINKED ISSUE #484: [Feature Request] New layout
I think the "Emphasize top/bottom" layout is a good idea, but why waste all that screen space when we could have something like this:
1920x1080

gal20:
Adds 'Hybrid' layout option: draws a large main window in addition to the top and bottom screens.

Features:
- Choosing 'Swap screens' will change the larger window to be the bottom instead of the top.
- Option to touch the hybrid screen when 'Swap screens' is on.

~~Limitations:~~
~~- Currently ignores rotation. It would be a bit more difficult to implement it (personally I don't see how the rotation feature is usefull so I didn't bother).~~
Added support for rotation.

Example screenshot:
hybrid

Closes #484

### DIFF
--- a/src/frontend/FrontendUtil.h
+++ b/src/frontend/FrontendUtil.h
@@ -116,6 +116,7 @@ void EnableCheats(bool enable);
 //     0 = natural (top screen above bottom screen always)
 //     1 = vertical
 //     2 = horizontal
+//     3 = hybrid
 // * rotation: angle at which the DS screens are presented: 0/1/2/3 = 0/90/180/270
 // * sizing: how the display size is shared between the two screens
 //     0 = even (both screens get same size)
@@ -140,7 +141,7 @@ const int MaxScreenTransforms = 3;
 
 // get a 2x3 transform matrix for each screen and whether it's a top or bottom screen
 // note: the transform assumes an origin point at the top left of the display,
-// X going left and Y going down
+// X going right and Y going down
 // for each screen the source coordinates should be (0,0) and (256,192)
 // 'out' should point to an array of 6*MaxScreenTransforms floats
 // 'kind' should point to an array of MaxScreenTransforms ints
--- a/src/frontend/Util_Video.cpp
+++ b/src/frontend/Util_Video.cpp
@@ -30,9 +30,13 @@ namespace Frontend
 
 float TopScreenMtx[6];
 float BotScreenMtx[6];
+float HybScreenMtx[6];
 float TouchMtx[6];
+float HybTouchMtx[6];
 bool TopEnable;
 bool BotEnable;
+bool HybEnable;
+int HybScreen;
 
 void M23_Identity(float* m)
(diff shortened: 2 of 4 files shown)


PoroCYon:
> personally I don't see how the rotation feature is usefull so I didn't bother

brain training

gal20:
Added support for rotation with hybrid layout

gal20:
Sorry about that indentation nightmare, I'm trying to fix it.

Anuskuss:
Nice job, @gal20. This fixes #484 (although I used a slightly different ratio). Fullscreen doesn't seem to recalculate right now and Screen Rotation rotates the whole thing, but I'm sure you'll figure it out. One small feature request from me: Support for Hotkeys (second bullet point: https://github.com/Arisotura/melonDS/issues/347#issuecomment-483509236). I mainly play in fullscreen so it would be nice if I could rotate the screens / switch the main with my controller (bonus points if you can teach melonDS to remember that per-game 😄).

gal20:
Pushed a new branch to comply with d529b650c06a4dd43904dfead49011be1ba9db9a. Now the `ScreenSwap` option is used to switch the larger window from top to bottom.

gal20:
> Nice job, @gal20. This fixes #484 (although I used a slightly different ratio). ~Fullscreen doesn't seem to recalculate right now~ and Screen Rotation rotates the whole thing, but I'm sure you'll figure it out. One small feature request from me: Support for Hotkeys (second bullet point: [#347 (comment)](https://github.com/Arisotura/melonDS/issues/347#issuecomment-483509236)). I mainly play in fullscreen so it would be nice if I could rotate the screens / switch the main with my controller (bonus points if you can teach melonDS to remember that per-game smile).

#953 adds a hotkey for swapping the main window

Anuskuss:
Awesome, thank you! Do you have any plans on changing the rotated layout? Because I don't really think it's useful in this state, or rather in a hybrid layout at all. If a game requires a rotated DS, I'd rather use the horizontal view with even screens (makes more sense on a widescreen). IMO you should drop it completely and if you want to introduce another hotkey (similar to DeSmuME) to cycle through hybrid view (0°, 180°) and rotated, horizontal view (90°, 270°).

RSDuck:
in https://github.com/Arisotura/melonDS/commit/b9a56bc4e47bbb315f88ba5eaa57ce2b3efece73 I made some changes to allow displaying only one screen which make this PR less invasive in the display code. Can you adapt to them, I think we can merge this PR then.

gal20:
> in [b9a56bc](https://github.com/Arisotura/melonDS/commit/b9a56bc4e47bbb315f88ba5eaa57ce2b3efece73) I made some changes to allow displaying only one screen which make this PR less invasive in the display code. Can you adapt to them, I think we can merge this PR then.

Done.

PR #778, opened 2020-10-03
Generate a simple non-bootable firmware when not provided.

Swordfish90:
Hello guys,

I'd really like to help with the HLE bios task #367. It seems the drastic bios file linked in the issue are working very well, but sadly the games won't work without a proper firmware.

With this pull request, I'm trying to salvage some work done in #527, generating a dummy firmware when none are provided.

Using the free drastic bios files (linked in #367) I've been able to play every game I tried, so it sounds like a promising start.

Let me know if you want me to change anything.

PR #779, opened 2020-10-04
Firmware editor

kyandora:
Addresses https://github.com/Arisotura/melonDS/pull/778#issuecomment-703165075.

### DIFF
--- a/src/SPI.h
+++ b/src/SPI.h
@@ -24,6 +24,7 @@
 namespace SPI_Firmware
 {
 
+u16 CRC16(u8* data, u32 len, u32 start);
 void SetupDirectBoot();
 
 u8 GetConsoleType();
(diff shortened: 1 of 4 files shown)


WaluigiWare64:
Is this not the same as #778?

kyandora:
We both started at the same time, but ended with different approaches.
If no firmware is present, @Swordfish90 creates an empty dummy firmware in memory and populates it's user settings.
My pull request updates an already existing firmware file.

kyandora:
@RSDuck @Swordfish90 What would you think about a hybrid approach with the user settings as an overlay on top of either the dummy or firmware? melonDS already does that with the touchscreen calibration data.

RSDuck:
#778 already does that mostly

kyandora:
In that case I'll close this and push all reusable parts to #778.

Swordfish90:
@kyandora Yeah, they look very close. My main intent was making melonDS useable without a firmware (and using the free bios). If you want to submit the pull request with the UI on my fork, I'll merge them into one with both the in-memory firmware and the configuration (which in your case seems a bit more advanced).

PR #783, opened 2020-10-07
Fix compilation issues on pedantic cpp compilers.

Swordfish90:
I've built melonDS for Android (the libretro core actually), and I found a couple of issues which most pedantic compilers mark as errors.

I also have an Android specific fix to avoid using memfd_create (which is only available from Android 11) in FastMem. If you're interested I'll upstream it, otherwise I'll try to push it into the libretro repository.

### DIFF
--- a/src/ARMJIT_A64/ARMJIT_Compiler.h
+++ b/src/ARMJIT_A64/ARMJIT_Compiler.h
@@ -187,6 +187,9 @@ public:
     void Comp_RegShiftReg(int op, bool S, Op2& op2, Arm64Gen::ARM64Reg rs);
 
     bool Comp_MemLoadLiteral(int size, bool signExtend, int rd, u32 addr);
+
+    static void UpdateMode(ARM* arm, u32 oldmode, u32 newmode);
+
     enum
     {
         memop_Writeback = 1 << 0,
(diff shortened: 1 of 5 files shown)


src/ARMJIT_A64/ARMJIT_Compiler.h
```diff
@@ -187,6 +187,9 @@ class Compiler : public Arm64Gen::ARM64XEmitter
     void Comp_RegShiftReg(int op, bool S, Op2& op2, Arm64Gen::ARM64Reg rs);
 
     bool Comp_MemLoadLiteral(int size, bool signExtend, int rd, u32 addr);
+
+    static void UpdateMode(ARM* arm, u32 oldmode, u32 newmode);
```
RSDuck, line 191: make this just a normal function in ARMJIT_Compiler.cpp. Name it something like TrampolineARMUpdateMode

RSDuck:
> I also have an Android specific fix to avoid using memfd_create (which is only available from Android 11) in FastMem. If you're interested I'll upstream it, otherwise I'll try to push it into the libretro repository.

yes please

Swordfish90:
Done. Great, for the Android changes I'll test them thoroughly in the next few days and I'll send a pull request.

WaluigiWare64:
This would also fix the issues using clang on my macOS PR (#771) because for some reason gcc can't build the final executable.

WaluigiWare64:
@Swordfish90 does anything else need to be done or is this ready to merge?

Swordfish90:
@WaluigiWare64 Nope, it's ready to be merged.

Issue #367, opened 2019-01-16
Add high level bios emulation


TheCoder3156:
Would it be possible to add this? It’s not possible for some people to get the bios9 and 7 for people with a dsi without pirating. Thanks!

Arisotura:
yeah that'd be a good thing to add

xperia64:
I would like to point out that DraStic ships a free drop-in BIOS replacement under a fairly liberal license. It isn't perfect, and obviously doesn't include any keys, but I've tested it with MelonDS 0.8.2, and it seems to be enough to boot games fine.

These do not replace firmware.bin, and there's obviously some weirdness if a blank firmware.bin is provided (such as touch not working due to lack of calibration), but games boot otherwise without the need of real bios dumps.

[drastic_bios.zip](https://github.com/Arisotura/melonDS/files/3366514/drastic_bios.zip)

Arisotura:
generating a firmware would be easy enough (but it wouldn't be bootable)

not sure about the BIOS thing, I need to audit the DraStic BIOS. I don't want to end up with a billion bug reports stemming from a buggy BIOS replacement.

TheCoder3156:
Maybe we could get desmume to relicense their bios code under GPLv2 or later so we could use it.

TheCoder3156:
There are only two contributors so it might be easy to get permission.

Arisotura:
DeSmuME doesn't have a full-fledged alternate BIOS. they set up a minimalistic IRQ handler, and they handle the SWI calls via HLE.

which has its set of issues (for example the decompression functions that use callbacks, ...)

FrostKnight:
If it really is licensed under a free license, then its good till you get a better implementation.
Still a bit puzzled as to what website you found it in, but meh... w/e 
Hope you guys are working on an alternate bios though in the future. 
And yeah, you don't want a buggy replacement. Wisdom! :)

WaluigiWare64:
wrong PR :P

FrostKnight:
whoops... my bad.

Arisotura:
done!

PR #787, opened 2020-10-08
Fallback to FreeBIOS when BIOS files are not found.

Swordfish90:
This pull request, together with #778 makes melonDS no longer reliant on external files, effectively closing #367 .

I've tested the free BIOSes on dozens of games these days and each one of them worked correctly.

This is the free BIOS that's bundled with Drastic, which has more than 1M downloads on Google Play so it should be quite battle-tested.


WaluigiWare64:
Could you make an option like "Use FreeBIOS instead of external BIOS and firmware"?
Let's say I want to test the FreeBIOS, but I don't want to remove my DS config and then re add them later.

g4jc:
I apologize if this is an obvious question, but is FreeBIOS really free? It appears to be a hex blob, but I was unable to find the original sources. In fact even the APK for Drastic cost $4.99 and there is no source code for it either.

nadiaholmquist:
Yes it is actually free (BSD license if I remember right) but I think it's only distributed in the DraStic APK. Might be a good idea to throw that source up on GitHub too.

Swordfish90:
@g4jc @nadiaholmquist You can find the source code (with the license) in the referenced GitHub issue, but I'll also include it within this pull request later today just to be on the safe side.

FrostKnight:
@Swordfish90 so it is legal? That's good, Don't want a good ds emulator like melonds dying.

Swordfish90:
@FrostKnight I understand the concearns, I'd also really like to see melonDS flourish. 

I don't think there is a problem with Nintendo. We have the source code of this project and Drastic, which is a paid application on Google Play with more than 1M downloads hasn't received a DMCA for it.

Drastic is also bundling the binary and the source code together with their APK. The license is pretty permissive and clear (it's included in the header file in this pull request).

IMHO only a minority of users today are using a legal bios file dumped from their own console, so I believe this is a step towards legal usage.

FrostKnight:
Can I see the source code? I am curious what website still has it up.

xperia64:
> 
> 
> Can I see the source code? I am curious what website still has it up.

https://github.com/Arisotura/melonDS/issues/367#issuecomment-509061317

Swordfish90:
Hi guys. @g4jc @nadiaholmquist I added the bios source and binary files to the pull request.

Feel free to let me know if you need anything else on my side.

SuuperW:
Is there anything else that should be done before merging this and #778? I'd like to see that happen.

poudink:
what does this do that https://github.com/Arisotura/melonDS/pull/778/ doesn't do?

RSDuck:
> what does this do that #778 doesn't do?

both PRs are complementary. This one's for bios files, the other one for the firmware.

Swordfish90:
To give you a short update, these two pull requests have been battle-tested in Lemuroid (in production) for the last two months (where you can't even specify a different BIOS file).

Just let me know what's your take on this, I can fix the merge conflicts, or I won't hold a grudge if you decide not to keep it.

RSDuck:
is FreeBIOS really the name of the bios replacement? It seems to appear nowhere in it's files. Otherwise I think we can merge this.

Swordfish90:
@RSDuck I don't think it has a proper name, so I went with the most obvious "FreeBIOS". We can call it "Drastic BIOS" or "Custom BIOS replacement" or any combination of the previous. Which one would you suggest?

RSDuck:
call it Drastic BIOS, that's a pretty unambigious name.

WaluigiWare64:
I think that instead of the BIOS replacement being loaded automatically when BIOS files are not provided, there should be an option.
image
(quick sketch in paint)

Also should the BIOS replacement files be in a folder under the root or src?

The makefile is unnecessary, so that can be removed.

Arisotura:
I'd rather call it 'internal BIOS', 'BIOS replacement' makes it sound like it's meant to replace something, and imo sounds less intuitive to the average joe

nadiaholmquist:
Agreed, that or "built-in BIOS". There should probably also be a warning under the box that it's not entirely accurate to the original DS BIOS, and that using the official BIOS is recommended if you can get a hold of it.

WaluigiWare64:
We don't really need to call it something, seeing as we all have different opinions on this:
image

Parakoopa:
I like the solution by @WaluigiWare64 best. Just have a checkbox that when unchecked greys out the other controls and uses the Drastic BIOS, and this should be the default.

vadosnaprimer:
Did https://github.com/Arisotura/melonDS/pull/1174 resolve this PR?

RayyanAnsari:
merged in #1174

PR #790, opened 2020-10-14
Check for updates from within melonDS

### LINKED ISSUE #469: [Feature Request] Update Notice?
Hello,
First of all thanks for all the work with the emulator, it works great!
However, I notice the lack of "Updater" option, or some sort of notice there's a new update available (Granted, I started using from the current 0.8.1 version, so I might be wrong and this feature is already implemented or something similar is already in place). Hopefully it wouldn't be too hard to implement, and, for example, when launching emulator, it could check for a new version of the application, and show a popup window saying something like this:
"There's a new update available, version {insert_new_version_number_here}. You have version {insert_current_version_number_here}. Would you like to visit the website and download a new version?".
Just thought that would be a small nice feature to have, if possible - if not, it's not a big deal. :)

abcdjdj:
Stuff in this PR - 

1. Check for updates manually under About > Check for Updates. 
image

2. Automatically check for updates on start under Config > Emu Settings > General
image
 
When an update is available, the new and current version numbers are displayed along with a Yes/No option to open the download page (http://melonds.kuribo64.net/downloads.php) in the browser.

image

When no update is available, there are 2 cases - if the user had manually checked for an update, then an info box is shown. If the update check was automatically triggered at boot, then no dialog box is shown and the user can peacefully continue working.

image

Fixes #469

### DIFF
--- a/src/frontend/qt_sdl/CMakeLists.txt
+++ b/src/frontend/qt_sdl/CMakeLists.txt
@@ -39,7 +39,7 @@ if (BUILD_STATIC AND QT5_STATIC_DIR)
     set(Qt5Widgets_DIR ${QT5_STATIC_BASE}Widgets)
 endif()
 
-find_package(Qt5 COMPONENTS Core Gui Widgets REQUIRED)
+find_package(Qt5 COMPONENTS Core Gui Widgets Network REQUIRED)
 
 set(CMAKE_AUTOMOC ON)
 set(CMAKE_AUTOUIC ON)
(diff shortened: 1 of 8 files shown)


abcdjdj:
PS : I haven't tested it on windows. If someone could give it a try on windows, I'd be grateful :)

nadiaholmquist:
Would be good to have a build-time option to disable the updater, for example for when melonDS is installed through a distro package or flatpak.

abcdjdj:
> 
> 
> Would be good to have a build-time option to disable the updater, for example for when melonDS is installed through a distro package or flatpak.

This PR only notifies the user about the availability of a new update. It does not actually download the new version. Sorry for the confusion.

In fact, flatpak and distro package users will also benefit from this as they can inform their package maintainers about the update asap, whenever it drops.

nathanielcwm:
I think #729 can also fetch the workflow artifacts.

abcdjdj:
> 
> 
> I think #729 can also fetch the workflow artifacts.

Oops. Didn't see that PR. Closing this one.

PR #798, opened 2020-10-29
Added 8/16-bit IPCFIFOSEND writes

Hypnotron:
Hardware tests (at least on my 3DS) have confirmed that writing to IPCFIFOSEND using STRB or STRH mirrors the value into a full word and sends it to the FIFO. I have not tested 8/16-bit reads from IPCFIFORECV yet. Attached is the test I wrote (which should beep eleven times if 8-bit mirroring is correct).
[fifo.zip](https://github.com/Arisotura/melonDS/files/5461033/fifo.zip)

### DIFF
--- a/src/NDS.cpp
+++ b/src/NDS.cpp
@@ -3116,6 +3116,10 @@ void ARM9IOWrite8(u32 addr, u8 val)
         NDSCart::WriteSPIData(val);
         return;
 
+    case 0x04000188:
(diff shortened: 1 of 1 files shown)


Arisotura:
thanks!

btw, quick tip for mirroring a 8-bit value to 32-bit:

val32 = val8 * 0x01010101;

also works for 16-bit or whatever

PR #806, opened 2020-11-08
Revert 1 addons

Buenia0:
I'm reverting this merging because said pull request resulted in compilation errors. Sorry about that.


WaluigiWare64:
Could you not revert it on your own fork instead of creating a new PR?

PR #816, opened 2020-11-14
Use ashmem instead of memfd_create on Android.

Swordfish90:
@RSDuck as promised, here's the fix for memfd_create on Android.

memfd_create is only available on Android 11, so only a very small percentage of devices can use it.

Ashmem has a similar purpose and it's what PPSSPP and Dolphin are using in these instances. 

I didn't push this before because sometimes the application hanged with fastmem enabled, but your latest changes on master fixed these issues.


src/ARMJIT_Memory.cpp
```diff
@@ -701,8 +711,24 @@ void Init()
     FastMem7Start = MemoryBase + AddrSpaceSize;
     MemoryBase = MemoryBase + AddrSpaceSize*2;
 
+#if defined(__ANDROID__)
+    static void* libandroid = dlopen("libandroid.so", RTLD_LAZY | RTLD_LOCAL);
+    using type_ASharedMemory_create = int(*)(const char *name, size_t size);
+    auto symbol = dlsym(libandroid, "ASharedMemory_create");
+    static auto shared_memory_create =reinterpret_cast<type_ASharedMemory_create>(symbol);
+
+    if (shared_memory_create) {
```
RSDuck, line 720: please but the braces always on a new line

src/ARMJIT_Memory.cpp
```diff
@@ -701,8 +711,24 @@ void Init()
     FastMem7Start = MemoryBase + AddrSpaceSize;
     MemoryBase = MemoryBase + AddrSpaceSize*2;
 
+#if defined(__ANDROID__)
+    static void* libandroid = dlopen("libandroid.so", RTLD_LAZY | RTLD_LOCAL);
+    using type_ASharedMemory_create = int(*)(const char *name, size_t size);
+    auto symbol = dlsym(libandroid, "ASharedMemory_create");
```
RSDuck, line 717: dlsym just returns void*, auto isn't necessary here

src/ARMJIT_Memory.cpp
```diff
@@ -701,8 +711,24 @@ void Init()
...
+    static auto shared_memory_create =reinterpret_cast<type_ASharedMemory_create>(symbol);
```
RSDuck, line 718: I think some spaces are missing

RSDuck (changes requested):

RSDuck:
if ASharedMemory_create creates a shared memory object similar to shm_open isn't it shared among all processes? Isn't this problematic if multiple instances of melonDS are running?

Also clean up code seems to be missing as well.

WaluigiWare64:
You might have to do the same thing I did here - https://github.com/Arisotura/melonDS/blob/325818e7c7542cf00a6f4b7979ca3ed62bea221f/src/ARMJIT_Memory.cpp#L716-L718

Swordfish90:
@RSDuck @WaluigiWare64 Sorry for the very late reply, but it has been a hectic weekend. I just pushed a commit addressing the code style issues.

Ashmem on Android behaves a little differently. For starters the name is not really meaningful, it is used mainly for debugging purposes. If two processes ask a region with the same name, they'll get two separate file descriptors. It can be used for IPC, but only by sharing the appropriate file descriptors.

The lack of cleanup code is also correct. Memory will be released when closing the file descriptor (which is already done in other implementations). If you want to dig deeper you can find more info [here](https://elinux.org/Android_Kernel_Features) and [here](https://developer.android.com/ndk/reference/group/memory).

PR #836, opened 2020-12-01
Rename jit linkage asm files

webgeek1234:
An extension of lower case s indicates to not run the preprocessor while upper case S does. See
https://www.keil.com/support/man/docs/armclang_intro/armclang_intro_chr1383143887630.htm

I do Android infrastructure work for libretro and have been working to get melonDS compiling and running on Android. For some reason, the assembler-with-cpp flag doesn't seem to be working with the Android NDK. But renaming the files per the above spec does work. Offering this upstream to simplify downstream rebases in the furture.


RSDuck:
who ever came up with the idea that there should be a difference between .S and .s

PR #849, opened 2020-12-12
Add a custom logging function - allows logging to file

RayyanAnsari:
This PR creates a custom logging function and gives an option to either log to console, or to log to file.
### Changes

- All `printf` calls are now `Platform::LogMessage`
- A new settings dialog has been created: Log Settings
- Two new variables are added to the config file: `LogToFile` and `LogFileLocation`
- The new `Platform::LogMessage` function uses either `vprintf` or `vfprintf` depending on the user's setting

### DIFF
--- a/src/frontend/qt_sdl/LogSettingsDialog.cpp
+++ b/src/frontend/qt_sdl/LogSettingsDialog.cpp
@@ -1,0 +1,91 @@
+/*
+    Copyright 2016-2020 Arisotura, WaluigiWare64
+
+    This file is part of melonDS.
+
+    melonDS is free software: you can redistribute it and/or modify it under
+    the terms of the GNU General Public License as published by the Free
+    Software Foundation, either version 3 of the License, or (at your option)
+    any later version.
+
+    melonDS is distributed in the hope that it will be useful, but WITHOUT ANY
+    WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+    FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License along
+    with melonDS. If not, see http://www.gnu.org/licenses/.
+*/
+
+#include <QMessageBox>
+#include <QFileDialog>
+
+#include "Platform.h"
+#include "Config.h"
+#include "PlatformConfig.h"
+
+#include "LogSettingsDialog.h"
+#include "ui_LogSettingsDialog.h"
+
+LogSettingsDialog* LogSettingsDialog::currentDlg = nullptr;
+
+extern char* EmuDirectory;
+
+LogSettingsDialog::LogSettingsDialog(QWidget* parent) : QDialog(parent), ui(new Ui::LogSettingsDialog)
+{
+    ui->setupUi(this);
+    setAttribute(Qt::WA_DeleteOnClose);
+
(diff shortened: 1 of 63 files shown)


src/frontend/qt_sdl/LogSettingsDialog.cpp
```diff
@@ -0,0 +1,91 @@
...
+                                                "Select Log File...",
+                                                EmuDirectory,
+                                                "Log files (*.log *.txt);;Any file (*.*)");
+
+    if (file.isEmpty()) return;
+
+    ui->txtLogFolderPath->setText(file);
+}
+
+void LogSettingsDialog::updateAdapterControls()
```
RSDuck, line 87: the name seems to be a left over

src/frontend/qt_sdl/Platform.cpp
```diff
@@ -127,6 +131,24 @@ void StopEmu()
...
 
+void LogMessage(const char* format, ...)
+{
+    va_list args;
+    va_start(args, format);
+    if (Config::LogToFile)
+    {
+        if (!LogFile)
+        {
+            LogFile = fopen(Config::LogFileLocation, "a+");
```
RSDuck, line 142: a+ seems to be dangerous, because melonDS can spit out a lot of data

WaluigiWare64: should i change to w+?

src/frontend/qt_sdl/LogSettingsDialog.cpp
WaluigiWare64: ah yes, I copied it over from WifiSettingsDialog.cpp.

src/frontend/qt_sdl/Platform.cpp
RSDuck: only w we don't need to read

WaluigiWare64: done

src/frontend/qt_sdl/Platform.cpp
```diff
@@ -127,6 +131,24 @@ void StopEmu()
...
+        {
+            LogFile = fopen(Config::LogFileLocation, "w");
+        }
+        vfprintf(LogFile, format, args);
+    }
+    else
+    {
+        vprintf(format, args);       
+    }
+     va_end(args);
```
RSDuck, line 150: va_end isn't properly aligned

WaluigiWare64: done.

src/sha1/sha1.c
```diff
@@ -267,10 +267,10 @@ main(int argc, char **argv)
         SHA1Update(&ctx, buf, BUFSIZE);
     SHA1Final(hash, &ctx);
 
-    Platform::LogMessage("SHA1=");
```
RSDuck: better not poke in external dependencies. Those are tests anyway, they are never run.

WaluigiWare64: Yep, which is why I changed it back to printf.

nadiaholmquist:
If we're doing a custom logging thing, we really should also have log levels I think?

WaluigiWare64:
We could, but currently this is just a wrapper around printf and fprintf. That would require modifying all of the printf statements we have.

RayyanAnsari:
superseded by #1642

PR #865, opened 2020-12-20
Add support for building with Qt6

nadiaholmquist:
This PR adds support for building using the newly released Qt6, a change or two was needed to replace some deprecated stuff with the new version, but that's all backwards compatible and doesn't break Qt5.

Currently this is only tested on Linux, but everything seems to work fine here. More tested needed before this can be merged, and maybe also CI with Qt6?


WaluigiWare64:
I don't think we should CI with Qt 6 yet.
A few platforms may not have it already in their repositories, and melonDS works just fine with Qt 5 anyway.

nadiaholmquist:
Yeah if we do CI it should build with both Qt5 and Qt6, since Qt5 won't be going anywhere on some platforms for quite a while, but having CI for both if this is merged would be good to ensure that all UI changes are forwards-compatible.

PR #876, opened 2020-12-21
Fixed fastforward toggle hotkey.

spikensbror:
This PR fixes the broken fastforward toggle hotkey, only tested on Windows so it might be a Windows-only issue. The state is not persisted in the configuration file.

### DIFF
--- a/src/frontend/qt_sdl/main.cpp
+++ b/src/frontend/qt_sdl/main.cpp
@@ -266,7 +266,7 @@ EmuThread::EmuThread(QObject* parent) : QThread(parent)
     connect(this, SIGNAL(windowEmuStop()), mainWindow, SLOT(onEmuStop()));
     connect(this, SIGNAL(windowEmuPause()), mainWindow->actPause, SLOT(trigger()));
     connect(this, SIGNAL(windowEmuReset()), mainWindow->actReset, SLOT(trigger()));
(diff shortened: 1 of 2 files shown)


RSDuck:
I thought I already fixed that in https://github.com/Arisotura/melonDS/commit/9673659db46dfbb3b2963b7acd0d68d58fb1c7fd ?

poudink:
Nope. Just tested. It's not working with the latest builds. This PR fixes it.

abcdjdj:
I can also confirm that this PR fixes the fastforward toggle hotkey. It might not be entirely relevant to this PR, but the "Limit Framerate" checkbox under Config is still broken for me. Lastly, we might wanna have the checkbox's state in-sync with the hotkey state for UI consistency.

spikensbror:
@abcdjdj I think the fast-forward and "limiting" features are far too contrived as they are currently implemented. What this PR does currently is basically toggle the fast-forward (disable limit FPS and audio-sync).

In my opinion, it would make more sense to rename the hotkey to "toggle fastforward" and have the "limiting" features completely separate.

The alternative to that is to change this PR to only affect the FPS limiter which would mean that it would have no effect if audio-sync is on. We would also as you mention have to synchronize both config and checkbox state in the menu in that case which this PR currently does not offer.

What is the preferred course of action here? I'll update the PR accordingly

abcdjdj:
I see what you mean. We should probably consult @RSDuck / @Arisotura on the preferred course of action.

PR #890, opened 2020-12-23
Fix JIT linkage on ARM64 Macs

WaluigiWare64:
This fix has been done for x86-64 Macs, but not the new ARM64 Macs.
In the Mach-O format (used in macOS) global function names in assembly must be prepended with an underscore.


StLouisCPhT:
This does not work with the ARM64 Mac builds. It causes an immediate crash of the application.

PR #911, opened 2020-12-28
Fix the ARM64 Code Emitter on macOS

WaluigiWare64:
The JIT Recompiler's Code Emitter tries to clear the CPU's instruction cache in assembly which fails on ARM64 Macs. However, macOS does have a kernel specific function (also in iOS). The iOS icache clearing code was changed to run on all Apple systems.


StLouisCPhT:
At work at the moment, will recompile and test when I get home.

StLouisCPhT:
LGTM, melonDS no longer crashes when trying to start app.

Issue #922, opened 2021-01-02
Save Hotkeys don't work in Fullscreen


bobhob314:

RSDuck:
fixed by #924

PR #924, opened 2021-01-03
Set menubar height to 0 in fullscreen

abcdjdj:
* Avoid using hide() on the mainwindow's menubar as it breaks menubar
actions.

* Fixes save/load state shortcuts not working in fullscreen (#922)

Signed-off-by: Madhav Kanbur <abcdjdj@gmail.com>

### DIFF
--- a/src/frontend/qt_sdl/main.cpp
+++ b/src/frontend/qt_sdl/main.cpp
@@ -1991,12 +1991,12 @@ void MainWindow::onFullscreenToggled()
     if (!mainWindow->isFullScreen())
     {
         mainWindow->showFullScreen();
-        mainWindow->menuBar()->hide();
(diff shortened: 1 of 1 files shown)


WaluigiWare64:
Oh, I see.
I never thought hiding the menu bar would cause it to stop accepting hotkeys.

poudink:
hotkeys still work, just not the savestate ones

abcdjdj:
Yup, poudink's right. It's only the QActions that were added to the menubar which stop functioning. Hotkeys are unaffected. Here's a relevant thread on the QT Forums, for interested readers - [https://forum.qt.io/topic/5039/problem-with-qaction-s-shortcut-and-qmenu](https://forum.qt.io/topic/5039/problem-with-qaction-s-shortcut-and-qmenu)

PR #930, opened 2021-01-08
Polish up archive support

abcdjdj:
~~1. Fix directory path when extracting from archive~~
~~1.1.  Don't create new dir in execution dir of melonds~~
~~1.2.  Create it beside the archive instead~~

2. ArchiveUtil : Use QT functions for I/O
~~2.1.  Fixes permission related crash on linux~~
2.2.  Cleaner, platform independent

TODO:

- [x] Extract ROMs directly to RAM
- [x] Integrate archive support with Recent Files menu
- [x] Ensure GBA ROMs also load from archives
- [x] Fix Frontend stuff (Reset, drag-n-drop etc.)
- [x] Error handling
- [x] Save state file names when loading from archives

### DIFF
--- a/src/GBACart.cpp
+++ b/src/GBACart.cpp
@@ -620,6 +620,32 @@ void DoSavestate(Savestate* file)
     if (HasSolarSensor) GBACart_SolarSensor::DoSavestate(file);
 }
 
+static void LoadROMCommon(const char *sram)
+{
+    char gamecode[5] = { '\0' };
+    memcpy(&gamecode, CartROM + 0xAC, 4);
+    printf("Game code: %s\n", gamecode);
+
+    for (int i = 0; i < sizeof(SOLAR_SENSOR_GAMECODES)/sizeof(SOLAR_SENSOR_GAMECODES[0]); i++)
+    {
+        if (strcmp(gamecode, SOLAR_SENSOR_GAMECODES[i]) == 0) HasSolarSensor = true;
+    }
+
+    if (HasSolarSensor)
+    {
+        printf("GBA solar sensor support detected!\n");
+    }
+
+    CartCRC = CRC32(CartROM, CartROMSize);
+    printf("ROM CRC32: %08X\n", CartCRC);
+
+    CartInserted = true;
+
+    // save
+    printf("Save file: %s\n", sram);
+    GBACart_SRAM::LoadSave(sram);
+}
+
 bool LoadROM(const char* path, const char* sram)
 {
     FILE* f = Platform::OpenFile(path, "rb");
@@ -640,36 +666,13 @@ bool LoadROM(const char* path, const char* sram)
     while (CartROMSize < len)
         CartROMSize <<= 1;
 
-    char gamecode[5] = { '\0' };
-    fseek(f, 0xAC, SEEK_SET);
-    fread(&gamecode, 1, 4, f);
-    printf("Game code: %s\n", gamecode);
-
-    for (int i = 0; i < sizeof(SOLAR_SENSOR_GAMECODES)/sizeof(SOLAR_SENSOR_GAMECODES[0]); i++)
-    {
-        if (strcmp(gamecode, SOLAR_SENSOR_GAMECODES[i]) == 0) HasSolarSensor = true;
-    }
-
-    if (HasSolarSensor)
-    {
-        printf("GBA solar sensor support detected!\n");
-    }
-
     CartROM = new u8[CartROMSize];
     memset(CartROM, 0, CartROMSize);
     fseek(f, 0, SEEK_SET);
     fread(CartROM, 1, len, f);
-
     fclose(f);
 
-    CartCRC = CRC32(CartROM, CartROMSize);
-    printf("ROM CRC32: %08X\n", CartCRC);
-
-    CartInserted = true;
-
-    // save
-    printf("Save file: %s\n", sram);
-    GBACart_SRAM::LoadSave(sram);
+    LoadROMCommon(sram);
 
     return true;
 }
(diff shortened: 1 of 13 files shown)


RSDuck:
just fyi #678 was merged too early, because ideally the roms should be kept in memory and not make the detour to file, so the file stuff is to be changed.

abcdjdj:
> 
> 
> just fyi #678 was merged too early, because ideally the roms should be kept in memory and not make the detour to file, so the file stuff is to be changed.

Yup ik. I sent in those 2 quick fixes in case a new release was being planned. However, if a new version won't drop in the next few days, I'll just leave this pr as a draft and implement the memory version. I hope you won't mind @WaluigiWare64

WaluigiWare64:
Thanks for the fixes @abcdjdj 
I should have tested more :P

src/frontend/qt_sdl/main.cpp
```diff
@@ -1573,28 +1611,31 @@ void MainWindow::onOpenFileArchive()
...
+    QString romFileName; // file name inside archive
+
     if (archiveROMList.size() > 2)
     {
         archiveROMList.removeFirst();
         QString toLoad = QInputDialog::getItem(this, "melonDS",
                                   "The archive was found to have multiple files. Select which ROM you want to load.", archiveROMList.toList(), 0, false);
         printf("Extracting '%s'\n", toLoad.toUtf8().constData());
-        QVector<QString> extractResult = Archive::ExtractFileFromArchive(filename.toUtf8().constData(), toLoad.toUtf8().constData());
+        QVector<QString> extractResult = Archive::ExtractFileFromArchive(archiveFileName.toUtf8().constData(), toLoad.toUtf8().constData(), romBuffer);
```
WaluigiWare64, line 1635: I don't know why I converted the QString results to const char*. It is only converted back to QString anyway.
```suggestion
        QVector<QString> extractResult = Archive::ExtractFileFromArchive(archiveFileName, toLoad, romBuffer);
```

abcdjdj:
Alrighty! LGTM, I've tested most scenarios and it seems to work. Hope this PR doesn't cause any regressions 🤞

WaluigiWare64:
Nice!
I'm going to test this.

src/GBACart.cpp
```diff
@@ -620,6 +620,32 @@ void DoSavestate(Savestate* file)
     if (HasSolarSensor) GBACart_SolarSensor::DoSavestate(file);
 }
 
+static void LoadROMCommon(const char *sram)
```
RSDuck, line 623: we usually don't use static since we use namespaces

src/NDSCart.cpp
```diff
@@ -885,46 +885,15 @@ void DecryptSecureArea(u8* out)
     }
 }
 
-
-bool LoadROM(const char* path, const char* sram, bool direct)
+static bool LoadROMCommon(u32 filelength, const char *sram, bool direct)
```
RSDuck, line 888: static again

src/frontend/Util_ROM.cpp
```diff
@@ -283,6 +287,85 @@ int LoadBIOS()
     return Load_OK;
 }
 
+int LoadROM(const QByteArray *romdata, QString archivefilename, QString romfilename, int slot)
```
RSDuck, line 290: please keep the QT specific stuff outside of FrontendUtil because it's used by other frontends as well

RSDuck (changes requested):
@WaluigiWare64 you're lucky @abcdjdj sorted your mess out :P

WaluigiWare64:
> 
> 
> @WaluigiWare64 you're lucky @abcdjdj sorted your mess out :P

oi :P

abcdjdj:
Ok so I have cleaned up Util_ROM, moved all QT stuff up by one layer (into MainWindow). As an additional bonus, ROMs inside archives that contain > 1 rom will also get loaded automatically during resets. The ROM picker will not be shown (it was earlier). 😄

src/frontend/Util_ROM.cpp
```diff
@@ -19,6 +19,7 @@
 #include <stdio.h>
 #include <string.h>
 
+#include "ArchiveUtil.h"
```
RSDuck, line 22: can you put the archieve code in frontend util behind an ifdef?

RSDuck (approved):
do you have anything to add? Otherwise I think we can merge this once my final points are through.

src/frontend/Util_ROM.cpp
```diff
@@ -422,16 +503,74 @@ int Reset()
...
+                return Load_ROMLoadError;
+        }
+        else
+        {
+            u8 *romdata = nullptr; u32 romlen;
+            char romfilename[1024] = {0}, sramfilename[1024];
+            strncpy(sramfilename, SRAMPath[ROMSlot_NDS], 1024); // Use existing SRAMPath
+
+            int pos = strlen(sramfilename) - 1;
+            while(pos > 0 && sramfilename[pos] != '/' && sramfilename[pos] != '\\')
```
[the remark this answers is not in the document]
abcdjdj: I attempted using strrchr and got crashes on windows. This happens whenever sramfilename contains both '/' and '\' . For example - "C:/Users/Abcdjdj/Desktop\Pokemon Platinum.sav" In this case, it simply takes the last '/' and proceeds. I can add some code to take whichever of the 2 characters appears last but this seems unnecessary as the older code handles this case (and all other cases) just fine. If you don't mind, I'd like to revert back to the old code and skip using strrchr.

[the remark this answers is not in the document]
RSDuck: sure. Ideally we would replace the entire cumbersome C string path code with one using std::filesystem, but that's a story for another time (which has it's own complications because this somehow would mean we couldn't support older macOS versions).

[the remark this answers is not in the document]
abcdjdj: Thanks, I've reverted it.

abcdjdj:
> do you have anything to add? Otherwise I think we can merge this once my final points are through.

Nope, it's all looking good from my end. It's ready to land :+1:

PR #934, opened 2021-01-10
quit on sigint

bendmorris:
Since the emulator can be launched from the terminal, it'd be nice to be able to close it with CTRL+C.

It's not safe to quit from the signal handler itself, so this uses a QSocketNotifier which is the [qt recommended approach](https://doc.qt.io/qt-5/unix-signals.html).

### DIFF
--- a/src/frontend/qt_sdl/main.cpp
+++ b/src/frontend/qt_sdl/main.cpp
@@ -35,6 +35,12 @@
 #include <QKeyEvent>
 #include <QMimeData>
 #include <QVector>
+#ifndef _WIN32
+#include <QSocketNotifier>
+#include <unistd.h>
+#include <sys/socket.h>
+#include <signal.h>
+#endif
 
 #include <SDL2/SDL.h>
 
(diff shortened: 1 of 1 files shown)


src/frontend/qt_sdl/main.cpp
```diff
@@ -1014,9 +1020,35 @@ void ScreenPanelGL::onScreenLayoutChanged()
     setupScreenLayout();
 }
 
+#ifndef _WIN32
+static int signalFd[2];
+QSocketNotifier *signalSn;
+
+static void signalHandler(int) {
```
RSDuck, line 1027: please put the bracket onto the next line

src/frontend/qt_sdl/main.cpp
```diff
@@ -1014,9 +1020,35 @@ void ScreenPanelGL::onScreenLayoutChanged()
...
+    char a = 1;
+    write(signalFd[0], &a, sizeof(a));
+}
+#endif
 
 MainWindow::MainWindow(QWidget* parent) : QMainWindow(parent)
 {
+#ifndef _WIN32
+    if (::socketpair(AF_UNIX, SOCK_STREAM, 0, signalFd)) {
```
RSDuck, line 1036: same here

RSDuck (changes requested):

PR #937, opened 2021-01-11
Change all mentions of qt5 to qt@5 on macOS

WaluigiWare64:
`brew install qt` will soon install Qt 6.
This is why a new `qt@5` alias has been created.
They recommend changing all `qt5` references to `qt@5`.
Reference: https://github.com/Homebrew/homebrew-core/pull/68241


WaluigiWare64:
The macOS CI currently fails because of this - https://github.com/actions/virtual-environments/issues/2322

PR #942, opened 2021-01-14
Qt6 port

Phantaztique:
Perhaps at some point when Ubuntu adds a qt6 package this will be possible. In the meantime it would need a separate branch.


nadiaholmquist:
https://github.com/Arisotura/melonDS/pull/865
I already did this, though my PR makes using Qt6 optional.

PR #948, opened 2021-01-18
Add support for loading NDS ROMs from a byte array.

SuuperW:
This is helpful for BizHawk integration.

I'm starting to re-do the BizHawk build. It would be helpful in terms of maintaining this build if some (or all) changes are included in the master branch, as this would reduce problems of conflict resolution whenever I pull the latest changes to update the BizHawk build. This PR does not introduce anything that I expect to impact the main build; I have refactored some code, but behavior should be identical.

### DIFF
--- a/src/NDS.cpp
+++ b/src/NDS.cpp
@@ -836,6 +836,20 @@ bool LoadROM(const char* path, const char* sram, bool direct)
         return false;
     }
 }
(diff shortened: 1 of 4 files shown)


RSDuck:
that's already done by #930 which will be merged soon

SuuperW:
Oh, sweet. I guess we don't need this then!

PR #949, opened 2021-01-18
Implement "lag frame" flag and counter

SuuperW:
and expose flag and both frame counters. BizHawk wants these.

This PR modifies the savestate minor version and format, to include new variables. None of the changes should impact any behaviors outside of BizHawk uses.

### DIFF
--- a/src/NDS.cpp
+++ b/src/NDS.cpp
@@ -917,102 +917,108 @@ u32 RunFrame()
 {
     FrameStartTimestamp = SysTimestamp;
 
-    if (!Running) return 263; // dorp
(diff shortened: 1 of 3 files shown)


RSDuck:
I'm not exactly sure how this works, but is it right that frames where the lid is closed aren't taken into account in that calculation?

SuuperW:
> 
> 
> I'm not exactly sure how this works, but is it right that frames where the lid is closed aren't taken into account in that calculation?

Frames where the lid is closed should be counted as lag frames. Thanks for pointing that out!

PR #959, opened 2021-01-22
Support savestate optionally using a memory buffer, instead of a file.

SuuperW:
Again, this is for BizHawk. This commit should not affect any behavior in the main build.

### DIFF
--- a/src/MemoryStream.cpp
+++ b/src/MemoryStream.cpp
@@ -1,0 +1,185 @@
+/*
+    Copyright 2016-2021 Arisotura
+
+    This file is part of melonDS.
+
+    melonDS is free software: you can redistribute it and/or modify it under
+    the terms of the GNU General Public License as published by the Free
+    Software Foundation, either version 3 of the License, or (at your option)
+    any later version.
+
+    melonDS is distributed in the hope that it will be useful, but WITHOUT ANY
+    WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+    FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License along
+    with melonDS. If not, see http://www.gnu.org/licenses/.
+*/
+
+#include "MemoryStream.h"
+#include "types.h"
+
+MemoryStream::MemoryStream()
+{
+	// Create new array of chunks, and an initial chunk.
+	numChunks = 16;
+	chunks = new u8*[numChunks];
+	memset(chunks, 0, numChunks * sizeof(u8*));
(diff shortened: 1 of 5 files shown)


src/Savestate.cpp
```diff
@@ -283,10 +308,32 @@ void Savestate::VarArray(void* data, u32 len)
 
     if (Saving)
     {
-        fwrite(data, len, 1, file);
+        if (mStream)
```
RSDuck, line 311: you can remove all fwrite and fread stuff here, always work on a memory stream and perform one big I/O operation at the start/end. It's cleaner and will probably be faster too because it doesn't constantly call into FS functions.

src/MemoryStream.cpp
```diff
@@ -0,0 +1,185 @@
...
+    with melonDS. If not, see http://www.gnu.org/licenses/.
+*/
+
+#include "MemoryStream.h"
+#include "types.h"
+
+MemoryStream::MemoryStream()
+{
+	// Create new array of chunks, and an initial chunk.
+	numChunks = 16;
```
RSDuck, line 25: this isn't performance critical code, you can just use an std::vector here

src/MemoryStream.cpp
```diff
@@ -0,0 +1,185 @@
...
+		pos += len;
+	}
+}
+
+void MemoryStream::Seek(s32 pos, s32 origin)
+{
+	if (origin == SEEK_SET)
+	{
+		if (pos < 0 || pos > CHUNK_SIZE * numChunks)
+			throw "Invalid seek operation.";
```
RSDuck, line 161: please keep the code base exception free

RSDuck (changes requested):

SuuperW:
I am not familiar with std::vector. Tell me if this looks good.
It's simple enough now that it could rather easily just be included in Savestate.

SuuperW:
This commit (should have been two, sorry) takes care of your other two comments.

src/Savestate.cpp
```diff
@@ -78,81 +89,147 @@ Savestate::Savestate(const char* filename, bool save)
...
-        fseek(file, 0, SEEK_END);
-        u32 len = (u32)ftell(file);
-        fseek(file, 8, SEEK_SET);
-        fwrite(&len, 4, 1, file);
+    // write actual file
+    if (file)
+    {
+        fwrite(buffer.data(), 1, size, file);
+        fclose(file);
+        file == NULL;
```
RSDuck, line 206: I think this is supposed to be a single `=`?

JesseTG:
@SuuperW @RSDuck I was about to work on something like this. Will the requested changes (and merge conflicts) be addressed, or should I just write my own implementation?

PR #970, opened 2021-01-27
SRAM things

SuuperW:
First commit is purely for BizHawk integration.

### DIFF
--- a/src/NDSCart_SRAMManager.cpp
+++ b/src/NDSCart_SRAMManager.cpp
@@ -45,7 +45,6 @@ namespace NDSCart_SRAMManager
     u32 FlushVersion;
 
     void FlushThreadFunc();
-    void FlushSecondaryBufferToFile();
 
     bool Init()
     {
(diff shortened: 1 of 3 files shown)


src/NDSCart_SRAMManager.cpp
```diff
@@ -66,7 +63,8 @@ namespace NDSCart_SRAMManager
             FlushSecondaryBuffer();
         }
 
-        delete SecondaryBuffer;
+        if (SecondaryBuffer) delete SecondaryBuffer;
```
RSDuck, line 66: deleting a null pointer is per spec a nop

src/NDSCart_SRAMManager.cpp
```diff
@@ -121,27 +120,52 @@ namespace NDSCart_SRAMManager
                 continue;
             }
 
-            FlushSecondaryBufferToFile();
+            FlushSecondaryBuffer();
         }
     }
     
-    void FlushSecondaryBufferToFile()
+    void FlushSecondaryBuffer(u8* dst, s32 dstLength)
```
RSDuck, line 131: how/when are these parameters filled in?

SuuperW: When called externally, e.g. from BizHawk. I have a local branch with all of that integration which I will make a PR for at some point.

Issue #208, opened 2017-11-12
[feature request] The ability to change the DS clock


ghost:
In Castlevania Portrait of Ruin, one of the side quests involves talking to a character at a certain time. 

Rather than have to wait hours in real time, I'd like the ability to change the clock.

Karthik99999:
just change the time on your actual computer or boot the ds menu and change the time there

Brian151:
funny, i don't think i've ever booted the DS menu. never had a reason to do so. Still don't. I just find this issue interesting.

mariomadproductions:
There are also programs that can trick an exe into thinking the time is different.

Zettymaster:
to extend on this, maybe we can add an offset to real time, specifying the days, hours and secs for convenience, no user wants to look up unix time.

wundrweapon:
I like Zettymaster's idea but I feel it'd be a lot easier to just have a tickbox for "override system time" and then provide a date/time. Under the hood perhaps that'd be implemented as a Unix time offset, though. Whether that time would update naturally during emulation or be artificially locked is the question this solution poses, though it remains OS-agnostic.

v-fox:
The settings in DS menu are ignored and system time is forced.
But in Linux `libfaketime` can be used to imitate arbitrary time. I completed that annoying PoR quest by starting emulator with `faketime -m '00:00:00' melonDS`, loading, getting completion, saving and restarting normally.

But then melon decided to rename my save-file with a suffix for some reason and neither non-numbered, numbered save-state and suffixed save-state saves-files had picked up the change. I have "separate save-files" enabled. I don't know if it always like that or with this option but melon creates save-file per save-state and only uses it if save-state successfully loaded. I saved to save-state 1's save and save-stated to make sure that it applied but neither non-state/non-numbered nor save-state 1 had quest completed. Until I manually renamed save-state's save-file into non-numbered one and loaded without restoring save-state.

So, either I heavily mixed up my save-files or melon got angry at time change and decided to "fix" that by somehow forcing old save-file to retain. Maybe by ignoring numbered save-state's save-file and silently loading non-numbered one or something.

Basically, expect it to mess with saves when you do that trick.

nadiaholmquist:
That happened because you have separate save files enabled, turn that option off if you don't want it to do this.

v-fox:
> That happened because you have separate save files enabled, turn that option off if you don't want it to do this.

To do what ? What if it just corrupts the only file, like with #1230 ?
[...]

PR #973, opened 2021-01-29
Emulate time

SuuperW:
New config option allows use of emulated clock, rather than taking the time from the host system.


RSDuck:
the option needs to be exposed in the qt gui

WaluigiWare64:
What about #616?

SuuperW:
The option is now available in the gui. It has failed the build checks, though. It looks like it is using QT 5.12, but my new code requires version 5.14. This can be solved by either updating to use version 5.14 (I wouldn't know how to do this.) or I can update the code, as I believe there are deprecated functions that can do the same thing.

@WaluigiWare64 that's an old one. This is the same thing, except compatible with the latest version. I've closed the old one.

src/frontend/qt_sdl/EmuSettingsDialog.cpp
```diff
@@ -24,6 +24,7 @@
 #include "Platform.h"
 #include "Config.h"
 #include "PlatformConfig.h"
+#include "../../RTC.h"
```
RSDuck, line 27: the src folder is part of the include path so you don't need to specify the exact path

RSDuck:
hm according to the log the Ubuntu x64 build is failing because the QT version on the package manager is too old (https://doc.qt.io/qt-5/qdate.html#startOfDay seems to be 5.14+ while the version is 5.12).

WaluigiWare64:
Yeah I think the best option would be to use some alternative function.

RSDuck:
https://code.woboq.org/qt5/qtbase/src/corelib/time/qdatetime.cpp.html#_ZNK5QDate10startOfDayEN2Qt8TimeSpecEi

RSDuck:
can we merge this?

nathanielcwm:
#208

PR #974, opened 2021-01-29
Add option to map cursor to keys

gal20:
Allows the user to move the cursor and press the screen without a mouse, for example it would allow to map a controller's analog stick to move the cursor.
Example screenshot of the new configuration tab:
melon_cursor


poudink:
Seems like a good rudimentary solution to the touch screen problem. Eventually it'd be nice to have a way to set macros (where you could set keys to a specific touch screen coordinate and stuff).

RSDuck:
my Switch port already supports a few options for this, maybe some of the code could probably be adapted from it: https://github.com/RSDuck/melonDS/blob/switch-new/src/frontend/switch/main.cpp#L425-L589 It even supports gyroscope control.

sandwichwater:
something that would be useful to add yo this would be to have a checkbox for something like "when cursor is moving press down cursor and re centre" for games where the touchscreen controls the camera but the main game is the top screen . for example the kingdom hearts games , it would give a more console like camera controls experience using a joystick to control , for mario 64 it would allow you to walk in all 360 degrees rather than the 8 on the dpad . for some of the zelda games that only use the touch screen as controls it would make it way more comfortable to use a joystick . if there was an option for "allow touchscreen input when not on touchscreen" then these games would benefit a lot if you wanted to just use one screen for fullscreen experience

vitor251093:
> for example the kingdom hearts games , it would give a more console like camera controls experience using a joystick to control the camera

I did that in my fork targetting 358/2 Days.

https://github.com/vitor251093/KHDays_FM

I honestly didn't know there were other DS games that used the touchscreen to move the camera as well. If that's the case, I can extract that modification specifically and open a pull request to melonDS, in order to add that feature. Although, I've only tested it in DS mode, not DSi mode. I didn't even implement the DSi support actually. I also didn't add an UI to config those controls, so I may take use of @gal20 's pull request to do that xD

This method is already working for Days, although there is still room for improvement, like supporting the analog sensibility instead of just "moving and not moving". 

It seems that, for some reason, the game makes a fixed small movement in the opposite direction of the move you just made. Example: If I tried to move 1 << 4 to the left on every frame, I would instead move to the right, because the fixed movement was actually bigger than my actual movement. In the other hand, if I tried to move 4 << 4 to the left, it would actually move to the left. 

Considering that, I added a multiplier to touch movement (which is currently 4), which I guess is the equivalent to your cursor speed. However, the horizontal move seems to be way more sensitive than the vertical move, so it might be a good idea to have different multipliers for the X and Y axis.

PR #975, opened 2021-01-30
Extra savestate info

SuuperW:
Three commits for graphical fixes and one lid-related fix.
Including the 3D buffer is necessary to properly render the 1st frame after loading a savestate.


SuuperW:
From the failed build check:

Preparing to unpack .../35-msodbcsql17_17.7.1.1-1_amd64.deb ...
ERROR: The EULA was not accepted. Installation aborted.
dpkg: error processing archive /tmp/apt-dpkg-install-dkNrQl/35-msodbcsql17_17.7.1.1-1_amd64.deb (--unpack):
 new msodbcsql17 package pre-installation script subprocess returned error exit status 1

Is someone messing with stuff here? Failure to accept a EULA seems a strange way to fail.

src/NDS.cpp
```diff
@@ -773,7 +773,8 @@ bool DoSavestate(Savestate* file)
         file->Bool32(&LagFrameFlag);
     }
 
-    // TODO: save KeyInput????
+    if (file->IsAtleastVersion(7, 2))
```
RSDuck, line 776: you don't really need to add that, we don't have any guarantees for savefile compability anyway

src/GPU3D.cpp
```diff
@@ -621,6 +621,27 @@ void DoSavestate(Savestate* file)
...
+
+        // This only works for software renderer. TODO: make it work with OpenGL?
+        u8 stateRenderer = GPU::Renderer;
+        file->Var8(&stateRenderer);
+        if (stateRenderer == 0)
+        {
+            if (GPU::Renderer == 0)
+            {
+                for (int i = 0; i < 192; i++)
+                    file->VarArray(GetLine(i), 256 * sizeof(u32));
```
RSDuck, line 639: this doesn't work when the lines are scrolled

src/GPU3D_Soft.cpp
```diff
@@ -2130,28 +2138,33 @@ void RenderThreadFunc()
...
 
-        RenderThreadRendering = true;
-        if (FrameIdentical)
-        {
-            Platform::Semaphore_Post(Sema_ScanlineCount, 192);
-        }
-        else
+        if (!FrameIdentical)
         {
+            // We cannot rely on each post of the previous frame being matched by a wait, because sleep mode can interrupt a frame.
```
RSDuck, line 2143: that's already accounted for via https://github.com/Arisotura/melonDS/blob/master/src/NDS.cpp#L1166

RSDuck (commented):
instead of saving the frame itself, wouldn't
- stopping the sw rasteriser
- setting loading all the render state from the savestate
- making the texture memory cache coherent
- restaring the frame (like after opening the lid)?
work as well?

RSDuck:
> 
> 
> From the failed build check:
> 
> Preparing to unpack .../35-msodbcsql17_17.7.1.1-1_amd64.deb ...
> ERROR: The EULA was not accepted. Installation aborted.
> dpkg: error processing archive /tmp/apt-dpkg-install-dkNrQl/35-msodbcsql17_17.7.1.1-1_amd64.deb (--unpack):
> new msodbcsql17 package pre-installation script subprocess returned error exit status 1
> 
> Is someone messing with stuff here? Failure to accept a EULA seems a strange way to fail.

the CI can be a bit flaky sometimes

vadosnaprimer:
> Including the 3D buffer is necessary to properly render the 1st frame after loading a savestate.

The core doesn't have to do it. Bizhawk optionally makes a screenshot when you save a state and puts it in when you load it. Unless the plan is to enable rerecording directly in melonds frontend, but even then it's better to make the framebuffer in savestates optional.

src/NDS.cpp
SuuperW: I like to preserve backwards compatibility, the feature exists, and this is a very easy way to do that at least until a major update.

src/GPU3D_Soft.cpp
SuuperW: Oh, that's new. I'll take a look at that, and getting the 3D buffer to re-render instead of saving the buffer itself.

SuuperW:
@vadosnaprimer Yes, I meant the first actually emulated frame. The 3D buffer is applied to the final output on the frame after it is rendered (or, the frame on which rendering began in the case of threaded rendering).

SuuperW:
I force-pushed and it now re-renders the 3D buffer instead of saving/loading it directly.
It only works with the software renderer. I have very little experience with OpenGL and have no interest in making that work.

I also made it blank the frame buffer when loading a state with the lid closed, but I couldn't figure out how to make this work on the frontend when rendering with OpenGL; it'll still show the frame from before loading the savestate.

poudink:
I had noticed that when loading a savestate there was a frame of messed up graphics. So, this is going to fix that?

WaluigiWare64:
> 
> 
> From the failed build check:
> 
> Preparing to unpack .../35-msodbcsql17_17.7.1.1-1_amd64.deb ...
> ERROR: The EULA was not accepted. Installation aborted.
> dpkg: error processing archive /tmp/apt-dpkg-install-dkNrQl/35-msodbcsql17_17.7.1.1-1_amd64.deb (--unpack):
> new msodbcsql17 package pre-installation script subprocess returned error exit status 1
> 
> Is someone messing with stuff here? Failure to accept a EULA seems a strange way to fail.

I've fixed the Ubuntu AArch64 CI issue (https://github.com/Arisotura/melonDS/commit/b5e601bb88858b124b0cc41a4ee7eb6896dee8d3).

SuuperW:
@poudink Yes, this PR fixes all the savestate related graphical bugs that I'm aware of, but some of those bugs are only fixed with the software renderer. I do not currently plan on fixing everything with OpenGL; I'll leave that to someone with a better understanding of how OpenGL is used here. There are two known issues when using OpenGL:
1) It does not render the 3D buffer after loading a savestate. (as of now, it only works when the savestate was both created and loaded with the software renderer) [GPU3D.cpp:659]
2) The display buffer is not updated after loading a savestate in which the lid is closed. [NDS.cpp:773]

RSDuck:
so my plan is the following: I would like to merge #990 first and then add appropriate functions for restarting the frame/giving a black frame with the OpenGL renderer to the new rasteriser interface which would simplify this PR.

SuuperW:
That works for me. I'll probably make a new branch from master after that merge is complete, and cherry-pick or re-work these commits as necessary to work with the new code.

SuuperW:
I force-pushed this branch after rebasing on master. There wasn't really much changed, so the commits are effectively identical to what they were before, just some code has moved from a .cpp file to a .h file. The OpenGL renderer still does not implement the RestartFrame function.

PR #990, opened 2021-02-06
Allow for a more modular renderer backends

Wunkolo:
Contains work based off the initial discussion here: https://github.com/Arisotura/melonDS/issues/984

Includes some fixes and re-encapsulations of global state into encapsulated objects and other fixes to facilitate the addition of additional renderers.

The goal of this interface is to try and provide abstractions that make it easier for alternative renderers such as OpenGL and OpenGL-ES and Vulkan and possibly even alternate software renderer implementations to easily be "plugged in" and "plugged out" of the emulator and possibly even hot-swapped at run-time. This PR can sustain the dialog of how such an interface will ultimately look like and any changes that need to be made before merging.


RSDuck (approved):
this would introduce smart pointers into the code base. I'm ok with that, I just wanted to note that (we need to do some refactory at some point as well, e.g. for switching from NULL to nullptr as well).

PR #991, opened 2021-02-07
Update CRC32 table to compile-time generation

Wunkolo:
This is an update to the CRC32 implementation to use a compile-time
generated CRC32 table rather than a run-time generated table protected
with a boolean.

Using `constexpr`, the table is guaranteed to be created statically using
C++17 features(will re-base once https://github.com/Arisotura/melonDS/pull/990 gets merged in).

Verified that the tables are statically generated across GCC, Clang, and MSVC:
https://godbolt.org/z/75Ta77

Verified CRCs are consistent with the previous implementation

This is based on code from another project of mines to reduce runtime load.
https://github.com/Wunkolo/qCheck

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,6 +1,6 @@
 project(core)
 
-set (CMAKE_CXX_STANDARD 14)
+set (CMAKE_CXX_STANDARD 17)
 
 add_library(core STATIC
 	ARCodeFile.cpp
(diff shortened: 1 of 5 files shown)


Wunkolo:
This PR will also have the CRC32 of NDS roms print to the output. Before, that was only happening for GBA roms.

RSDuck:
thanks for the effort, though we wanted to get rid of the CRC code entirely. We actually used to to calculate the CRC of NDS roms but we removed it because the value was only printed out where nobody would read it. We'll probably replace the CRC part for GBA roms with xxhash which is faster, not much less insecure and already a dependency in other parts of the code base.

PR #1002, opened 2021-02-13
Don't special case the AL condition code, remove (inaccurate) ARM7 multiply carry flag behavior

wheremyfoodat:


RSDuck:
what is the purpose behind this?

Arisotura:
I recall that the AL special case makes it a bit faster (but might as well get rid of it if it's the kind of optimization that is tiny and CPU-dependent)

on the other hand, removing existing behavior because it's inaccurate? do we have hardware tests to confirm what the correct behavior is?

wheremyfoodat:
> I recall that the AL special case makes it a bit faster (but might as well get rid of it if it's the kind of optimization that is tiny and CPU-dependent)
> 
> on the other hand, removing existing behavior because it's inaccurate? do we have hardware tests to confirm what the correct behavior is?

Endrift ran some tests on the GBA ARM7 in December or so iirc and the results were absolutely wack, don't think the carry flag behavior was ever fully discovered.

Arisotura:
don't see any point in removing it either. you never know when that'll unexpectedly break something. my policy is to either keep things as they are (based on available documentation) or reimplement them based on hardware tests. I don't want to change the emulation behavior without any base that isn't an assumption.

wheremyfoodat:
Oh, no problem <3

endrift:
I can confirm, the behavior of the carry flag with mul is very very strange, but I have yet to figure out the proper behavior.

PR #1003, opened 2021-02-13
Fix edge case in the division engine, fix edge case in the CPU

wheremyfoodat:
Both of these fix part of this test ROM: https://github.com/RockPolish/rockwrestler
(Look at the math engine tests, and the last test of the CPU section)

Fixed bugs:
    - Division with INT64_MIN as dividend and -1 as divisor should give a remainder of 0 (look at the test ROM)
    - Certain ALU ops such as MOV were able to switch to Thumb mode. To fix this I made it so the address gets force aligned. Didn't fix the ALU ops with the S bit on yet because they're a bit more complicated

### DIFF
--- a/src/ARMInterpreter_ALU.cpp
+++ b/src/ARMInterpreter_ALU.cpp
@@ -290,7 +290,7 @@ void A_##x##_REG_ROR_REG(ARM* cpu) \
     if (c) cpu->AddCycles_CI(c); else cpu->AddCycles_C(); \
     if (((cpu->CurInstr>>12) & 0xF) == 15) \
     { \
-        cpu->JumpTo(res & ~3); \
+        cpu->JumpTo(res & ~1); \
     } \
     else \
     { \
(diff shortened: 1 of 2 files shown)


src/ARMInterpreter_ALU.cpp
```diff
@@ -321,7 +321,7 @@ A_IMPLEMENT_ALU_OP(AND,_S)
     if (c) cpu->AddCycles_CI(c); else cpu->AddCycles_C(); \
     if (((cpu->CurInstr>>12) & 0xF) == 15) \
     { \
-        cpu->JumpTo(res); \
+        cpu->JumpTo(res & ~3); \
```
RSDuck, line 324: you only need to clear the lowest bit, the rest is handled in JumpTo

RSDuck:
> To fix this I made it so the address gets force aligned. Didn't fix the ALU ops with the S bit on yet because they're a bit more complicated

more complicated as in they have a different behaviour?

wheremyfoodat:
> > To fix this I made it so the address gets force aligned. Didn't fix the ALU ops with the S bit on yet because they're a bit more complicated
> 
> more complicated as in they have a different behaviour?

They can't switch to Thumb mode either, but changing them this way would probably break the flag behavior... and if I change them Heart Gold also refuses to boot.
Rd == 15 and S bit set is usually used to return from exception handlers (it's an edge case of an edge case; it jumps AND loads SPSR into CPSR), so it's probably related to that.

src/ARMInterpreter_ALU.cpp
wheremyfoodat: Alright let me change that real quick

wheremyfoodat: Should be better now

Issue #751, opened 2020-09-09
statically link libQtCore and libslirp in the linux build


eadmaster:
The latest 0.9 linux build is not starting in my Ubuntu Bionic setup due to missing libs:
````
$ ldd melonDS  | grep found

./melonDS: /usr/lib/x86_64-linux-gnu/libQt5Core.so.5: version `Qt_5.12' not found (required by ./melonDS)
        libslirp.so.0 => not found
````

Edit: sorry, wrong ubuntu release name

PoroCYon:
Qt cannot typically be linked statically on Linux (unless you compile it from source yourself, afaics). Static linking will also be annoying with OpenGL video, as those can absolutely not be linked in statically.

There's a Flatpak thingy available in the repo, however, it still uses the old dependencies (libpcap and in-tree libui) and thus will break. @nadiaholmquist something you should look at?

Qt 5.12 and libslirp are both available on Ubuntu Focal, is there a reason you can't update to it? I suppose supporting Qt 5.9 (the one on Bionic) is still possible (with linking libslirp statically), but, I don't know how realistic supporting Xenial/Qt 5.5 is.

eadmaster:
As alternative, a flatpak package would be great for me :-)

gal20:
Try this one #769.

eadmaster:
is there  a public build i can try?

gal20:
Currently no. You'll have to build yourself like this:
`flatpak install flathub org.kde.Platfom//5.15 org.kde.Sdk//5.15`
`flatpak-builder --user --install <build-dir> net.kuribo64.melonDS.yml`
This will create two folders, <build-dir> and .flatpak-builder. You can delete them after installation is done.

FrostKnight:
appimage might be a good way to go, but obviously make sure it depends on the qt version within the appimage and not the one that is outside of the container.

eadmaster:
UPDATE: [build added to flathub](https://flathub.org/apps/details/net.kuribo64.melonDS).

PR #1005, opened 2021-02-16
Add CI workflow using Nix

YoshiRulz:
If you're not familiar with [Nix](https://nixos.org), it's... a strange beast. Nix refers to both a hybrid package manager and build tool, and to the language its build scripts are written in. Nix' main selling point as a build tool is *determinism* (they call it "reproducibility"): only explicitly specified libraries are available in `$PATH` to e.g. `make`, so there are no implicit dependencies on `glibc` or whatever.

Anyway, I think all the headache caused by using Ubuntu for CI can be solved by using Nix. As long as the ["pinned" nixpkgs](https://nixos.org/guides/towards-reproducibility-pinning-nixpkgs.html) hardcoded into the file isn't changed, Qt5 will live forever! (Unfortunately, the CI workflow needs to be Nix *in Ubuntu*, because the great GitHub gremlins have decided that Docker is unnecessary, and they don't offer Alpine or any other Linux distro.)

I'd be happy to answer any questions about Nix in general, though I'm far from being an expert, or about the expression in my commit. As noted, it's basically the same as the one in the unstable version of NixOS.

---

Some more notes:

The melonDS expression I copied from nixpkgs seems to work on `aarch64` [as shown here](https://search.nixos.org/packages?channel=unstable&show=melonDS&from=0&size=50&sort=relevance&query=melonds), and it can probably be cross-compiled from `x86_64`. I'm willing to go down the cross-compilation rabbit hole if there's interest in replacing your existing, misbehaving workflow with a Nix one.

I'm not experienced enough to attempt this, but I think #751 re: static linking is also possible using Nix.


.github/workflows/build-nix-on-ubuntu.yml
```diff
@@ -0,0 +1,27 @@
...
+    - uses: actions/checkout@v1
+    - uses: cachix/install-nix-action@v12
+      with:
+        nix_path: nixpkgs=https://github.com/NixOS/nixpkgs/archive/a127deeb889b111138f5152e54681577706ab741.tar.gz
+        # ^ this is also specified in the expression file, but I think using any other value here may cause superfluous downloads from the binary cache
+    - name: Grab dependencies and build # and move output to CWD
+      run: |
+        nix-build ./ci-build-drv.nix
+        cp -r "$(realpath "$PWD/result")/bin" ${{runner.workspace}}
+    - uses: actions/upload-artifact@v1
```
Margen67, line 24:
```suggestion
    - uses: actions/upload-artifact@v2
```

.github/workflows/build-nix-on-ubuntu.yml
```diff
@@ -0,0 +1,27 @@
...
+    - master
+  pull_request:
+    branches:
+    - master
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
```
Margen67, line 15:
```suggestion
    - uses: actions/checkout@v2
```
If you need all commits:
```suggestion
    - uses: actions/checkout@v2
      with:
        fetch-depth: 0
```

Issue #1016, opened 2021-02-25
Build failure on NetBSD


enkl7:
Build fails on both ARM64 and x86 on NetBSD 9.1

ARM64:

```
[ 56%] Building CXX object src/CMakeFiles/core.dir/ARMJIT_Memory.cpp.o
/home/x/melonDS/src/ARMJIT_Memory.cpp: In function ‘void SigsegvHandler(int, siginfo_t*, void*)’:
/home/x/melonDS/src/ARMJIT_Memory.cpp:173:60: error: ‘struct mcontext_t’ has no member named ‘fault_address’
         desc.EmulatedFaultAddr = (u8*)context->uc_mcontext.fault_address - curArea;
                                                            ^~~~~~~~~~~~~
/home/x/melonDS/src/ARMJIT_Memory.cpp:174:50: error: ‘struct mcontext_t’ has no member named ‘pc’
         desc.FaultPC = (u8*)context->uc_mcontext.pc;
                                                  ^~
/home/x/melonDS/src/ARMJIT_Memory.cpp:194:34: error: ‘struct mcontext_t’ has no member named ‘pc’
             context->uc_mcontext.pc = (u64)desc.FaultPC;
                                  ^~
*** Error code 1
```

I'm able to bypass this error by adding an ifdef:
```
#elif defined(__NetBSD__)
    _UC_MACHINE_SET_PC(context, (u64)desc.FaultPC);
```

However, now I'm stuck at the linking process which is where it stops on x86 as well with the same error:
```

[ 67%] Building ASM object src/CMakeFiles/core.dir/ARMJIT_A64/ARMJIT_Linkage.S.o
[ 68%] Linking CXX static library libcore.a
Error running link command: No such file or directory
*** Error code 2
```

WaluigiWare64:
The error is due to LTO. We set the `CMAKE_AR` and `CMAKE_RANLIB` to `gcc-ar` and `gcc-ranlib`.
...however, they don't seem to exist on NetBSD.
In the [NetBSD Docs](https://man.netbsd.org/gcc.1) they mention:
```
           If object files containing GIMPLE bytecode are stored in a library
           archive, say libfoo.a, it is possible to extract and use them in an
           LTO link if you are using a linker with plugin support.  To create
           static libraries suitable for LTO, use gcc-ar and gcc-ranlib
           instead of ar and ranlib; to show the symbols of object files with
           GIMPLE bytecode, use gcc-nm.  Those commands require that ar,
[...]

PR #1018, opened 2021-02-26
Only use special ar and ranlib command when ENABLE_LTO is ON.

niacat:
This allows melonDS to be built with the standard system toolchain on NetBSD with -DENABLE_LTO=OFF, see discussion in #1016. I tested NetBSD 9.99.79 (-current) which comes with gcc 9.3.0 by default. NetBSD 9 with gcc 7 probably also works.


WaluigiWare64:
Looks good to me.
@nadiaholmquist what do you think?

nadiaholmquist:
As long as it doesn't break any platforms it should be all good.

PR #1020, opened 2021-02-28
Add hotkeys for cycling screen layout and screen sizing.

AdmiralCurtiss:
Just seems handy to have, especially on a controller where you'd otherwise need to grab a mouse to change it.


Koas9000:
@RobLoach OK this pr is done can part master build

PR #1040, opened 2021-03-24
ARMJit Compiler::Reset(): Cleaner way to fill memory with BRK.

AdmiralCurtiss:
See https://melonds.kuribo64.net/comments.php?id=137&cid=2971#comment2971

I have no way to test this because of a lack of debuggable high-power ARM devices, but I think this is right.


StLouisCPhT:
Let me test on my M1 with the full change.

RSDuck:
we break aliasing all the time, so while this is more correct, it's really a drop in the ocean

AdmiralCurtiss:
Ah there was a hidden const_cast in there before, that's even more sketchy. Well either way, fair enough, I just saw the blog post and my brain went 'wait this is not right'.

StLouisCPhT:
> Ah there was a hidden const_cast in there before, that's even more sketchy. Well either way, fair enough, I just saw the blog post and my brain went 'wait this is not right'.

It's definitely not. 
It's not so much a crash as a hang when the JIT and FastMem are both enabled.

XiaomiRedmi9:
How about this now?

RSDuck:
I mean we can merge this, though we still rely massively break aliasing on so many other places in so much more complex ways, that if the compiler ever bits back, it'll do somewhere else first.

PR #1042, opened 2021-03-27
Use Azure Pipelines so we can target macOS 10.14

WaluigiWare64:


azure-pipelines.yml
```diff
@@ -7,13 +7,24 @@ trigger:
...
-    echo See https://aka.ms/yaml
-  displayName: 'Run a multi-line script'
+- script: mkdir $(Pipeline.Workspace)/build
+  displayName: 'Create build environment'
+
+- script: cmake $(Build.SourcesDirectory) -DCMAKE_BUILD_TYPE=$BUILD_TYPE -DQt5_DIR=$(brew --prefix qt5)/lib/cmake/Qt5
+  displayName: 'Configure'
+  workingDirectory: mkdir $(Pipeline.Workspace)/build
+  
+- script: make -j$(sysctl -n hw.ncpu)
```
Margen67, line 23: `ncpu` is deprecated: https://github.com/ziglang/zig/issues/1252 https://mail.haskell.org/pipermail/ghc-tickets/2013-December/008198.html
```suggestion
- script: make -j$(sysctl -n hw.availcpu)
```

WaluigiWare64: thanks, will fix.

PR #1054, opened 2021-04-10
Add .desktop file for linux users

Nine-H:
Hey there, this is a micro commit, it just adds a linux .desktop file to the repository, these metadata files are really powerful, having one means you get a listing in the launcher, and can set up file association properly. I also added some instructions for manually adding it and the icons to a linux system. It's broken out into two commits, one with the metadata file and one with the readme edits in case you want to rewrite them, or move them to a wiki page, or to a bug to make an installer.

I considered writing a shell script to install everything, but it's way out of scope for a driveby pull request.

Have a good one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,21 @@ As for the rest, the interface should be pretty straightforward. If you have a q
   cmake ..
   make -j$(nproc --all)
   ```
+4. Add melonDS to system
+  ```bash
+  # add binary to bin folder
+  sudo cp melonds /usr/bin
+  # add launcher to system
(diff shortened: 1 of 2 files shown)


hcorion:
If you do a `make install` instead of trying to manually install melonDS it installs the icon and a desktop file, making this change redundant.
There's already a .desktop file here https://github.com/Arisotura/melonDS/blob/d5dbef2c22f755ad82ec59e04948111797bc270a/net.kuribo64.melonDS.desktop

WaluigiWare64:
See above comment.

PR #1067, opened 2021-04-20
Improve macOS bundling

nadiaholmquist:
This PR adds a script that wraps `macdeployqt` and ensures all the libraries it pulls in have their load paths fixed. Specifying `-DMACOS_BUNDLE_LIBS=ON` should now bundle all the libraries needed for the .app file to be fully portable across machines.

Also moves the dist scripts to a `tools/` directory as that seemed a bit cleaner to me.


.github/workflows/build-macos.yml
```diff
@@ -26,15 +26,12 @@ jobs:
     - name: Configure
       working-directory: ${{runner.workspace}}/build
       run: |
-          export PKG_CONFIG_PATH="$(brew --prefix libarchive)/lib/pkgconfig"
-          cmake $GITHUB_WORKSPACE -DCMAKE_BUILD_TYPE=$BUILD_TYPE -DQt5_DIR=$(brew --prefix qt5)/lib/cmake/Qt5
+          cmake $GITHUB_WORKSPACE -DCMAKE_BUILD_TYPE=$BUILD_TYPE -DCMAKE_PREFIX_PATH="$(brew --prefix qt@5);$(brew --prefix libarchive)" -DMACOS_BUNDLE_LIBS=ON -DMACOS_BUILD_DMG=ON
```
WaluigiWare64, line 29: Could you disable bundling libraries in CI, otherwise it will only work on 10.15 and above.

tools/mac-libs.sh
```diff
@@ -0,0 +1,67 @@
...
+	done
+}
+
+find "$app/Contents/Frameworks" -maxdepth 1 -name '*.dylib' | while read lib; do
+	fixup_libs "$lib"
+done
+
+fixup_libs "$app/Contents/MacOS/melonDS"
+
+if [[ $build_dmg == 1 ]]; then
```
WaluigiWare64, line 61: Macdeployqt has an option to build a dmg, see the documentation.

WaluigiWare64 (dismissed):
Other than that, looks good!

.github/workflows/build-macos.yml
nadiaholmquist: Problem is, if i do that the resulting binary will not actually run unless you have homebrew and all required libraries already installed on the system you want to run the build on.

tools/mac-libs.sh
nadiaholmquist: Yes but we need to do some more stuff to the app bundle after calling macdeployqt - technically I could run it a second time, but that would be rather pointless and slow and also print a bunch of errors, so it's just easier that we build the DMG.

PR #1072, opened 2021-04-24
Fix MSYS2 build command line

rzumer:


RSDuck:
am I missing something? Because msys_dist.sh is still in the tools folder.

rzumer:
I was on the `cart_refactor` branch which was behind master and did not have it there. Didn't notice it moved there on master.

PR #1095, opened 2021-05-04
Add ROM Header struct and ROM info dialog

WaluigiWare64:
**To-do list:**
- [x] Separate into one icon buffer and one index buffer.
- [x] Don't use seperate loops

image

### DIFF
--- a/src/NDSCart.cpp
+++ b/src/NDSCart.cpp
@@ -58,8 +58,8 @@ u32 CartID;
 bool CartIsHomebrew;
 bool CartIsDSi;
 
-struct NDSHeader Header;
-struct NDSBanner Banner;
+NDSHeader Header;
+NDSBanner Banner;
 
 CartCommon* Cart;
 
--- a/src/NDSCart.cpp
+++ b/src/NDSCart.cpp
@@ -58,6 +58,9 @@ u32 CartID;
 bool CartIsHomebrew;
 bool CartIsDSi;
 
+struct NDSHeader Header;
+struct NDSBanner Banner;
+
 CartCommon* Cart;
 
 u32 Key1_KeyBuf[0x412];
@@ -1495,8 +1498,11 @@ void DecryptSecureArea(u8* out)
     // * .srl ROMs (VC dumps) have encrypted secure areas but have precomputed
     //   decryption data at 0x1000 (and at the beginning of the DSi region if any)
 
-    u32 gamecode = *(u32*)&CartROM[0x0C];
-    u32 arm9base = *(u32*)&CartROM[0x20];
+    u32 gamecode = (u32)Header.GameCode[3] << 24 |
+                   (u32)Header.GameCode[2] << 16 |
+                   (u32)Header.GameCode[1] << 8  |
+                   (u32)Header.GameCode[0];
+    u32 arm9base = Header.ARM9ROMOffset;
 
     memcpy(out, &CartROM[arm9base], 0x800);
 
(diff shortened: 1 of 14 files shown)


src/frontend/Util_ROM.cpp
```diff
@@ -461,6 +461,37 @@ int LoadROM(const char* file, int slot)
     }
 }
 
+u32* RomIcon(u8 data[512], u16 palette[16])
```
RSDuck, line 464: if the data size is known to the caller and never too large, it's preferable for the caller to pass in a pointer where the data can be stored so the stack can be used for this.

src/NDSCart.cpp
```diff
@@ -58,6 +58,9 @@ u32 CartID;
 bool CartIsHomebrew;
 bool CartIsDSi;
 
+struct NDSHeader Header;
```
RSDuck, line 61: writing struct before a struct type shouldn't be necessary in C++, unless it's different with packed one?

src/frontend/Util_ROM.cpp
```diff
@@ -461,6 +461,37 @@ int LoadROM(const char* file, int slot)
...
+{
+    // Get the 4-bit palette indexes
+    u8 indexes[1024];
+    for (int i = 0; i < 512; i++)
```
RSDuck, line 468: you don't have to use three loops for this, the entire operation can be done at once.

RSDuck (changes requested):

src/frontend/Util_ROM.cpp
WaluigiWare64: Yeah i didn't write this, this is Hydr8gon's code that you showed me.

Arisotura: Hydr8gon is a derp

src/frontend/Util_ROM.cpp
Arisotura: RomIcon? the capitalization there is inconsistent with the rest of the code

src/NDSCart.cpp
Arisotura: shrug idk how all this struct shit works and why everybody does it differently

src/frontend/Util_ROM.cpp
WaluigiWare64: > 
> 
> RomIcon? the capitalization there is inconsistent with the rest of the code

Should I change it to ROMIcon?

src/frontend/Util_ROM.cpp
```diff
@@ -461,6 +461,76 @@ int LoadROM(const char* file, int slot)
...
+        ROMIcon(data[SEQ_BMP(sequence[i])], palette[SEQ_PAL(sequence[i])], &animatedTexRef[32 * 32 * i]);
+
+        if (SEQ_FLIPH(sequence[i]))
+        {
+            for (int x = 0; x < 32; x++)
+            {
+                for (int y = 0; y < 32/2; y++) 
+                {
+                    int temp = frame[x * 32 + y];
+                    frame[x * 32 + y] = frame[x * 32 + (32 - 1 - y)];
```
RSDuck, line 511: use std::swap for that

src/frontend/Util_ROM.cpp
```diff
@@ -461,6 +461,76 @@ int LoadROM(const char* file, int slot)
...
+#define SEQ_DUR(i) ((i & 0b0000000011111111) >> 0)
+
+void AnimatedROMIcon(u8 (&data)[8][512], u16 (&palette)[8][16], u16 (&sequence)[64], u32 (&animatedTexRef)[32 * 32 * 64], std::vector<int> &animatedSequenceRef)
+{
+    for (int i = 0; i < 64; i++)
+    {
+        if (!sequence[i])
+            break;
+        u32* frame = &animatedTexRef[32 * 32 * i];
+        ROMIcon(data[SEQ_BMP(sequence[i])], palette[SEQ_PAL(sequence[i])], &animatedTexRef[32 * 32 * i]);
```
RSDuck, line 502: replace &animatedTexRef[32 * 32 * i] by frame

PR #1117, opened 2021-05-29
Added 'About' menu item w/ scmrev information

jorundev:
Allows the user to know on which revision they're on, alongside some useful links.
[DO NOT MERGE]

- Added an 'About' section (Located in the 'Help' menubar item or in the 'melonDS' menubar item on macOS)
- Added a way to get the current git revision and the branch from code (those informations are also present on the about section)

This is not a super-big commit but I thought it was useful. I'm not sure the git-scmrev.sh script works on Windows tho, I don't have access to a machine with windows right now so I need feedback. I'm also not sure about the informations I put on the about page. Maybe this PR is shit and it was a bad idea from the beginning but it was fun playing with the source code anyway.


RSDuck:
it doesn't build on Windows, see the CI.

jorundev:
> it doesn't build on Windows, see the CI.

Yep I know I mentionned it in the issue. I don't have a windows machine with build tools available right now so unless someone wants to help me with that I will have to wait like a week before being able to do anything

PR #1119, opened 2021-06-05
Add "frame step" hotkey

Yukitty:
Automatically pauses emulation if necessary, and then emulates a single frame with every activation.


Yukitty:
(oops 😓)

RSDuck:
thanks for the PR.

PR #1123, opened 2021-06-06
Add PoroCYon's DSP code.

purringChaos:
This is a rebase of https://gitlab.ulyssis.org/pcy/melonds/-/tree/dsp/

Adds incomplete DSP code but not enough to cause more problems.

Also includes slimmed down Teakra repo.


src/DSi_DSP.cpp
```diff
@@ -0,0 +1,591 @@
...
+u16 DSP_PCLEAR;
+u16 DSP_CMD[3];
+u16 DSP_REP[3];
+
+u64 DSPTimestamp;
+
+FIFO<u16, 16> PDATAReadFifo/*, *PDATAWriteFifo*/;
+int PDataDMALen = 0;
+
+static u16 GetPSTS()
```
RSDuck, line 50: why is the static here?

purringChaos: remind me what static does again? i dont think it needs to be static that was ust a side eect o some tracking code. i can revert it to non-static.

purringChaos: i think that might o been a change in the codebase a year ago, should i remove static rom all methods?

RSDuck: static shouldn't be necessary because we use C++ namespaces

src/DSi_DSP.cpp
```diff
@@ -0,0 +1,591 @@
...
+    while (backlog & (1uLL<<32)) // god I hope this never happens
+    {
+        Run((u32)(backlog & ((1uLL<<32)-1)));
+        backlog = curtime - DSPTimestamp;
+    }
+    Run((u32)backlog);
+
+    return true;
+}
+static void DSPCatchUpU32(u32 _) { DSPCatchUp(); }
```
RSDuck: you can just make add a default parameter to DSPCatchUp

src/DSi_DSP.cpp
```diff
@@ -0,0 +1,592 @@
...
+    {
+        printf("WTF?? (DSP MBK recfg, nonsense bank '%c')\n", bank);
+        return;
+    }
+
+    bool olddsp = (oldcfg & 3) >= 2, // was mapped to the DSP
+         newdsp = (newcfg & 3) >= 2; // will be mapped to the DSP
+
+    // nothing changes
+    if (olddsp == newdsp)
```
PoroCYon, line 179: (note to self) This code forgets about the case of mapping the same NWRAM bank differently to the DSP (if that's a possibility, should be investigated), and thus needs to be reworked a bit.

Issue #936, opened 2021-01-11
[Feature Request] Make MelonDS remember if it was closed on windowed mode or fullscreen mode


LOuroboros:
This came to mind after seeing #933.
If you close MelonDS on Fullscreen Mode, the next time you open it, it opens in Windowed Mode.
It would be nice if MelonDS could boot up on Fullscreen Mode is that was its state when last closed.

bell07:
Just for reference, my workaround to start MelonDS in fullscreen mode on linux, using wrapper script and xdotool

```
$ cat run-melonds.sh 
#!/bin/bash

# run MelonDS in background
melonDS "$@" &

# wait 2 seconds and send alt+enter key
sleep 2
xdotool search --sync --class melonDS | while read WINDOW; do
	xdotool key --window $WINDOW alt+Return
done

```

PR #1128, opened 2021-06-11
Partial support for remembering last launched full screen mode.

dalchen:
This patch only works for toggling full screen via the hotkey. For
some reason, QT's isFullScreen is false if you go from not full screen
to full screen via the green macOS window button.

Tested the following on macOS (hotkey means configured hotkey for full screen):

-- Working cases:
* launch, quit, launch. Result: not full screen.
* launch, hotkey, quit, launch. Result: full screen.
* launch, hotkey, hotkey, quit, launch. Result: not full screen.
* launch, resize, hotkey, quit, launch (Result: full screen), hotkey.
  Result: not full screen with previously resized window size.

-- Not working cases (mouse means mouse click on green macOS window button):
* launch, mouse, quit, launch. Incorrect result: not full screen.
* launch, hotkey, mouse, quit, launch. Incorrect result: full screen.

### DIFF
--- a/src/frontend/qt_sdl/PlatformConfig.cpp
+++ b/src/frontend/qt_sdl/PlatformConfig.cpp
@@ -35,6 +35,7 @@ int JoystickID;
 int WindowWidth;
 int WindowHeight;
 int WindowMaximized;
+int WindowFullScreen;
 
 int ScreenRotation;
 int ScreenGap;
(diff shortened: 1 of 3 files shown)


dalchen:
First time contributor here! This PR is for #936 but doesn't fix it all the way as described in the commit message. I have also only tested this on macOS. Let me know if this is helpful or if any of you have any suggestions, thanks!

PR #1130, opened 2021-06-12
Create ROMManager class to manage ROM/Save Data/Cheat File resources in Frontend.

dalchen:
This patch aims to create a class that manages the resources used for
files in the Frontend namespace. Original functionality of all the
functions moved into this class is kept. Resources are managed by the
Meyers Singleton ROMManager, and members are managed via RAII.


dalchen:
Reordered method definitions around for a cleaner diff.

Please let me know if this patch is reasonable! I will totally understand if you don't want to merge this since this patch has quite a few (unwarranted since things were working properly) changes.

RSDuck:
sorry, I don't really see the point of this. It doesn't seem to be useful to apply OOP in this case, especially since the singleton then destroys all the potential benefits of it again.

PR #1132, opened 2021-06-16
Make building with libslirp optional

nadiaholmquist:
This commit makes building using libslirp optional and disables indirect mode in the GUI if libslirp was not found. Currently required to build on some platforms that do not have libslirp, such as FreeBSD.

### DIFF
--- a/src/frontend/qt_sdl/CMakeLists.txt
+++ b/src/frontend/qt_sdl/CMakeLists.txt
@@ -12,7 +12,6 @@ SET(SOURCES_QT_SDL
     InterfaceSettingsDialog.cpp
     Input.cpp
     LAN_PCap.cpp
-    LAN_Socket.cpp
     OSD.cpp
     OSD_shaders.h
     font.h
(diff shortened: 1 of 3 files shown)


src/frontend/qt_sdl/CMakeLists.txt
```diff
@@ -70,21 +69,41 @@ find_package(Threads REQUIRED)
...
+
 if (WIN32 AND (CMAKE_BUILD_TYPE STREQUAL Release))
     add_executable(melonDS WIN32 ${SOURCES_QT_SDL})
 else()
     add_executable(melonDS ${SOURCES_QT_SDL})
 endif()
 
+if (ENABLE_LIBSLIRP)
+	target_compile_definitions(melonDS PRIVATE -DHAVE_LIBSLIRP)
+    target_sources(melonDS PRIVATE LAN_Socket.cpp)
```
WaluigiWare64, line 90: I think you mixed up tabs and spaces here.

nadiaholmquist: fark

We really need an editorconfig file.

src/frontend/qt_sdl/CMakeLists.txt
```diff
@@ -70,21 +69,41 @@ find_package(Threads REQUIRED)
...
-pkg_check_modules(SLIRP REQUIRED slirp)
+pkg_check_modules(SLIRP slirp)
 pkg_check_modules(LIBARCHIVE REQUIRED libarchive)
 add_compile_definitions(ARCHIVE_SUPPORT_ENABLED)
 
+if (SLIRP_FOUND)
+    option(ENABLE_LIBSLIRP "Enable use of libslirp for network sockets" ON)
+else()
+    option(ENABLE_LIBSLIRP "Enable use of libslirp for network sockets" OFF)
+endif()
```
WaluigiWare64, line 80:
```suggestion
if (SLIRP_FOUND)
    option(ENABLE_LIBSLIRP "Enable use of libslirp for network sockets" ON)
    message(STATUS "libslirp found, and libslirp support is enabled")
else()
    option(ENABLE_LIBSLIRP "Enable use of libslirp for network sockets" OFF)
    message(WARNING libslirp not found, building without libslirp support!")
endif()
```
Add a generate time message so it's clear if `libslirp` is found or not.

PR #1149, opened 2021-07-07
NWRAM & Direct Boot

DesperateProgrammer:
The patches enables switching between NDS and DSi Bios in DSI mode depending in the SCFG_BIOS register.

If the Header->UnitCode of the NDS shows it is NDS based (NDS or DSi Extended) the correct ROM is accessed.
Also the libNDS crt0 now detects correctly that is running in NDS mode on the DSi if this SCFG_BIOS selection is for the NDS, fixing such files jumping into nirwana.

I used a NWRAM Testcase as an example:

DSi Emulation - NDS Mode (SCFG_BIOS = 0303h):
grafik

DS Emulation:
grafik

### DIFF
--- a/src/DSi.cpp
+++ b/src/DSi.cpp
@@ -2055,30 +2055,50 @@ void ARM9IOWrite8(u32 addr, u8 val)
         return;
 
     case 0x04004006:
+        if (!(SCFG_EXT[1] & (1 << 31))) /* no access to SCFG Registers if disabled*/
+            return;
         SCFG_RST = (SCFG_RST & 0xFF00) | val;
         DSi_DSP::SetRstLine(val & 1);
         return;
 
-    case 0x04004040: MapNWRAM_A(0, val); return;
-    case 0x04004041: MapNWRAM_A(1, val); return;
-    case 0x04004042: MapNWRAM_A(2, val); return;
-    case 0x04004043: MapNWRAM_A(3, val); return;
-    case 0x04004044: MapNWRAM_B(0, val); return;
-    case 0x04004045: MapNWRAM_B(1, val); return;
-    case 0x04004046: MapNWRAM_B(2, val); return;
-    case 0x04004047: MapNWRAM_B(3, val); return;
-    case 0x04004048: MapNWRAM_B(4, val); return;
-    case 0x04004049: MapNWRAM_B(5, val); return;
-    case 0x0400404A: MapNWRAM_B(6, val); return;
-    case 0x0400404B: MapNWRAM_B(7, val); return;
-    case 0x0400404C: MapNWRAM_C(0, val); return;
-    case 0x0400404D: MapNWRAM_C(1, val); return;
-    case 0x0400404E: MapNWRAM_C(2, val); return;
-    case 0x0400404F: MapNWRAM_C(3, val); return;
-    case 0x04004050: MapNWRAM_C(4, val); return;
-    case 0x04004051: MapNWRAM_C(5, val); return;
-    case 0x04004052: MapNWRAM_C(6, val); return;
-    case 0x04004053: MapNWRAM_C(7, val); return;
+    case 0x04004040:
+    case 0x04004041:
+    case 0x04004042:
+    case 0x04004043:
+        if (!(SCFG_EXT[1] & (1 << 31))) /* no access to SCFG Registers if disabled*/
+            return;
+        if (!(SCFG_EXT[1] & (1 << 25))) /* no access to NVRAM Registers if disabled*/
+            return;
+        MapNWRAM_A(addr & 3, val);
+        return;
+    case 0x04004044:
+    case 0x04004045:
+    case 0x04004046:
+    case 0x04004047:
+    case 0x04004048:
+    case 0x04004049:
+    case 0x0400404A:
+    case 0x0400404B:
+        if (!(SCFG_EXT[1] & (1 << 31))) /* no access to SCFG Registers if disabled*/
+            return;
+        if (!(SCFG_EXT[1] & (1 << 25))) /* no access to NVRAM Registers if disabled*/
+            return;
+        MapNWRAM_B((addr - 0x04) & 7, val);
+        return;
+    case 0x0400404C:
+    case 0x0400404D:
+    case 0x0400404E:
+    case 0x0400404F:
+    case 0x04004050:
+    case 0x04004051:
+    case 0x04004052:
+    case 0x04004053:
+        if (!(SCFG_EXT[1] & (1 << 31))) /* no access to SCFG Registers if disabled*/
+            return;
+        if (!(SCFG_EXT[1] & (1 << 25))) /* no access to NVRAM Registers if disabled*/
+            return;
+        MapNWRAM_C((addr-0x0C) & 7, val);
+        return;
     }
 
     if ((addr & 0xFFFFFF00) == 0x04004200)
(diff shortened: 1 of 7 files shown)


DesperateProgrammer:
Sorry, meant to do these in a seperate PR.

But since they are already here:
The NWRAM is now working as on HW in regard to the priorities and direct boot the stage2 bootloader is no longer loaded.
The SCFG and NWRAM registers can not be written after they are disabled (i.e. in NDS Mode)

The result of this is loading NDS & DSi enhanced titles just fine.
More work is to be done in blending out HW / using NDS Hardware when disabled (i.e. the TSC)

Comparison pre to post patch laoding SMB-DS in DSi Emulation, direct boot:
grafik

src/DSi.cpp
```diff
@@ -864,15 +943,19 @@ u16 ARM9Read16(u32 addr)
 
     return NDS::ARM9Read16(addr);
 }
+#pragma GCC push_options
+#pragma GCC optimize ("O0")
```
RSDuck, line 947: what is this for?

Arisotura:
I have criticism about the SCFG part:

1. cart-dependent SCFG setup should be done in a separate function, DSi::SetupDirectBoot or something of that sort. the base idea with the way melonDS is structured is, Reset() prepares the system for a regular boot (ie from BIOS/firmware), SetupDirectBoot() does the extra setup needed to start from the cart directly on top of that.

2. SCFG BIOS selection/lockout (in the ARMxReadxx functions) was already implemented, your code there is redundant. (unless I missed something in that domain?)

I don't have much to say about the NWRAM code.

src/DSi.cpp
DesperateProgrammer: Good catch.  That is a debugging remainder. I will remove it. Thought i caught them all.

DesperateProgrammer:
> I have criticism about the SCFG part:
> 
>     1. cart-dependent SCFG setup should be done in a separate function, DSi::SetupDirectBoot or something of that sort. the base idea with the way melonDS is structured is, Reset() prepares the system for a regular boot (ie from BIOS/firmware), SetupDirectBoot() does the extra setup needed to start from the cart directly on top of that.

Will change this.
 
>     2. SCFG BIOS selection/lockout (in the ARMxReadxx functions) was already implemented, your code there is redundant. (unless I missed something in that domain?)

Hmm. I can only find references to these two bits by the JIT.
Could you point me where this BIOS selection is done?

> I don't have much to say about the NWRAM code.

Arisotura:
> > ```
> > 2. SCFG BIOS selection/lockout (in the ARMxReadxx functions) was already implemented, your code there is redundant. (unless I missed something in that domain?)
> > ```
> 
> Hmm. I can only find references to these two bits by the JIT.
> Could you point me where this BIOS selection is done?
https://github.com/Arisotura/melonDS/blob/master/src/DSi.cpp#L786 for example.

if bit1 is set, it will let the function fall through to the equivalent NDS function, which will read from the NDS BIOS.

DesperateProgrammer:
> https://github.com/Arisotura/melonDS/blob/master/src/DSi.cpp#L786 for example.
> 
> if bit1 is set, it will let the function fall through to the equivalent NDS function, which will read from the NDS BIOS.

Oh, I must have been blind. You are right. My branch is never taken.
Will fix.

DesperateProgrammer:
Done. 
Btw. I have attached the NWRAM Testcase I used. So you can see yourself.
[nwramTest.zip](https://github.com/Arisotura/melonDS/files/6785597/nwramTest.zip)

src/DSi.cpp
```diff
@@ -312,6 +336,21 @@ bool LoadBIOS()
 
 bool LoadNAND()
 {
+    // Skip loading the stage2 if we started directly a ROM
+//    if (NDS::IsDirectBoot)
```
RSDuck, line 340: doesn't this also belong into SetupDirectBoot?

src/DSi.cpp
```diff
@@ -583,17 +622,29 @@ void MapNWRAM_A(u32 num, u8 val)
...
-        if (NWRAMMap_A[oldval & 0x01][(oldval >> 2) & 0x3] == ptr)
-            NWRAMMap_A[oldval & 0x01][(oldval >> 2) & 0x3] = NULL;
+        NWRAMMap_A[0][part] = NULL;
+        NWRAMMap_A[1][part] = NULL;
     }
-
-    if (val & 0x80)
+    // can only map if NWRAM is available
+    if (!NWRAM_A)
+        NWRAM_A = new u8[NWRAMSize];
```
RSDuck, line 637: NWRAM_A should always be allocated here. Besides you can't just allocate heap memory here at all because when fast memory is enabled special pointers related to it are used for this.

src/DSi.cpp
```diff
@@ -884,19 +961,32 @@ u32 ARM9Read32(u32 addr)
         break;
 
     case 0x03000000:
+    case 0x03800000:
         if (addr >= NWRAMStart[0][0] && addr < NWRAMEnd[0][0])
         {
             u8* ptr = NWRAMMap_A[0][(addr >> 16) & NWRAMMask[0][0]];
+#if DSI_NWRAM_DEBUGOUT
+            printf("Set A9: Read %08x from NWRAM at %08x (source=%08x)\n", ptr ? *(u32*)&ptr[addr & 0xFFFF] : 0, addr, ptr);
+            fflush(stdout);
```
RSDuck, line 970: this shouldn't be necessary

src/DSi.cpp
```diff
@@ -918,42 +1008,92 @@ void ARM9Write8(u32 addr, u8 val)
...
+                    continue;
+                u8* ptr = &NWRAM_A[page * 0x8000];
+                *(u8*)&ptr[addr & 0x7FFF] = val;
+#if DSI_NWRAM_DEBUGOUT
+                printf("Set A9: Wrote %08x to NWRAM at %08x (source=%08x)\n", val, addr, ptr);
+                fflush(stdout);
+#endif
 #ifdef JIT_ENABLED
-                ARMJIT::CheckAndInvalidate<0, ARMJIT_Memory::memregion_NewSharedWRAM_A>(addr);
+                ARMJIT::CheckAndInvalidate<1, ARMJIT_Memory::memregion_NewSharedWRAM_A>(addr);
```
RSDuck, line 1035: why did you change that parameter? 0=ARM9

src/DSi.cpp
```diff
@@ -185,6 +186,29 @@ void Reset()
     ARM7Write16(eaddr+0x42, 0x0001);
 }
 
+void SetupDirectBoot()
+{
+    // If loading a NDS directly, this value should be set
+    // depending on the console type in the header at offset 12h
+    if (NDS::IsDirectBoot)
```
RSDuck, line 193: the if and the variable is unnecessary

RSDuck (changes requested):

src/DSi.cpp
DesperateProgrammer: Thanks. This is a c&p error I did not spot.

src/DSi.cpp
DesperateProgrammer: Given that the access massively changed, This allows observing the accesses to verify it along the testcase.
Since this is guarded at compile time, It is not included in the build unless forced to.

src/DSi.cpp
DesperateProgrammer: I would agree if we move the LoadNAND out of the Reset() and SofReset() as it messes up the NWRAM content and mapping, even tho this is not a hardware function of the reset.
Rather it is just another nds image just taken from a different location.

Maybe we should introduce a "SetupBottloader" parallel to the SetupDirectBoot?

src/DSi.cpp
DesperateProgrammer: You are correct. I had issues with NWRAM_A/B/C beein empty in the process, but that later in debugging resolved as an out of bounds write to the MBK array.

src/DSi.cpp
DesperateProgrammer: It is indeed DirectBoot when we are in the function (And thus to be removed)
However to the Variable, please see the comment at the LoadNAND function.

src/DSi.cpp
```diff
@@ -918,41 +1024,81 @@ void ARM9Write8(u32 addr, u8 val)
...
-            if (ptr)
+            if (addr >= NWRAMStart[0][0] && addr < NWRAMEnd[0][0])
             {
-                *(u8*)&ptr[addr & 0xFFFF] = val;
+                // Write to a bank is special, as it writes to all
+                // parts that are mapped and not just the highest priority
+                // See http://melonds.kuribo64.net/board/thread.php?pid=3974#3974
+                // so we need to iterate through all parts and write to all mapped here
+                u32 destPart = (addr >> 15) & NWRAMMask[0][0];
+                for (int page = 0; page < 4; page++)
```
RSDuck, line 1037: this seems to bit intense for every memory write. Would it be possible to do precalculate some of this similar to the way it's [done](https://github.com/Arisotura/melonDS/blob/master/src/GPU.cpp#L530) for [VRAM](https://github.com/Arisotura/melonDS/blob/master/src/GPU.h#L346-L382)?

src/DSi.cpp
```diff
@@ -659,28 +724,51 @@ void MapNWRAM_C(u32 num, u8 val)
...
-
-        if (NWRAMMap_C[oldval & 0x03][(oldval >> 2) & 0x7] == ptr)
-            NWRAMMap_C[oldval & 0x03][(oldval >> 2) & 0x7] = NULL;
+        NWRAMMap_C[0][part] = NULL;
+        NWRAMMap_C[1][part] = NULL;
+        NWRAMMap_C[2][part] = NULL;
     }
-
-    if (val & 0x80)
+    for (s8 part = 7; part >= 0; part--)
```
RSDuck, line 738: use int or s32, there's no point of using a tiny size as it's going to be sign extended in a 64-bit register anyway.

RSDuck (approved):
once this is done I need to update some JIT related memory stuff.

src/DSi.cpp
DesperateProgrammer: You would need to precalculate a mask for every 32kB within the image size and still would require to do a check for each page. The compare of a single bit (precalculated mappings) or by the 3 bits (mapping bits in page) should be equal in performance.

src/DSi.cpp
```diff
@@ -583,22 +621,35 @@ void MapNWRAM_A(u32 num, u8 val)
...
 
-    u8* ptr = &NWRAM_A[num << 16];
-
-    if (oldval & 0x80)
+    // When we only update the mapping on the written MBK, we will
+    // have priority of the last witten MBK over the others
+    // However the hardware has a fixed order. Therefor 
+    // we need to iterate through them all in a fixed order and update
+    // the mapping, so the result is independend on the MBK write order
+    for (u8 part = 0; part < 4; part++)
```
RSDuck, line 629: the same thing applies for here

src/DSi.cpp
RSDuck: ok well the most common case where there are no weird overlappings is going to be handled fastmem, so atleast with JIT+fastmem it won't make a difference.

DesperateProgrammer: I reduced the number of comparison (included the disable/assigned core) into the same compare. Thanks for letting me look over it again :-)

DesperateProgrammer:
I think, this is complete.
In the last days I ran several images, that showed no problems, other than those they already had.

PR #1154, opened 2021-07-13
Fallback to backup DSi Footer in NAND Image

DesperateProgrammer:
DSi MMC image loading now uses to the backup of the DSi Footer, if the footer at end of the file is absent or corrupted.
Changes output format of the ConsoleID to be consistent with all other tools (PRIx64 instead of PRIu64)

### DIFF
--- a/src/DSi.cpp
+++ b/src/DSi.cpp
@@ -457,15 +457,23 @@ bool LoadNAND()
         fread(nand_footer, 1, 16, SDMMCFile);
         if (memcmp(nand_footer, nand_footer_ref, 16))
         {
-            printf("ERROR: NAND missing nocash footer\n");
-            return false;
+            // There is another copy of the footer at 000FF800h for the case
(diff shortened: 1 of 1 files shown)


src/DSi.cpp
```diff
@@ -457,15 +457,23 @@ bool LoadNAND()
         fread(nand_footer, 1, 16, SDMMCFile);
         if (memcmp(nand_footer, nand_footer_ref, 16))
         {
-            printf("ERROR: NAND missing nocash footer\n");
-            return false;
+            // There is another copy of the footer at 000FF800h for the case
+            // that by external tools the image was cut off 
+            // See https://problemkaputt.de/gbatek.htm#dsisdmmcimages
+            fseek(SDMMCFile, 0x000FF800, SEEK_SET) ;
```
RSDuck, line 463:
```suggestion
            fseek(SDMMCFile, 0x000FF800, SEEK_SET);
```

PR #1162, opened 2021-07-23
Allow for using LLVM from Homebrew

nadiaholmquist:
This allows ugins the LLVM package from homebrew and also bundling its copy of libc++. As we build on 10.14 we are using a rather outdated version of LLVM and its C++ stdlib, which doesn't let us use things such as `std::filesystem`, this PR should address this.

CI still has to be adjusted to install llvm from homebrew and use the CMake toolchain file. **Do not merge yet.**


WaluigiWare64:
How much bigger does this make the app bundle?

nadiaholmquist:
About 1MB

PR #1164, opened 2021-07-24
Implement very basic WiFi packet capturing

rhowsl:
https://github.com/Arisotura/melonDS/issues/1152

this PR implements a new option in the melonDS config called "EnableWifiPacketCapturing". this will generate .pcap files inside a folder named "packetdumps", containing the wlan packets sent and recieved by melonDS when using Indirect mode.

sorry if the code looks bad, I don't usually code in C++/C.

### DIFF
--- a/src/Config.cpp
+++ b/src/Config.cpp
@@ -42,6 +42,7 @@ int DSiSDEnable;
 char DSiSDPath[1024];
 
 int RandomizeMAC;
+int EnableWifiPacketCapturing;
 
 #ifdef JIT_ENABLED
 int JIT_Enable = false;
@@ -67,6 +68,7 @@ ConfigEntry ConfigFile[] =
     {"DSiSDPath", 1, DSiSDPath, 0, "", 1023},
 
     {"RandomizeMAC", 0, &RandomizeMAC, 0, NULL, 0},
+    {"EnableWifiPacketCapturing", 0, &EnableWifiPacketCapturing, 0, NULL, 0},
 
 #ifdef JIT_ENABLED
     {"JIT_Enable", 0, &JIT_Enable, 0, NULL, 0},
--- a/src/frontend/qt_sdl/LAN_Capture.cpp
+++ b/src/frontend/qt_sdl/LAN_Capture.cpp
@@ -3,37 +3,41 @@
 #include <string.h>
 #include <iostream>
 #include <sys/time.h>
-#include <endian.h>
+
+#ifdef __WIN32__
+	#include <ws2tcpip.h>
+#else
+	#include <sys/socket.h>
+	#include <netdb.h>
+	#include <poll.h>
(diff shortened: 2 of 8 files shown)


src/frontend/qt_sdl/LAN_Capture.cpp
```diff
@@ -0,0 +1,89 @@
...
+	#include <netdb.h>
+	#include <poll.h>
+	#include <time.h>
+#endif
+
+#include "../types.h"
+#include "LAN_Capture.h"
+#include "Config.h"
+
+std::ofstream packetDumpOutput;
```
RSDuck, line 19: please use C files instead.

Also global variables in PascalCase

src/frontend/qt_sdl/LAN_Capture.h
```diff
@@ -0,0 +1,37 @@
+#pragma once
+
+#include <fstream>
+
+#include "../types.h"
+
+// Header for PCAP file
```
RSDuck, line 7: this doesn't need to be part of the public interface, it can go into LAN_Capture.cpp

src/Config.cpp
```diff
@@ -42,6 +42,7 @@ int DSiSDEnable;
 char DSiSDPath[1024];
 
 int RandomizeMAC;
+int EnableWifiPacketCapturing;
```
RSDuck, line 45: the setting needs to be exposed in the gui

src/frontend/qt_sdl/LAN_Socket.cpp
```diff
@@ -216,6 +219,12 @@ bool Init()
 
     Ctx = slirp_new(&cfg, &cb, nullptr);
 
+    if (Config::EnableWifiPacketCapturing)
+    {
+        LAN_Capture::Prepare();
```
RSDuck, line 224: why can't those hooks go into Platform.cpp?

rhowsl: Platform.cpp is the right place to place them, but weird stuff happens to the captured TX packets when I load the pcap file in wireshark:

image

this shouldn't be too hard for me to figure out

src/frontend/qt_sdl/LAN_Capture.cpp
rhowsl: C files?
also, just fixed this style, among some other things you've mentioned. gonna commit those next

RSDuck: https://en.cppreference.com/w/c/io

WaluigiWare64: you can read the new contributor guide: https://github.com/Arisotura/melonDS/blob/master/contributing.md

PR #1171, opened 2021-08-02
Fix for https://github.com/Arisotura/melonDS/issues/1169

DesperateProgrammer:
LoadNAND was storing the stage 2 bootloader, while NWRAM access was disabled by Bit25 of SCFG_EXT putting the stage 2 into old shared WRAM, while executing it later with the bit enabled and fetching from an empty NWRAM.

### DIFF
--- a/src/DSi.cpp
+++ b/src/DSi.cpp
@@ -1048,14 +1048,14 @@ void ARM9Write8(u32 addr, u8 val)
                 // parts that are mapped and not just the highest priority
                 // See http://melonds.kuribo64.net/board/thread.php?pid=3974#3974
                 // so we need to iterate through all parts and write to all mapped here
-                unsigned int destPartSetting = ((addr >> 13) & (NWRAMMask[0][0] << 3)) | 0x80;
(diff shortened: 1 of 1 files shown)


src/DSi.cpp
```diff
@@ -348,6 +348,13 @@ bool LoadNAND()
 {
     printf("Loading DSi NAND\n");
 
+    // Make sure NWRAM is accessible.
+    // The Bits are set to the startup values in Reset() and we might
+    // still have them on default (0) or some bits cleared by the previous
+    // run.
+    SCFG_EXT[0] |= (1 << 25) ;
```
RSDuck, line 355:
```suggestion
    SCFG_EXT[0] |= (1 << 25);
```

src/DSi.cpp
```diff
@@ -348,6 +348,13 @@ bool LoadNAND()
...
+    SCFG_EXT[1] |= (1 << 25) ;
```
RSDuck, line 356:
```suggestion
    SCFG_EXT[1] |= (1 << 25);
```

PR #1172, opened 2021-08-03
Fix Deadlock on unlaunch starting apps, when using software renderer

DesperateProgrammer:
Fix for a possible deadlock where RenderDone is waited for while the renderer is not rendering and waiting for the RenderStart semaphore


DesperateProgrammer:
Close because of a better approach by RSDuck

PR #1174, opened 2021-08-05
Fallback to FreeBIOS when BIOS files are not found. (v2)

asiekierka:
Updated version of https://github.com/Arisotura/melonDS/pull/787 with community feedback.


freebios/Makefile
```diff
@@ -0,0 +1,17 @@
...
+
+all:
+	$(AS) bios_common.S -DBIOS_ARM7 -march=armv4 -c -Wa,-asl=$(BIN_ARM7).list -o $(BIN_ARM7).o
+	$(AS) bios_common.S -DBIOS_ARM9 -march=armv5 -c -Wa,-asl=$(BIN_ARM9).list -o $(BIN_ARM9).o
+	$(OBJCOPY) -O binary $(BIN_ARM7).o $(BIN_ARM7).bin
+	$(OBJCOPY) -O binary $(BIN_ARM9).o $(BIN_ARM9).bin
+
+clean:
+	rm -f $(BIN_ARM7).bin $(BIN_ARM7).o $(BIN_ARM9).bin $(BIN_ARM9).o
+
```
WaluigiWare64, line 17: We don't need the Makefile.

asiekierka: We technically don't need the source code either, given the license - but it's good to have, and should be fairly easy to edit in the event of making custom patches in the future.

PR #1175, opened 2021-08-05
Generate a simple non-bootable firmware when not provided. (v2)

asiekierka:
Updated version of https://github.com/Arisotura/melonDS/pull/778 with community feedback.


src/Config.cpp
```diff
@@ -59,6 +67,14 @@ ConfigEntry ConfigFile[] =
     {"DLDIEnable", 0, &DLDIEnable, 0, NULL, 0},
     {"DLDISDPath", 1, DLDISDPath, 0, "", 1023},
 
+    {"FirmwareUsername", 1, FirmwareUsername, 0, "MelonDS", 63},
```
RSDuck, line 70: melonDS is the proper capitalisation

RSDuck (approved):

src/Config.cpp
asiekierka: Might be a good time to ask what the default username *should* be?

Arisotura: if it's going to be melonDS it should atleast be capitalized correctly :P 

other than that, I'm currently out of ideas, other than stupid shit like 'Kagamine Len'

maybe we should run a poll somewhere?

src/SPI.cpp
```diff
@@ -155,6 +159,52 @@ void Reset()
...
+    Firmware[UserSettings+0x50] = messageLength;
+}
+
+void Reset()
+{
+    if (Firmware) delete[] Firmware;
+    Firmware = NULL;
+
+    if (NDS::ConsoleType == 1)
+        strncpy(FirmwarePath, Config::DSiFirmwarePath, 1023);
```
DesperateProgrammer, line 194: Can we use sizeof()-1 as the limit?

asiekierka: This should be better.

src/SPI.cpp
```diff
@@ -142,19 +146,75 @@ void Reset()
...
+    }
     else
     {
-        f = Platform::OpenLocalFile(firmbkp, "wb");
-        fwrite(Firmware, 1, FirmwareLength, f);
         fclose(f);
     }
+}
+
+void LoadUserSettingsFromConfig() {
```
WaluigiWare64, line 174: `{` on newline please.

```suggestion
void LoadUserSettingsFromConfig()
{
```

PR #1181, opened 2021-08-16
SPU: Emulate SOUNDBIAS and 10-bit degrade

nadiaholmquist:
Also adds a config option to override the sound bitrate, defaulting to automatic based on model.

Needs more testing probably.

### DIFF
--- a/src/SPU.cpp
+++ b/src/SPU.cpp
@@ -192,11 +192,13 @@ void SetBias(u16 bias)
     Bias = bias;
 }
 
-void SetApplyBias(bool enable) {
+void SetApplyBias(bool enable)
+{
     ApplyBias = enable;
 }
 
-void SetDegrade10Bit(bool enable) {
+void SetDegrade10Bit(bool enable)
+{
     Degrade10Bit = enable;
 }
 
@@ -803,7 +805,7 @@ void Mix(u32 dummy)
 
     // Add SOUNDBIAS value
     // The value used by all commercial games is 0x200, so we subtract that so it won't offset the final sound output.
-    if (ApplyBias == true)
+    if (ApplyBias)
     {
         leftoutput += (Bias << 6) - 0x8000;
         rightoutput += (Bias << 6) - 0x8000;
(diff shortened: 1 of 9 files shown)


src/SPU.cpp
```diff
@@ -795,6 +805,21 @@ void Mix(u32 dummy)
...
+    // The original DS and DS lite degrade the output from 16 to 10 bit before output
+    if (Degrade10Bit)
+    {
+        leftoutput &= 0xFFFFFFC0;
+        rightoutput &= 0xFFFFFFC0;
+    }
+
+    // Add SOUNDBIAS value
+    // The value used by all commercial games is 0x200, so we subtract that so it won't offset the final sound output.
+    if (ApplyBias == true)
```
RSDuck, line 817: == true is unnecessary

src/SPU.cpp
```diff
@@ -190,6 +192,14 @@ void SetBias(u16 bias)
     Bias = bias;
 }
 
+void SetApplyBias(bool enable) {
```
RSDuck, line 195: brackets are wrong

WaluigiWare64: yep, it should look like this
```suggestion
void SetApplyBias(bool enable)
{
```

src/SPU.cpp
```diff
@@ -789,12 +799,28 @@ void Mix(u32 dummy)
...
+    // The original DS and DS lite degrade the output from 16 to 10 bit before output
+    if (Degrade10Bit)
+    {
+        leftoutput &= 0xFFFFFFC0;
+        rightoutput &= 0xFFFFFFC0;
+    }
+
     // OutputBufferFrame can never get full because it's
     // transfered to OutputBuffer at the end of the frame
     OutputBackbuffer[OutputBackbufferWritePosition    ] = leftoutput >> 1;
```
WaluigiWare64, line 826: Extra whitespace here.
```suggestion
    OutputBackbuffer[OutputBackbufferWritePosition] = leftoutput >> 1;
```

nadiaholmquist: I didn't even touch that line, Arisotura did it like that.

WaluigiWare64: well, wouldn't hurt to fix it :D

WaluigiWare64: just realised that's to make it look nice.

PR #1182, opened 2021-08-18
Fix a touchless issue in hybrid layout

2jun0:
In the previous commit, there was a touchless error on the hybrid screen.

image

This commit fix a touchless issue in hybrid layout.


RSDuck (approved):
looks ok to me, but I'm not going to merge any touchscreen related code until @Arisotura approves.

PR #1234, opened 2021-10-04
BizHawk Integration (v2)

CasualPokePlayer:
Second round of https://github.com/Arisotura/melonDS/pull/1233, realized I don't really need to use memory buffers for everything due to bizhawk's wbx system providing a virtual file system, although I need to avoid ever savestating while a file is open so heuristics are in place to close a file right away.

### DIFF
--- a/src/DSi.cpp
+++ b/src/DSi.cpp
@@ -80,8 +80,8 @@ DSi_NDMA* NDMAs[8];
 DSi_SDHost* SDMMC;
 DSi_SDHost* SDIO;
 
-FILE* SDMMCFile;
-FILE* SDIOFile;
+char* SDMMCFilePath;
+char* SDIOFilePath;
 
 u64 ConsoleID;
 u8 eMMC_CID[16];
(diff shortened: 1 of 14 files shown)


CasualPokePlayer:
> 
> 
> we're not really interested in (I haven't asked around for this time, maybe they changed their mind) having Bizhawk integration upstreamed.
> 
> Though we would merge some particular features or refactorings put into their own PRs like SuuperW did in the past. Currently there are still a few of them lying around (#973 is probably ready/was ready at some point and we should have merged it, the #975 is waiting on me, #959 was waiting on an answer, we don't want to merge #1009).

In retrospect I'm realizing some of these changes were a bit overneeded, they mostly ended up relating to how DSi/SD cards kept files opened which isn't exactly friendly to bizhawk. I probably will just remove those, which would keep most changes to just time related (which in practicality is all we need, states get handled by the wbx system, file i/o is somewhat cheated by wbx's virtual file system as long as it's kept read only, although firmware is questionable to deal with still, and I don't give a fuck about that hi-res crap).

CasualPokePlayer:
Alright, this should be the first of three: https://github.com/Arisotura/melonDS/pull/1235

PR #1235, opened 2021-10-07
SRAM API - ExportSRAM and SRAMIsDirty

CasualPokePlayer:
ExportSRAM will export SRAM to a buffer, and SRAMLength is return. Can provide NULL to just get SRAMLength.
SRAMIsDirty will return SRAMFileDirty, or false if no SRAM exists.

### DIFF
--- a/src/NDSCart.cpp
+++ b/src/NDSCart.cpp
@@ -531,7 +531,7 @@ int CartRetail::ImportSRAM(const u8* data, u32 length)
 int CartRetail::ExportSRAM(u8* data)
 {
     if (data)
-        memcpy(SRAM, data, SRAMLength);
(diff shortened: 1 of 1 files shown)


RSDuck:
SRAMFileDirty will be set to false after each frame via FlushSRAMFile, so shouldn't querying it be kind of useless?

CasualPokePlayer:
Re-reading the SRAM Manager code, I'm realizing that there is mostly already appropriate API for handling SRAM. ImportSRAM handles getting SRAM in, NeedsFlush is effectively a visible dirty flag, and FlushSecondaryBuffer can be used to ExportSRAM, and even SRAMLength is visible via SecondaryBufferLength, woops on me.

PR #1236, opened 2021-10-08
ARM7Peek8/ARM9Peek8 Support

CasualPokePlayer:
This is mainly for BizHawk for peeking at the memory bus. These alternative methods avoid updating emulator state during the read, needed when externally reading to avoid any sort of desyncs (we don't want any side effects when reading externally). Most of reading doesn't update emulator state anyways, so this is mostly copy paste of the read methods.

### DIFF
--- a/src/GPU3D.cpp
+++ b/src/GPU3D.cpp
@@ -3060,6 +3060,39 @@ void Write32(u32 addr, u32 val)
     printf("unknown GPU3D write32 %08X %08X\n", addr, val);
 }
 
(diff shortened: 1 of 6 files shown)


CasualPokePlayer:
I'm now realizing this kinda unneeded, the few unsafe values are not too important for peeking (and what you would get would probably not be accurate in the first place), and can just be filtered out externally to avoid unsafe reads. Also the code here is kinda ugly, so might as well keep ugliness out.

PR #1237, opened 2021-10-09
RTC Improvements

CasualPokePlayer:
Adds various features

Adds ability to either use emulated or real time.
Adds ability to select a fixed time on boot, or use current time instead.
Adds ability to change the clock via firmware (closes https://github.com/Arisotura/melonDS/issues/208)
And for any other frontends (e.g. BizHawk), the ability to set a custom RTC callback, which overrides time options.


src/NDS.cpp
```diff
@@ -177,6 +177,9 @@ void SetGBASlotTimings();
 
 bool Init()
 {
+    NumFrames = 0;
```
RSDuck, line 180: shouldn't this happen on reset?

CasualPokePlayer: This is so that time doesn't reset by just resetting (which a user will have to reset if they change the time in firmware).

RSDuck: we call Init only once when the program is started

CasualPokePlayer: Which is exactly the intent, to not reset frame counters unless the entire program is reset.

CasualPokePlayer:
> continuing from [#1237 (comment)](https://github.com/Arisotura/melonDS/pull/1237#discussion_r733865082) for better visibility.
> 
> yeah, but having to close and reopen the program is just not a good solution. The underlying problem is that melonDS has no distinction between those two kinds of resets.

I see. This is an odd case, it's preferable to not have time resetting when a reset happens, although having the frame counters set to 0 on init seems icky. Not sure how to proceed with this.

PR #1239, opened 2021-10-10
allow for not discarding bios/rom/firmware on reset

CasualPokePlayer:
This is mainly for BizHawk (and potentially other frontends which don't want to clear out these files on reset). Clearing out these files became problematic for BizHawk since that ended up blowing up states to above 64 MB, due to all that memory getting marked as dirty from the deallocation/reloading of these files.

### DIFF
--- a/src/NDS.cpp
+++ b/src/NDS.cpp
@@ -451,7 +451,7 @@ void SetupDirectBoot()
     SetWifiWaitCnt(0x0030);
 }
 
-void Reset()
+void Reset(bool discard)
 {
     FILE* f;
     u32 i;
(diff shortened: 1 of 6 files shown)


vadosnaprimer:
Would it be accurate to call this "emulate hard reset instead of resetting the whole emulator"?

CasualPokePlayer:
Thinking of it, this is kinda ugly and upstream probably doesn't want it unless it's actually implemented as some option for the user (don't eject cartridge on reset? also I guess just general smarter reloading of these files). And I'm lazy so eh.

It wouldn't be too hard to resolve any conflicts with upstream I guess regardless when pulling as long as changes are minimal on our side. Only https://github.com/Arisotura/melonDS/pull/1237 I can see getting merged anyways (considering that there are a lot of users which would like this kind of functionality anyways). I do have some code also [committed](https://github.com/CasualPokePlayer/melonDS/compare/5b415bae34c09653625861c03125af23beae5a36..0034e6bbc4dbf1c95aa04c6a41a33ab4758ae2d3) on my side for getting/setting regs, although I don't know if upstream would want these, so eh.

Issue #947, opened 2021-01-17
[Feature Request] Change Speed Limit


KostaSaizo7:
melonDS has a "Limit Framerate" option that limits the FPS at 60 (or 30 depending on the game). It would be helpful if the user could change the limit. 

~This would help regarding local multiplayer, as in its current state, is more stable when the game runs above 60fps. Unfortunately, turning "Limit Framerate" off, results in very high speeds and the game becomes unplayable.~

Dolphin has a similar option pictured below.
![dol](https://user-images.githubusercontent.com/39852873/104839042-71728980-58c7-11eb-81a2-b21c9b341e12.png)

RSDuck:
> This would help regarding local multiplayer, as in its current state, is more stable when the game runs above 60fps. Unfortunately, turning "Limit Framerate" off, results in very high speeds and the game becomes unplayable.

that's not the case. It's more stable because both instances always run (and are never stopped by the frame limiter). Thus local wireless would works similarly well on a slower computer where both instances only reach 60 fps.

KostaSaizo7:
Resolved by https://github.com/melonDS-emu/melonDS/commit/7dd4152d6711db76fbe2a2acffcc208a524bb25d
![image](https://github.com/melonDS-emu/melonDS/assets/39852873/01fe48a9-af9f-4ba3-a6fd-9d3fe61c8203)

Lyhtael:
Hello where is it possible to get a version with this menu ? I only found a 2022 version on the website 
ty by advance :)

acxz:
There has not been a release version with the commit that resolves this issue. You can either build from source or try downloading and running the nightly's until the next release. See: https://github.com/melonDS-emu/melonDS/issues/2094#issuecomment-2225906499 for links to the nightly's and more info on the next release.

PR #1247, opened 2021-10-23
Configure the FPS limit

klathem:
Adds a new setting to the ini to control the FPS to target when the framerate limit is on. Allows the user to set a specific playback speed instead of only choosing between "normal" and "as fast as possible".


acxz:
rev issue: #947

PR #1248, opened 2021-10-24
Implement MainRAM management dialog😁

2jun0:
*I'm not good at English. Please understand and read it.*

I Added RAM info dialog!

It will help modify main ram value!

RAM info dialog

This function is located in the System menu tab.

image

**But this function might have a small bug.😕**


src/frontend/qt_sdl/RAMInfoDialog.cpp
```diff
@@ -0,0 +1,296 @@
...
+
+ramInfo_ByteType RAMSearchThread::GetSearchByteType() const
+{
+    return SearchByteType;
+}
+
+std::vector<ramInfo_RowData>* RAMSearchThread::GetResults()
+{
+    return RowDataVector;
+}
```
RayyanAnsari, line 296: could you add a newline to the end of the file?
```suggestion
}

```

2jun0: Sure! (sorry for replying late)

src/frontend/qt_sdl/RAMInfoDialog.cpp
```diff
@@ -0,0 +1,302 @@
...
+extern EmuThread* emuThread;
+
+s32 GetMainRAMVAlue(const u32& addr, const ramInfo_ByteType& byteType)
+{
+    switch (byteType)
+    {
+    case ramInfo_OneByte:
+        return ((s8*)NDS::MainRAM)[addr&NDS::MainRAMMask];
+    case ramInfo_TwoBytes:
+        return ((s16*)NDS::MainRAM)[addr&NDS::MainRAMMask];
```
RSDuck, line 33: isn't this the wrong way around? Since addr is in bytes, it should be `*(s16*)(&NDS::MainRAM[addr&NDS::MainRAMMask])`?

src/frontend/qt_sdl/RAMInfoDialog.cpp
```diff
@@ -0,0 +1,302 @@
...
+    case ramInfo_FourBytes:
+        return ((s32*)NDS::MainRAM)[addr&NDS::MainRAMMask];
```
RSDuck, line 35: same here

RSDuck (changes requested):

src/frontend/qt_sdl/RAMInfoDialog.cpp
2jun0: Oh, I was like a stupid. 
You're right. I'll fix it right away

src/frontend/qt_sdl/RAMInfoDialog.cpp
2jun0: same here 👍

RSDuck:
ok I can't see anything wrong now, so I'll merge it!

Issue #1278, opened 2021-12-02
Is there a cli command for Boot firmware?


RokkumanX:
Hi

This isn't an issue per se and not a deal breaker by any means but I can't seem to find any information whether you can straightly boot firmware from cli or not and it's not covered in the Howto/FAQ either.

When the emulator is open there is an option to boot firmware and it works perfectly fine (I only use melonDS for DSi firmware/nand which I only have 1 game on that I care for)

Is there a cli command for starting melonDS and booting straight to firmware?

Otherwise this might be better suited as a Feature Request.

Thanks in advance!

nadiaholmquist:
There aren't currently any command line parameters, though adding them is planned.

RokkumanX:
Thank you very much for the feedback, now I know it's planned at least.

Have a good one!

RokkumanX:
I actually created a nice little circumvention to my dilemma, in my cli launcher I pointed to the game and unchecked the option Boot game directly in the melonDS GUI.

So when I launch the command I get straight into the firmware just as I want to.

If I launch the game (DSi) standalone with that option checked nothing happens and the emulator sorta softlocks/freezes, not that it really matters anyway (I know DSi emulation very experimental) because the specific game is already installed in my nand backup.

Cheers!

Issue #1108, opened 2021-05-18
Add more CLI options


rzumer:
Requests for additional command line flags/options to override default configuration parameters are pretty common, but there was no existing GitHub issue for it, so here's one.

bg2908:
Right now, I can't open ROM inside archive from lutris. A argument to do that would help me save some space from the ROM.

itsonlyMiRE:
leaving a comment as a +1/bump, because of course that will motivate the devs to prioritize CLI stuff :)

vgambier:
Being able to specify NDS/DSi mode via the CLI would also be a great addition.

PR #1280, opened 2021-12-03
Command-line interface

patataofcourse:
This PR improves the command-line interface, something which has been asked for in issues such as #1108 and #1278

Planned changes:
- [x] Moving everything related to reading command-line arguments (except QT arguments and portable/win32's EmuDirectory) to a `CLI` namespace
- [x] Loading DS/GBA ROMs from command-line (new implementation for new CLI)
- [x] Options (arguments with a `-s` / `--something` format)
    - [x] Help menu (`-h` / `--help`)
    - [ ] Open ROM from archive
    - [ ] Boot firmware (incompatible with loading a ROM, but should it boot the ROM instead or not do anything / return an error?)
    - [ ] Verbosity toggle*
    - [x] Start in fullscreen (`-f` / `--fullscreen`)
- [ ] Config overrides as options
    - [ ] Namespace for overridable config options
    - [ ] Figuring out how config being overridden should affect editing the actual config
    - [ ] Firmware settings: language and MAC address
    - [ ] Toggling usage of FreeBIOS or an external BIOS

This isn't a full list of features the CLI should have by any means, and I'm willing to implement whatever the core team considers necessary, plus some popular community suggestions. (Of course, all this is limited by my skill, and I might leave a stub if there's any option I can't implement.)

*Regarding the verbosity toggle, I'm not very sure what messages should always be displayed and which should be hiding behind that toggle, or even if melonDS would benefit from having multiple levels of verbosity, if having a verbosity toggle isn't worth it, etc... I'd appreciate input about this.

Feel free to submit reviews, everything that isn't implemented yet but is planned to be has a TODO comment.

### DIFF
--- a/src/frontend/qt_sdl/main.cpp
+++ b/src/frontend/qt_sdl/main.cpp
@@ -3354,17 +3354,14 @@ int main(int argc, char** argv)
 
     QObject::connect(&melon, &QApplication::applicationStateChanged, mainWindow, &MainWindow::onAppStateChanged);
 
-    if (CLI::DSRomPath != "")
-    {
-        int res = Frontend::LoadROM(CLI::DSRomPath, Frontend::ROMSlot_NDS);
+    if (!CLI::GBARomPath.isEmpty())
+        if (!ROMManager::LoadGBAROM(CLI::GBARomPath))
+            printf("Failed to load GBA ROM: %s\n", CLI::GBARomPath[0].toLatin1().cbegin());
 
-        // should this be moved to *after* the GBA ROM is loaded?
-        if (res == Frontend::Load_OK)
-            emuThread->emuRun();
-    }
 
-    if (CLI::GBARomPath != "")
-        Frontend::LoadROM(CLI::GBARomPath, Frontend::ROMSlot_GBA);
+    if (!CLI::DSRomPath.isEmpty())
+        if (!ROMManager::LoadROM(CLI::DSRomPath, true))
+            printf("Failed to load NDS ROM: %s\n", CLI::DSRomPath[0].toLatin1().cbegin());
 
     int ret = melon.exec();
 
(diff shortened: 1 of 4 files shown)


src/frontend/qt_sdl/main.cpp
```diff
@@ -2890,40 +2894,29 @@ int main(int argc, char** argv)
     Input::OpenJoystick();
 
     mainWindow = new MainWindow();
+    if (CLI::StartOnFullscreen) {
```
RSDuck, line 2897: braces go onto the next line

patataofcourse: missed that one somehow :sweat_smile:
fixed!

RayyanAnsari (changes requested):
I think this is a good feature, but instead of manually iterating and checking, we could use Qt.
https://doc.qt.io/qt-5/qcommandlineparser.html

patataofcourse:
> I think this is a good feature, but instead of manually iterating and checking, we could use Qt. https://doc.qt.io/qt-5/qcommandlineparser.html

Hadn't realized that was a thing - will change the implementation to use that instead, thanks!

mariomadproductions:
> Boot firmware (incompatible with loading a ROM, but should it boot the ROM instead or not do anything / return an error?)
I don't think its incompatible - in the GUI, if you load a ROM and choose to boot from firmware, it shows the ROM in the DS menu.

patataofcourse:
> I don't think its incompatible - in the GUI, if you load a ROM and choose to boot from firmware, it shows the ROM in the DS/i menu.

Wasn't a thing when I first made this PR- the system got changed, yeah

patataofcourse:
Will start a new PR 'cause the QT stuff and also the time I've had have changed my view on how I want this to be (...and I really don't want to have to touch the old messages in this PR)

PR #1319, opened 2021-12-29
don't link rt when oglrenderer disabled on apple

ccRlVDSyBZT1UK:


nadiaholmquist:
What is the purpose of this? Of course there is no need to link an unused library, but I'm curious when this is relevant?

ccRlVDSyBZT1UK:
@nadiaholmquist i'm working on a wasm build and i'm currently running both a native apple build and wasm build with the opengl renderer disabled and this was breaking the apple build for me at least.

nadiaholmquist:
Apparently the `rt` library doesn't actually exist at all on macOS and is ignored for POSIX compatibility, from what I can find on StackOverflow. I'm going to try compiling it without that and see if that's actually the case, and if it is I'll make it never link it on macOS.

RayyanAnsari:
`librt` was required for NetBSD, because it provided `shm_open` and `shm_unlink`.

PR #1326, opened 2022-01-01
Fix RTC date and time registers

joonhyungshin:
The current RTC implementation does not comply with [GBATEK](https://problemkaputt.de/gbatek.htm) (see **Date Registers** and **Time Registers** sections). This PR fixes some issues.

* The `year` is divided by 100, not subtracted by 100 (which seems to be a typo in the current master).
* Implements 12/24 mode as indicated by Status Register 1.
* Adds AM/PM flag.

### DIFF
--- a/src/RTC.cpp
+++ b/src/RTC.cpp
@@ -138,7 +138,7 @@ void ByteIn(u8 val)
                     Output[3] = timedata.tm_wday;
 
                     int hour = timedata.tm_hour;
-                    int pm_flag = hour < 12 ? 0x00 : 0x40;
+                    u8 pm_flag = hour < 12 ? 0x00 : 0x40;
                     if (!(StatusReg1 & 0x2)) hour %= 12;
                     Output[4] = BCD(hour) | pm_flag;
                     Output[5] = BCD(timedata.tm_min);
@@ -153,7 +153,7 @@ void ByteIn(u8 val)
                     localtime_r(&timestamp, &timedata);
 
                     int hour = timedata.tm_hour;
-                    int pm_flag = hour < 12 ? 0x00 : 0x40;
+                    u8 pm_flag = hour < 12 ? 0x00 : 0x40;
                     if (!(StatusReg1 & 0x2)) hour %= 12;
                     Output[0] = BCD(hour) | pm_flag;
                     Output[1] = BCD(timedata.tm_min);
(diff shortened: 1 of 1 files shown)


nadiaholmquist:
Have you verified the new behavior against hardware?

GBATek has many times actually been wrong during the development of melonDS, so best to verify what it says to be safe.

joonhyungshin:
Of course I did.

In fact, I found this issue when I was trying to RNG abuse a Pokemon game. Here are some related discussions.
- [Reddit thread](https://www.reddit.com/r/pokemonrng/comments/okiby3/need_help_with_bw_rng_on_melonds/?utm_medium=android_app&utm_source=share)
- [Smogon thread](https://www.smogon.com/forums/threads/past-gen-rng-research.61090/page-33#post-3976513)
I know that GBATEK is not always correct, but at least for the issues I strongly believe we should follow it. Similar code also appears in [DesMuME source](https://github.com/TASEmulators/desmume/blob/3ba38212281096f8417a0cce24396200923a2810/desmume/src/rtc.cpp#L133).

CasualPokePlayer:
>     * The `year` is divided by 100, not subtracted by 100 (which seems to be a typo in the current master).
This is really to do with how struct tm works. The year for struct tm has an epoch of 1900. But the DS has an epoch of 2000 for its clock. So what you get off of localtime is a year 100 years in the future in the eyes of the DS.

However, granted, a % 100 might be better for bounds reasons, although that's something to verify.

>     * Implements 12/24 mode as indicated by Status Register 1.
I don't know if this is actually a thing the DS actually had. No firmware ever gave the option to "switch" to 12 hour mode. The info seems more to be extracted from datasheets of the RTC chip. At most this would be something some homebrew could maybe do, but even then, to be verified (although the latter suggests it does exist).

>     * Adds AM/PM flag.
This does indeed exist, if RNGers insights are anything to go by.

On another note I have another [PR](https://github.com/Arisotura/melonDS/pull/1237) which does happen to implement the last thing (because some RNGers complained), although the previous 2 I left alone due to uncertainty. Of course that PR would need to be rebased anyways and has another problem not so easy to fix so eh.

joonhyungshin:
After some testing, I found that there are more works to do. For now, I'd better close this PR. Thank you all for your feedback!

PR #1328, opened 2022-01-01
add ids as a valid DS rom filetype

ccRlVDSyBZT1UK:
add ids as a valid DS rom filetype

ids is the extension for iQue DS roms


nadiaholmquist:
Will be added once the ROM loading refactor is merged.

Arisotura:
I incorporated this into #1333. let me know if I missed anything.

PR #1372, opened 2022-02-19
CI: Simplify Windows build workflow

Biswa96:
* Use pacboy to download dependencies instead of a separate step.
* Use ninja instead of make because ninja is faster and native in msys2.
* Use cmake options instead of changing working directory every steps.

### DIFF
--- a/.github/workflows/build-windows.yml
+++ b/.github/workflows/build-windows.yml
@@ -25,23 +25,32 @@ jobs:
       with:
           msystem: MINGW64
           update: true
-
-    - name: Install dependencies
-      run: pacman -Sq --noconfirm git make mingw-w64-x86_64-{cmake,mesa,SDL2,qt5-static,libslirp,libarchive,libepoxy,toolchain}
-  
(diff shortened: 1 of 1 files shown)


nadiaholmquist:
Look good to me.

Do you have a link to where these extra properties you are using are documented? I had no idea about them.

Biswa96:
msys2's docs: https://www.msys2.org/docs/ci/
GitHub Marketplace docs: https://github.com/marketplace/actions/setup-msys2
msys2-setup repo: https://github.com/msys2/setup-msys2

RayyanAnsari:
This PR looks good - it makes the workflow file simpler as well.

Though, with Ninja is there supposed to be no colour in the output?

Biswa96:
> with Ninja is there supposed to be no colour in the output?

Yes. There are some hacks https://medium.com/@alasher/colored-c-compiler-output-with-ninja-clang-gcc-10bfe7f2b949. But it is not msys2 specific.

PR #1377, opened 2022-02-25
fix typo in enum for screen sizing

Brankale:


RSDuck:
bop is funnier though :D

PR #1395, opened 2022-03-13
Add ability to change FPS Rate and move Framerate settings

### LINKED ISSUE #947: [Feature Request] Change Speed Limit

### LINKED ISSUE #1553: [Feature Request] Add a option to let the user manually change the fps to 100, 110, 120 etc.
So far the user is only able to run games at 144 fps or 370fps. The emulator needs to have a option where the user can choose to keep the fps at 100, 110, 120 etc. Not only that but the emulator running at 144 fps or higher makes the cpu and the gpu hotter than they should be, which makes the lifespan of the cpu and gpu shorter. Please give us the choice to manually change the fps to any number we want by implementing some sort of option.

acxz:
Closes #947
Closes #1553
Supersedes and closes #1247 

This PR adds the capability of changing the LimitFPS value via the GUI interface.
I built off of @klathem's logic and added the GUI portion that @RSDuck requested.

I also moved the checkbox for LimitFPS from the Menu into the DialogBox so that all framerate settings can be set in one DialogBox. WIth @nadiaholmquist's suggestions moved AudioSync to the new DialogBox as well.

Please feel free to check it out!

To see some pics:

menu-dialog

Notice that the Limit Framerate and Audio Sync checkbox is removed in the main Config Menu and the Framerate settings option is added.
When clicking the framerate settings options the following dialog box appears. When Limit Framerate checkbox is disabled then the spinbox is disabled as well. Either Audio Sync or Limit Framerate can be checked at once, not both.

framerate-dialog

### DIFF
--- a/src/frontend/qt_sdl/CMakeLists.txt
+++ b/src/frontend/qt_sdl/CMakeLists.txt
@@ -20,6 +20,7 @@ set(SOURCES_QT_SDL
     VideoSettingsDialog.cpp
     CameraSettingsDialog.cpp
     AudioSettingsDialog.cpp
+    FramerateSettingsDialog.cpp
     FirmwareSettingsDialog.cpp
     PathSettingsDialog.cpp
     MPSettingsDialog.cpp
--- a/src/frontend/qt_sdl/Config.cpp
+++ b/src/frontend/qt_sdl/Config.cpp
@@ -61,6 +61,7 @@ int GL_ScaleFactor;
 bool GL_BetterPolygons;
 
 bool LimitFPS;
+int FPSRate;
 bool AudioSync;
 bool ShowOSD;
 
@@ -251,6 +252,7 @@ ConfigEntry ConfigFile[] =
     {"GL_BetterPolygons", 1, &GL_BetterPolygons, false, false},
 
     {"LimitFPS", 1, &LimitFPS, true, false},
+    {"FPSRate", 0, &FPSRate, 60},
     {"AudioSync", 1, &AudioSync, false},
     {"ShowOSD", 1, &ShowOSD, true, false},
 
--- a/src/frontend/qt_sdl/Config.h
+++ b/src/frontend/qt_sdl/Config.h
@@ -105,6 +105,7 @@ extern int GL_ScaleFactor;
 extern bool GL_BetterPolygons;
 
 extern bool LimitFPS;
+extern int FPSRate;
 extern bool AudioSync;
 extern bool ShowOSD;
 
--- a/src/frontend/qt_sdl/EmuThread.cpp
+++ b/src/frontend/qt_sdl/EmuThread.cpp
@@ -86,7 +86,6 @@ EmuThread::EmuThread(QObject* parent) : QThread(parent)
     connect(this, SIGNAL(windowEmuPause()), mainWindow->actPause, SLOT(trigger()));
     connect(this, SIGNAL(windowEmuReset()), mainWindow->actReset, SLOT(trigger()));
     connect(this, SIGNAL(windowEmuFrameStep()), mainWindow->actFrameStep, SLOT(trigger()));
-    connect(this, SIGNAL(windowLimitFPSChange()), mainWindow->actLimitFramerate, SLOT(trigger()));
     connect(this, SIGNAL(screenLayoutChange()), mainWindow->panel, SLOT(onScreenLayoutChanged()));
     connect(this, SIGNAL(windowFullscreenToggle()), mainWindow, SLOT(onFullscreenToggled()));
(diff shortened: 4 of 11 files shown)


acxz:
@Arisotura @RSDuck added some picks to clarify what I mean. If you guys can take a look at this QoL PR that would be great!

acxz:
@Arisotura do you mind providing closure to this PR? Should be pretty easy to test.

nadiaholmquist:
It would probably be better to have this option in the video settings dialog instead of its own, the number of settings dialogs is already quite excessive in my opinion.

acxz:
@nadiaholmquist any more thoughts on this? Is coupling the rendering and framerate settings a requirement for this to be merged?

acxz:
@Arisotura @RSDuck @nadiaholmquist 
sorry for pinging this again, but just wanted to see if any you all can provide closure to this PR?

nadiaholmquist:
> @nadiaholmquist any more thoughts on this? Is coupling the rendering and framerate settings a requirement for this to be merged?

Another option is that it could go into interface settings. The audio sync option should probably be moved to the same place as these options as well, as it's another way to limit the emulation speed and doesn't really make sense to have remaining in the menu on its own.

acxz:
> The audio sync option should probably be moved to the same place as these options

I do like that idea and was originally planning on doing so, only reason I move it back was because if someone is looking for audio sync they may not immediately decide to look in the framerate dialog.

> go into interface settings

But with this suggestion, my concern is alleviated.

I'll go ahead and add a commit on top with this UI change. (Will be a while, tbh end of semester...)

acxz:
Also as I sidenote, by adding this commit, will this PR be merged in? GUI coding takes a while for me, and I'd like to make sure the design is final before I work on this since I don't want to redo my work.

SSUPII:
> > Another option is that it could go into interface settings.
> 
> I'm assuming that merging the Framerate dialog into an existing dialog is a requirement for this PR then?
> 
> So basically we have: Interface Settings dialog with a two new tabs 1st tab is called ?Activity? with the existing content 2nd tab is called Framerate with the new content in this PR
> 
> or Video Settings dialog with two new tabs 1st tab is called Render with the existing content 2nd tab is called Framerate with the new content in this PR
> 
> Can you comment/approve this design before I start working on it? I think I do like modifying the Video Settings dialog more since I feel like people will look for Framerate options there first.

I agree to use the Video settings dialog, even tough Audio Sync makes little sense in both. Unfortunately it's an option that _in practice_ limits the frame rate by itself.

bouceto:
Hello, I just want to ask what is the progress on this issue. Based on the comments I see, that the functionality already exists, but it is not merged into master yet, so it is not in MelonDS v 0.9.5. I would greatly appreciate if this could be put into production.

Thanks a lot,
Tomas

SSUPII:
> Hello, I just want to ask what is the progress on this issue. Based on the comments I see, that the functionality already exists, but it is not merged into master yet, so it is not in MelonDS v 0.9.5. I would greatly appreciate if this could be put into production.
> 
> Thanks a lot,
> Tomas

Some time has passed since the PR was made, so the code is likely needing an update.

GABRlEL:
Also wanted to mention here that, perhaps you might be able to use code from the android port? This has an option for Fast Forward Speed choices, which might achieve the same thing: https://github.com/rafaelvcaetano/melonDS-android-lib/ (fastForwardSpeedMultiplier)

acxz:
> > Hello, I just want to ask what is the progress on this issue. Based on the comments I see, that the functionality already exists, but it is not merged into master yet, so it is not in MelonDS v 0.9.5. I would greatly appreciate if this could be put into production.
> > Thanks a lot,
> > Tomas
> 
> Some time has passed since the PR was made, so the code is likely needing an update.

rebased

acxz:
> Also wanted to mention here that, perhaps you might be able to use code from the android port? This has an option for Fast Forward Speed choices, which might achieve the same thing: https://github.com/rafaelvcaetano/melonDS-android-lib/ (fastForwardSpeedMultiplier)

I don't think that would be necessary as the functionality has been already implemented.

Arignir:
This feature would be really appreciated!

Pokemon games are a bit too slow for my taste but if I uncap the FPS then it goes way too fast. Having the option to set a middle ground at 2x or 3x the FPS limit is exactly what I'm looking for 🙏

nadiaholmquist:
implemented by 7dd4152d6711db76fbe2a2acffcc208a524bb25d

Issue #1406, opened 2022-03-23
RAM search crashes when searching just after game startup


libjet:
After starting Pokémon Diamond, whenever I run RAM search it crashes the whole program and closes down. Sometimes the RAM search works but after closing it and running it again it crashes 100% of the time.
Happens on both internal and external BIOS, running windows x64 version.

RSDuck:
@2jun0 maybe you can take a look at this?

2jun0:
maybe this error was occurred because of table timer thread.
I'm fixing now and it will be fine

PR #1409, opened 2022-03-27
Fix: RAM search crashes when searching just after game startup

2jun0:
resolved:#1406

- This error seems to be caused when table update timer refer to empty table row data

- And renamed class, file name from RamInfo to RamSearch


Adamillo:
This sadly doesn't fix my problem itself where when I scroll and search for an address the game itself crashes. It would be very appreciated if you add an option where you could just find the address without having to scroll around

RSDuck:
also sorry, I wanted to review this PR, but somehow forgot it.

2jun0:
@Adamillo 
I will solve this error first and then consider it positively

PR #1434, opened 2022-05-03
Modernize CMake build system

nadiaholmquist:
These changes modernize the CMake build system to (hopefully) match newer best practices

Improvements include
 * Library linking is simpler and more automatic because of using imported targets
 * Multi-configuration builds should be supported (Ninja Multi-Config, Visual Studio, etc. generators)
 * Clean up build options using `cmake_dependent_option`
 * Let CMake do its job in more cases, like finding the math/dl libraries and detecting and enabling LTO support
 * Remove platform-specific kludges like the Fedora/flatpak LTO workaround and a bunch of Windows stuff
 * Simplify Windows static builds
 * Consistent formatting
 * Probably other minor stuff I forgot about already


cmake/FixInterfaceIncludes.cmake
```diff
@@ -0,0 +1,24 @@
+# The entire codebase quire reasonably does things like #include <SDL2/SDL.h> or <epoxy/gl.h>
```
autofire372, line 1: quite*

autofire372:
Builds fine on my end; no errors or warnings during cmake or the actual build.

nadiaholmquist:
Awesome, thanks for testing :)

PR #1485, opened 2022-07-20
prevent screensaver

pepper-jelly:
simple solution to disable screen blanking

1. init video and screensaver
2. disable screensaver


RSDuck:
is there no QT function for this?

RSDuck:
I guess if we have SDL, we might also just use it.

PR #1486, opened 2022-07-24
Added note on how to remove a mapping

k4kirin:
As noted here https://github.com/Arisotura/melonDS/issues/1461 , added a note on the frontend in the Input and hotkeys page that describes how to clear a mapping.


asutaa:
How do I apply this setting?

PR #1489, opened 2022-07-27
constexpr-s the DMA timing tables

### LINKED ISSUE #1488: constexpr or const the DMA timings
Hi, I noticed that the tables in DMA_Timings.h are never modified by the program but are not declared as constant (and I'm no DS hardware expert but it seems like those really shouldn't change at runtime). I think adding constexpr or const to the tables and making MRAMBurstTable a pointer to const would better communicate this and maybe slightly improve performance.

fixgoats:
small change that would close #1488

### DIFF
--- a/src/DMA.h
+++ b/src/DMA.h
@@ -96,7 +96,7 @@ private:
     bool IsGXFIFODMA;
 
     u32 MRAMBurstCount;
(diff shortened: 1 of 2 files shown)


RSDuck:
thank you

PR #1494, opened 2022-08-01
Changes some macros to constexpr functions, templates and lambdas

fixgoats:
It's generally a good goal to keep macro usage to a minimum, so I changed some macros to equivalent but type safe code.

- In WifiAP.cpp PWRITEs turned into constexpr function template + a few other constexpr functions. In WifiAP.h AP_MAC and AP_NAME turned into constexpr constants.
- In GPU_OpenGL.cpp SETVERTEX turned into a lambda.
- In DSi_SD.cpp SD_DESC turned into a constexpr function.
- In DSi.cpp BINARY_GOOD turned into a lambda.


RSDuck:
we still have to decide internally, whether we want so much of this new fangled C++ stuff or not.

fixgoats:
Yeah, I figured it'd be at least a bit controversial, especially since the macros evidently do the right thing already. But I thought it was at least worth suggesting.

PR #1500, opened 2022-08-18
Fix nocash Char Out debug register

QuinnPainter:
To match no$gba, this register should output any ASCII character written to it, not print the number as it currently does.


RSDuck:
a bit late, but looks good, thanks.

PR #1516, opened 2022-09-22
merge local_wifi

Arisotura:
June 2017 all over again

basically, local wifi but better


abx-dx:
Now multiple instance of MelonDS connect each other successfully (For example I confirm Pokemon Black 2-Pokemon HeartGold connection via DS Download Play using the new menu "Launch new instance").

Issue #1521, opened 2022-09-29
Better audio enhancements


Powerlated:
The DS isn't exactly known for not having problems with sound quality. Most of these issues originate from the DS's use of nearest-neighbor resampling in its audio mixer, which creates non-harmonic alias frequencies in the output stream of existing frequencies from the input stream. By default, melonDS emulates the mixer with nearest-neighbor resampling, which would be fine (and hardware-accurate) if melonDS didn't do another nearest-neighbor resampling step in the frontend from the core's output sample rate to the sample rate that SDL wants.

My goal is to improve melonDS's audio quality to hardware level at minimum, by introducing a proper resampler in the frontend, and eventually by replacing melonDS's resampling options (Linear, Cubic, Cosine) with methods better suited for DS audio.

**Why a sinc resampler?**

The sinc function `Sinc(x)=sin(x)/x` is the impulse response (IR) of the perfect low-pass filter (LPF). To avoid the problem of infinite latency from using an impulse response that has an infinite domain, we have to window the Sinc(x) function so that it has a reasonable length. Using a resampler with a windowed sinc function scaled so that it is the IR of an LPF it at the DS's Nyquist rate, we can avoid nonharmonic aliases in the output stream while preserving the harmonic image frequencies of the input sample that appear in the output stream above the input sample's Nyquist rate. Most of these harmonic image frequencies are desirable since they add texture to the sound that the samples themselves don't have high enough of a sample rate to represent. Because the linear, cubic, and cosine resamplers dull these upper harmonics too much (this is especially noticable in games that use low sample-rate sounds like Pokémon Gen IV), I plan to replace them with a sinc resampler, along with an optional 96db/octave low-pass filter set at the sample's Nyquist rate to remove some of the high image harmonics that add to perceived harshness rather than adding a fuller sound.

- [ ] Add a proper resampler to the frontend
- [ ] Add Sinc resampling
- [ ] Add optional low-pass filter after Sinc resampling

RSDuck:
hey someone who actually knows signal theory! A warm welcome from us, we appreciate any help with this.

Arisotura:
my brain is gone

if anyone knows better than me about all this, I'll let them have fun :P

carmiker:
[...]

PR #1523, opened 2022-09-30
Audio interpolation rework

Powerlated:
See #1521

### DIFF
--- a/src/NDS.cpp
+++ b/src/NDS.cpp
@@ -670,10 +670,10 @@ void Reset()
         degradeAudio = false;
     }
 
-    int bitrate = Platform::GetConfigInt(Platform::AudioBitrate);
-    if (bitrate == 1) // Always 10-bit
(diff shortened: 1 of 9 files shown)


nadiaholmquist:
Ah lol the bitrate thing, I asked in the Discord if anyone knew what the correct term was but didn't really get a response, so that's what ended up being committed. You can probably tell I'm not the most knowledgeable about audio...

Only realized the mistake after the release with that in it was shipped, oops.

Arisotura:
why is this named 'Full audio rework' when it's just about changing one term

I'm all for using the proper terms, tho; just want to hear about this bit, are there more changes planned to this pull request? should we wait before merging it?

nadiaholmquist:
See the issue linked in the OP, I think this is work in progress.

Arisotura:
'aight -- I'll wait for their instructions before merging this

RayyanAnsari:
probably a good idea to mark this as a draft

PR #1532, opened 2022-10-14
Wifi: Fix -Wformat

carmiker:
This PR fixes -Wformat warnings.

The first change adds what I believe is the correct variable to the printf statement -- there is one less argument than there should be.

The other changes may be architecture dependent. I am on a musl-based Linux amd64 distro.

### DIFF
--- a/src/Wifi.cpp
+++ b/src/Wifi.cpp
@@ -872,7 +872,7 @@ bool ProcessTX(TXSlot* slot, int num)
             }
 
             if ((num != 5) && (RAM[slot->Addr+4] > 0))
-                printf("SLOT %d RETRY COUNTER %d\n", RAM[slot->Addr+4]);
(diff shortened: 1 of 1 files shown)


CasualPokePlayer:
The second would be wrong on Windows where long is 32 bits rather than 64 bits. long long is needed for 64 bits on Windows (and would be fine on Linux anyways as that's also 64 bits, just a warning). You probably could just cast the variable to `unsigned long long` to supress the warning on Linux.

carmiker:
@CasualPokePlayer I've just force pushed to only include the first fix.

carmiker:
@Arisotura @nadiaholmquist any thoughts?

orbea:
This fixes a QA notice on Gentoo, can it please be merged?

Issue #1393, opened 2022-03-12
Allow passing archives on the command line


Janfel:
As of melonDS `0.9.4`, you can open an archive either using the dialogue or by dragging it onto the main window. So you would expect you could just as easily pass an archive as command line argument, but in that case the archive is not recognized and extracted, it is just treated as an invalid ROM.

Adding this functionality would allow the user to simply right-click any compressed ROM and “Open with melonDS”, which is not possible at the moment.

Arisotura:
this is actually possible, it may not be obvious but it works the same way as recent file paths

for example:

`melonDS.exe "C:/example/folder/duperarchive.zip|dupermario.nds"`

or under Linux

`melonDS "/example/folder/duperarchive.zip|dupermario.nds"`

will load file dupermario.nds from archive duperarchive.zip when starting melonDS.

Arisotura:
however that won't work if you just open an archive with melonDS directly, yeah. I see what you want to do there...

I could make it so that just specifying the archive path does all the work if a given archive contains only one file.

Janfel:
I sort of implemented that in #1394, but I'm still not happy with those changes alone. Ideally I would prefer something like this:

1. Both single-member-archives and compressed ROMs (e.g. `game.nds.gz`) are treated just like uncompressed ROMs on the command line and in the recent list. There should be no difference in user experience between those three.
  This would require some more sophisticated filename extension parsing (even though one could get around that by just listing all supported combinations for now), and the use of `libarchive`s "raw" reader module.
2. The `archive.zip|game.nds` syntax is removed in favor of a command line option for selecting the archive member. When this option is not given, a single-member-archive will be treated as above, while a multiple-member-archive will prompt the user with the archive member selection box.
  This would require something like `QCommandLineParser` as well as some changes to the recent list.

While the `archive.zip|game.nds` syntax is a handy addition, it breaks the fundamental assumption that the path being passed in (e.g. `/example/foo|bar/game.nds`) will be handled by the program as-is.

Those are some ideas I had when looking over the code. And while I would really like to see point 1 implemented, the core functionality is obviously more important.

Janfel:
Thanks. Please let me elaborate on why I think that 2) would be necessary complexity.

[...]

PR #1546, opened 2022-10-27
CLI 2.0: Electric Boogaloo

### LINKED ISSUE #1108: Add more CLI options

### LINKED ISSUE #1459: Fullscreen
One question, do you plan to add the fullscreen when a game starts or give it a command line so that it runs better? (I say this more for the frontend that uses these elements) I know that it is currently but only as a form of hotkey, an option could be worth it

patataofcourse:
New CLI implementation cause hyperfixation go brr. Still a WIP but should be usable. Uses Qt's QCommandLineParser this time!

Supersedes #1280 and #1402.
Fixes #1108, fixes #1459 (and probably #1393 too)

Seems to build on both Qt5 and Qt6 right now, but I can't actually test outside of Linux so enabling Actions would be appreciated (apparently I can't have them outside master on my fork).

Feel free to review, although may be better to wait 'till I'm done.

Planned shit:
- [x] Load ROMs as usual
  - [x] Load NDS and GBA roms
  - [x] Reimplement (undocumented) support for x.zip|y.nds syntax
  - [x] Load from archive (+ try to find .nds file without having to explicitly give --archive file)
  - [x] --archive-file option
- [x] Choose whether you want to run the game immediately/boot firmware or not
  - Using option `--boot=<always/never/auto>`; `always` and `never` being as expected and `auto` running the game only if a DS ROM is supplied
- [x] Start on fullscreen
- [ ] whatever else the peoples want me to add as long as it's not too much for me

### DIFF
--- a/src/frontend/qt_sdl/CLI.cpp
+++ b/src/frontend/qt_sdl/CLI.cpp
@@ -1,0 +1,91 @@
+/*
+    Copyright 2021-2022 melonDS team
+
+    This file is part of melonDS.
+
+    melonDS is free software: you can redistribute it and/or modify it under
+    the terms of the GNU General Public License as published by the Free
+    Software Foundation, either version 3 of the License, or (at your option)
+    any later version.
+
+    melonDS is distributed in the hope that it will be useful, but WITHOUT ANY
+    WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+    FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License along
+    with melonDS. If not, see http://www.gnu.org/licenses/.
+*/
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include <QStringList>
+#include <QCommandLineParser>
+
+#include "CLI.h"
+
+namespace CLI
+{
+
+QStringList DSRomPath = {};
+QStringList GBARomPath = {};
+bool StartOnFullscreen = false;
+
+void ManageArgs(int argc, char** argv)
+{
+
+    int posArgCount = 0;
+    
+    for (int i = 1; i < argc; i++)
+    {
+        char* arg = argv[i];
+        bool isFlag;
+        if (arg[0] == '-')
+        {
+            if (!strcasecmp(arg, "-h") || !strcasecmp(arg, "--help"))
+            {
+                //TODO: QT options
+                printf(
+                   "usage: melonDS [options] ... [dspath] [gbapath]\n"
+                   "Options:\n"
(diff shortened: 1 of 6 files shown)


src/frontend/qt_sdl/CLI.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+    options->fullscreen = parser.isSet("fullscreen");
+    
+    QStringList posargs = parser.positionalArguments();
+    switch (posargs.size())
+    {
+        default:
+            printf("Too many positional arguments; ignoring 3 onwards\n");
+        case 2:
+            options->gbaRomPath = QStringList(posargs[1]);
```
RSDuck, line 60: fallthrough?

src/frontend/qt_sdl/CLI.cpp
```diff
@@ -0,0 +1,136 @@
...
+        if (options->dsRomPath.isEmpty())
+        {
+            options->errorsToDisplay += "Option -a/--archive-file given, but no archive specified!";
+        }
+        else
+        {
+            options->dsRomPath += parser.value("archive-file");
+        }
+    } 
+    else if(!options->dsRomPath.isEmpty())
```
RSDuck, line 98: missing space between the if and (

src/frontend/qt_sdl/CLI.cpp
```diff
@@ -0,0 +1,136 @@
...
+        if (options->gbaRomPath.isEmpty())
+        {
+            options->errorsToDisplay += "Option -A/--archive-file-gba given, but no archive specified!";
+        }
+        else
+        {
+            options->gbaRomPath += parser.value("archive-file-gba");
+        }
+    }
+    else if(!options->gbaRomPath.isEmpty())
```
RSDuck, line 120: here too

src/frontend/qt_sdl/main.cpp
```diff
@@ -3335,6 +3356,12 @@ int main(int argc, char** argv)
     Input::OpenJoystick();
 
     mainWindow = new MainWindow();
+    if (options->fullscreen)
+    {
+        // onFullscreenToggled is private and I don't know if I should copy what's inside it or make it public
```
RSDuck, line 3361: add a public function something like setFullscreen and put the logic from onFullscreenToggled there so that it's not duplicated.

RSDuck (approved):

Tehdead:
Hey thank you! This was definitely awesome to included. That said could we get
-c , --console <ds/dsi>

Thanks again!

patataofcourse:
that would require overriding settings, which i'm not sure how to go about- but i can tackle it sometime

PR #1557, opened 2022-11-06
Fix: Some errors in Ram Search

2jun0:
### Avoiding too many memory updates (#1406)
Each time a row in the table was updated, the RAM was updated.
So accessing memory that is frequently modified will break your program.

Currently the RAM is updated when the user changes the value.

### Fixed error that some rows are partially invisible in mac (like below)
image

### Fixed an error where rows were accumulated
There was an error where existing information was accumulated even when searching for a new one.

## 📌 Others
Updated default search byte type from One to Four.

### DIFF
--- a/src/frontend/qt_sdl/RAMInfoDialog.cpp
+++ b/src/frontend/qt_sdl/RAMInfoDialog.cpp
@@ -201,16 +201,12 @@ void RAMInfoDialog::on_ramTable_itemChanged(QTableWidgetItem *item)
     s32 newValue = item->text().toInt();
 
     if (rowData.Value != newValue)
+    {
         rowData.Value = newValue;
-}
 
-void RAMInfoDialog::on_ramTable_currentItemChanged(QTableWidgetItem *item)
-{
-    ramInfo_RowData& rowData = SearchThread->GetResults()->at(item->row());
-    s32 newValue = item->text().toInt();
-
-    if (rowData.Value != newValue)
-        NDS::MainRAM[rowData.Address&NDS::MainRAMMask] = (u32)newValue;
+        if (item->isSelected())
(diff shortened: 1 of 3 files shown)


TheGag96:
I was having some crashing issues in melonDS and tried your fix hoping it would solve my problme. However, I still get a crash on opening RAM Search while in-game playing Animal Crossing: Wild World:

```
terminate called after throwing an instance of 'std::out_of_range'
  what():  vector::_M_range_check: __n (which is 2147483628) >= this->size() (which is 4194304)
```

Would you like me to help you debug this further? Thanks for trying, at the very least!!

src/frontend/qt_sdl/RAMInfoDialog.cpp
```diff
@@ -182,10 +200,15 @@ void RAMInfoDialog::on_radiobtn4bytes_clicked()
...
+    s32 newValue = item->text().toInt();
 
-    if (rowData.Value != itemValue)
-        rowData.SetValue(itemValue);
+    if (rowData.Value != newValue)
+    {
+        rowData.Value = newValue;
+
+        if (item->isSelected())
+            NDS::MainRAM[rowData.Address&NDS::MainRAMMask] = (u32)newValue;
```
RSDuck, line 210: this does not take the size of the value to be changed into account.

src/frontend/qt_sdl/RAMInfoDialog.cpp
```diff
@@ -233,18 +256,17 @@ void RAMSearchThread::run()
     SearchRunning = true;
     u32 progress = 0;
 
-    // Pause game running
```
RSDuck, line 236: why did you remove the pausing and unpausing?

Ukalus:
I have the same issue, would be cool if it could be fixed

priverop:
Any chance to continue with this fix? @2jun0

PR #1574, opened 2022-11-30
Audio Input Device Setting

### LINKED ISSUE #1571: [Feature Request] Mic source selection
Hello

In audio settings, you can choose an external (physical) mic attached to the computer:
image

However, i suppose it use the default one, the one selected in the OS as the primary mic source.

The point of the request is to allow to choose between the available mic sources, between the "Default/Primary" one, and all the other available in the system to override the default selection.

In Windows is common to have that list in other apps, in other OS like linux i don't know if it's possible to detect all the available sources for microphone, or the only option is to write the path to the device.

Thanks in advance

falsidge:
Adds an audio input combobox, which allows users to select an audio device.  Closes #1571 
image
Also some edits to readme to include other dependencies that are needed for windows build.

### DIFF
--- a/src/frontend/qt_sdl/AudioSettingsDialog.cpp
+++ b/src/frontend/qt_sdl/AudioSettingsDialog.cpp
@@ -40,7 +40,7 @@ AudioSettingsDialog::AudioSettingsDialog(QWidget* parent, bool emuActive) : QDia
     setAttribute(Qt::WA_DeleteOnClose);
 
     oldInterp = Config::AudioInterp;
-    oldBitrate = Config::AudioBitrate;
+    oldBitDepth = Config::AudioBitDepth;
     oldVolume = Config::AudioVolume;
     oldDSiSync = Config::DSiVolumeSync;
 
@@ -50,10 +50,10 @@ AudioSettingsDialog::AudioSettingsDialog(QWidget* parent, bool emuActive) : QDia
     ui->cbInterpolation->addItem("Cubic");
     ui->cbInterpolation->setCurrentIndex(Config::AudioInterp);
 
-    ui->cbBitrate->addItem("Automatic");
-    ui->cbBitrate->addItem("10-bit");
-    ui->cbBitrate->addItem("16-bit");
-    ui->cbBitrate->setCurrentIndex(Config::AudioBitrate);
+    ui->cbBitDepth->addItem("Automatic");
+    ui->cbBitDepth->addItem("10-bit");
+    ui->cbBitDepth->addItem("16-bit");
+    ui->cbBitDepth->setCurrentIndex(Config::AudioBitDepth);
 
     bool state = ui->slVolume->blockSignals(true);
     ui->slVolume->setValue(Config::AudioVolume);
@@ -90,7 +90,7 @@ AudioSettingsDialog::AudioSettingsDialog(QWidget* parent, bool emuActive) : QDia
     {
         ui->lblInstanceNum->setText(QString("Configuring settings for instance %1").arg(inst+1));
         ui->cbInterpolation->setEnabled(false);
-        ui->cbBitrate->setEnabled(false);
+        ui->cbBitDepth->setEnabled(false);
         for (QAbstractButton* btn : grpMicMode->buttons())
             btn->setEnabled(false);
         ui->txtMicWavPath->setEnabled(false);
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -76,7 +76,7 @@ As for the rest, the interface should be pretty straightforward. If you have a q
 If everything went well, melonDS and the libraries it needs should now be in the `dist` folder.
 
 #### Static builds (without DLLs, standalone executable)
-5. Install dependencies: `pacman -S mingw-w64-x86_64-{cmake,SDL2,toolchain,qt5-static,libslirp,libarchive}`
+5. Install dependencies: `pacman -S mingw-w64-x86_64-{cmake,ninja,SDL2,toolchain,qt5-static,libslirp,libarchive}`
```
RayyanAnsari, line 79: Adding `ninja` as a dependency shouldn't be needed - the instructions use `make`.

src/frontend/qt_sdl/AudioSettingsDialog.cpp
```diff
@@ -17,6 +17,13 @@
 */
 
 #include <stdio.h>
+#include <QtGlobal>
+#if QT_VERSION >= 0x060000
```
RayyanAnsari, line 21: Use the `QT_VERSION_CHECK()` macro instead.
https://doc.qt.io/qt-6/qtglobal.html#QT_VERSION_CHECK

src/frontend/qt_sdl/AudioSettingsDialog.cpp
```diff
@@ -68,6 +75,35 @@ AudioSettingsDialog::AudioSettingsDialog(QWidget* parent) : QDialog(parent), ui(
     ui->txtMicWavPath->setEnabled(iswav);
     ui->btnMicWavBrowse->setEnabled(iswav);
 
+    bool isext = (Config::MicInputType == 1);
+    ui->cbMic->setEnabled(isext);
+
+    #if QT_VERSION >= 0x060000
```
RayyanAnsari, line 81: Same here.

RayyanAnsari (changes requested):

src/frontend/qt_sdl/AudioSettingsDialog.cpp
```diff
@@ -17,6 +17,13 @@
 */
 
 #include <stdio.h>
+#include <QtGlobal>
```
RayyanAnsari, line 20: Add a new line between the `std` and `Qt` header includes.

src/frontend/qt_sdl/AudioSettingsDialog.cpp
```diff
@@ -68,6 +75,35 @@ AudioSettingsDialog::AudioSettingsDialog(QWidget* parent) : QDialog(parent), ui(
...
+        const auto deviceInfos = QMediaDevices::availableDevices(QAudioDevice::Input);
+        for (const QAudioDevice &deviceInfo : deviceInfos)
+        {
+            ui->cbMic->addItem(deviceInfo.description());
+        }
+        
+        if (Config::MicDevice == "")
+        {
+            Config::MicDevice = QAudioDevice::defaultInputDevice().deviceName().toStdString();
+        }
```
RayyanAnsari, line 91: This bit doesn't work in Qt 6 - see the macOS CI (which uses Qt 6 instead of Qt 5 like the rest).

README.md
```diff
@@ -65,8 +65,8 @@ As for the rest, the interface should be pretty straightforward. If you have a q
    cd melonDS
    ```
 #### Dynamic builds (with DLLs)
-5. Install dependencies: `pacman -S mingw-w64-x86_64-{cmake,SDL2,toolchain,qt5-base,qt5-svg,qt5-multimedia,libslirp,libarchive}`
-6. Compile:
+1. Install dependencies: `pacman -S mingw-w64-x86_64-{cmake,SDL2,toolchain,qt5-base,qt5-svg,qt5-multimedia,qt5-tools,libslirp,libarchive}`
+2. Compile:
```
RayyanAnsari, line 69: There's no need to change the numbers here - they're labelled steps 5 and 6 because there are 4 steps before that you need to do, regardless of whether you are making a static or dynamic build.

README.md
RayyanAnsari, line 79: The `ninja` dependency doesn't need to be here.

src/frontend/qt_sdl/AudioSettingsDialog.cpp
```diff
@@ -68,6 +76,35 @@ AudioSettingsDialog::AudioSettingsDialog(QWidget* parent) : QDialog(parent), ui(
...
+        {
+            ui->cbMic->addItem(deviceInfo.deviceName());
+        }
+
+        if (Config::MicDevice == "")
+        {
+            Config::MicDevice = QAudioDeviceInfo::defaultInputDevice().deviceName().toStdString();
+        }
+    #endif
+    ui->cbMic->setCurrentText(QString(Config::MicDevice.c_str()));
```
RayyanAnsari, line 105: You can convert directly from `std::string` to `QString` without going through `c_str()`:
```suggestion
    ui->cbMic->setCurrentText(QString::fromStdString(Config::MicDevice));
```

RayyanAnsari (changes requested):

patataofcourse:
I've noticed that on Linux it has tons of audio devices, most of which I don't even recognize and are probably internal to the system. Any way to weed those out?
image

KostaSaizo7:
I noticed something on windows. There is an icon on taskbar that shows when melonDS uses the microphone. 
Screenshot 2022-12-06 211122
In master the microphone is enabled: A) when selecting audio settings>external microphone>ok, B) when launching any game. With this pr, only A enables the microphone.

falsidge:
> I noticed something on windows. There is an icon on taskbar that shows when melonDS uses the microphone. Screenshot 2022-12-06 211122 In master the microphone is enabled: A) when selecting audio settings>external microphone>ok, B) when launching any game. With this pr, only A enables the microphone.

Sometimes the icon is hidden. You may need to unhide it. I added a check if the config is empty to use default or (null), which may have caused a problem.

falsidge:
> I've noticed that on Linux it has tons of audio devices, most of which I don't even recognize and are probably internal to the system. Any way to weed those out? image

Maybe checking if they support a certain audio format will narrow the list?

Issue #318, opened 2018-12-20
Suggestion: Rich Presence Support for Discord


nathanielcwm:
I think it might be cool having the ability to display like the game you are currently playing, maybe the melonds version, game region and other info in your discord profile. Definitely shouldn't be higher priority over more important things but it would be nice.

StapleButter:
not likely to be a thing anytime soon considering my stance on Discord

nathanielcwm:
Ok

PR #1579, opened 2022-12-06
Add Discord Rich Presence

yeah-its-gloria:
This would implement the request from #318, just, uhh, 4 years late


yeah-its-serena:
This is about as clean as it gets, I gave `building my own implementation` a shot and quickly realized it is not worth the effort:tm:

src/Platform.h
```diff
@@ -82,7 +82,7 @@ enum ConfigEntry
     Firm_Message,
     Firm_MAC,
 
-    AudioBitrate,
+    AudioBitrate
```
RayyanAnsari, line 85: Why has the `,` been removed here?

src/frontend/qt_sdl/DiscordRPC.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+#include "Config.h"
+#include "Platform.h"
+#include <stdio.h>
+#include <string.h>
+#include <time.h>
+
+#include <discord_rpc.h>
+
+const char* APP_ID          = "1049597419152744458";
```
RayyanAnsari, line 29: How has this number been derived?

src/frontend/qt_sdl/InterfaceSettingsDialog.ui
```diff
@@ -20,47 +20,61 @@
...
+     </property>
+    </widget>
+   </item>
+   <item row="4" column="0" colspan="4">
+    <widget class="QCheckBox" name="cbDiscordTrackTime">
+     <property name="text">
+      <string>Track time on Discord</string>
+     </property>
+    </widget>
+   </item>
```
RayyanAnsari, line 77: This should probably be in its own `QGroupBox`.

RayyanAnsari (changes requested):

src/Platform.h
yeah-its-serena: I had previously exposed the Discord related settings there thinking that's just where settings belong, until I realized they don't, so I removed them again and that comma went away in the process

src/frontend/qt_sdl/DiscordRPC.cpp
yeah-its-serena: That's just the [snowflake](https://discord.com/developers/docs/reference#snowflakes) of the [application](https://discord.com/developers/docs/intro#bots-and-apps) (doesn't go deep into detail about what it is, but an application is just, well, an application on Discord) needed for Rich Presence to function.

It's just needed because RPC requires it to function, and doesn't have much aside from the melonDS icon as part of its "Rich Presence Assets" (you upload an image, give it a name and then can use it as either a small or large image on profiles' activity thing).

patataofcourse: basically you csn't really use RPC with any text you want, you need to make an application for it in the discord developer page

yeah-its-gloria: Do you still have questions about this or is this resolved?

yeah-its-gloria:
are there any remaining questions or issues with this PR?

yeah-its-gloria:
It's been a while since anything happened and I have other things to deal with, therefore I'm closing it, since I can't keep this code functional

If anyone wants to use this code to any extent, have fun :)

PR #1581, opened 2022-12-13
DSi: add option to boot the full boot ROMs

PoroCYon:
Added a config option for this so that this can be enabled or disabled, while for now it's not very useful for most people.

However, more importantly, it fixes the default SPI firmware/settings image as well, as it was slightly wrong before (though it's only noticeable when you try to get the bootroms to boot anything with it).

Also added IO regs for DSi GPIO, but those don't do anything yet.

### DIFF
--- a/src/DSi.cpp
+++ b/src/DSi.cpp
@@ -82,6 +82,14 @@ DSi_SDHost* SDIO;
 u64 ConsoleID;
 u8 eMMC_CID[16];
 
+// FIXME: these currently have no effect (and aren't stored in a savestate)
+//        ... not that they matter all that much
+u8 GPIO_Data;
+u8 GPIO_Dir;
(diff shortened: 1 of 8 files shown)


src/DSi.cpp
```diff
@@ -82,6 +82,14 @@ DSi_SDHost* SDIO;
 u64 ConsoleID;
 u8 eMMC_CID[16];
 
+// FIXME: these currently have no effect (and aren't stored in a savestate)
+//        ... not that they matter all that much
+u8 GPIO_Data;
```
RSDuck, line 87: these registers don't seem to be set back again to 0 on Reset

PoroCYon: fixed (data actually resets to 0xff instead of 0, I've observed this with a scope and logic analyzer)

Issue #73, opened 2017-06-20
Guitar Hero - On Tour (USA) (En,Fr) missing guitar grip


benderscruffy:
cannot play the game as you  need the guitar grip
![guitar hero - on tour usa en fr -0](https://user-images.githubusercontent.com/2696404/27332406-573f446a-55f4-11e7-954a-bcb9eceada89.png)

Buenia0:
Is anyone still working on this issue? Because I'm currently working on a custom fork of melonDS with support for this addon...

Arisotura:
fixed with asie's PR

Issue #49, opened 2017-06-06
SVatG Loves You by St. Vincent and the Grenadines demo black screen


benderscruffy:
http://www.pouet.net/prod.php?which=59692
nothing but black screens

DarkSwan86:
Still present on ad58a68 with a lot of:

unknown ARM9 IO read16 040002A2 0202005E
unknown ARM9 IO read16 040002AA 0202005E
unknown ARM9 IO read16 040002B2 0202005E
unknown ARM9 IO read16 0402991C 0202005E
unknown ARM9 IO read16 04104672 0202005E
unknown ARM9 IO read16 0402922E 0202005E
unknown ARM9 IO read16 0402991A 0202005E
unknown ARM9 IO read16 04039A16 0202005E

StapleButter:
it seems to shit itself rather badly, it's accessing garbage addresses

benderscruffy:
still present in 0.7

benderscruffy:
still present in 0.7.2

StapleButter:
it's probably shitting itself just because DLDI isn't emulated

nathanielcwm:
Still stuck on a black screen with dldi enabled in the settings on fbca473

Arisotura:
you need to make sure the ROM file is present within your DLDI image.

poudink:
I tried placing the rom at the root of my virtual SD card. Still black screen.

iamthelaw2041:
This demo works now. Can this issue be closed?

![Capture](https://user-images.githubusercontent.com/94656803/144357334-5def1323-75df-46a9-b137-c8112431fda0.PNG)

PR #1583, opened 2022-12-13
GDB stub

PoroCYon:
The GDB stub is a feature to add debug capabilities (using an external debugger like GDB) to melonDS, to reverse-engineer games and system software, or to help with making homebrew NDS/DSi code.

There's an option in the emu settings menu to enable the GDB stub, configure which ports it uses, and to optionally wait until a debugger is attached before starting the system. (Don't enable this last option for both CPU cores at once!)

The GDB stub can be disabled by setting the `ENABLE_GDBSTUB` CMake option to `Off`, it is on by default.

What works:

* Register read/write
* Memory read/write
* Single-stepping
* Code breakpoints

What doesn't work:

* Loading a new binary through the debugger and executing it (you have to restart melonDS)
* Data watchpoints
* Sending over XML information about the memory map (cf. NWRAM mapping)

----------

Oriignal PR comment below:

> * ARM9 port is 3333
> * ARM7 port is 3334
>
> Marking this as WIP because I want more people to test it, to see if there's stuff they need that's not (properly) implemented yet, to see if they can find bugs, etc.
>
> Currently, it's again behind a CMake option (`ENABLE_GDBSTUB`), and no UI options for enabling the stubs and configuring their ports is present (which should be added before this is merged).


nathanfarlow:
This is amazing! Only tested out the ARM9 part. It works extremely well. With it, I was able to create a [DSi browser exploit](https://github.com/nathanfarlow/stylehax). Thank you for this!

One thing which I noticed is that reading memory is pretty slow, which makes searching in memory for things impossible. This is my first time interacting with ARM, but isn't the lsb of the pc register supposed to be 1 in thumb mode? The lsb was never set in thumb mode in my experience. Maybe I was supposed to look somewhere else to determine that?  Setting the registers with `set $r0 = 3` works, but only after a few seconds delay and a `Ignoring packet error, continuing...` message.

Again thanks for this addon. Couldn't have made the exploit without it.

RSDuck:
> This is my first time interacting with ARM, but isn't the lsb of the pc register supposed to be 1 in thumb mode?

no, now thinking about it, it's a bit confusing, but the bit is only mostly in return addresses to signal a mode switch. So it's put there by Thum blx in the link register and on ARMv5 all jumps will pick up this bit to switch to thumb mode. But it will always discard it afterwards. Whether the processor is currently in Thumb mode is stored though in CPSR bit 5.

PoroCYon:
> One thing which I noticed is that reading memory is pretty slow, which makes searching in memory for things impossible.

Hm, if GDB sends these read packets as 4 bytes at a time (instead of an entire memory chunk), it could be that the event polling queue thingy is only processing one of those messages at a time, which could be slowing down the thing. There's a number of `printf()` statements commented out in gdbstub.c and gdbproto.c, could you uncomment these and give the output when you do a memory search?

> This is my first time interacting with ARM, but isn't the lsb of the pc register supposed to be 1 in thumb mode?

> Whether the processor is currently in Thumb mode is stored though in CPSR bit 5.

While what RSDuck says is true, I wouldn't be 100% sure that re-adding the bit isn't needed. For example, due to CPU pipelining, `pc` is always an instruction or two ahead compared to what GDB expects, so I have to [correct for it](https://github.com/melonDS-emu/melonDS/blob/cd15080141b5378da9590428ff91adcff2df3d8d/src/ARM.cpp#L986), otherwise GDB gets things wrong. I do have a Cypress FX3 devboard here (with an ARM926EJ-S), which has a JTAG debugger so I'll check there whether it has to be set or not.

> Setting the registers with set $r0 = 3 works, but only after a few seconds delay and a Ignoring packet error, continuing... message.

Hm, that's slightly weird. Can you also say which error exactly is happening here as well, when you uncomment the `printf()` statements in the code?

PoroCYon:
image

Looks like `pc` always has an even value in GDB on the FX3, so I guess the current behavior is fine.

Though, I should probably support all the mode-banked registers as well, that's currently not implemented. Sadly there's no way to get a good register index↔name linking afaik, so time to hunt through either the OpenOCD or the GDB sources, yaay...

PoroCYon:
Also note to self: currently, when trying to connect to the GDB stub with the JIT enabled, GDB will time out, and melonDS sometimes dies with a SIGILL(?).

Debugging and JIT are mutually exclusive, as the latter makes it impossible to single-step or to put breakpoints anywhere, and I don't see that changing anytime soon (unless @RSDuck wants to rearchitect the JIT, which I doubt.) So for now it's probably best to disable the GDB stubs completely when the JIT is enabled, and maybe emit a warning when someone tries to connect, instead of it crashing and burning without giving users a clue at all.

SeleDreams:
I'm not too sure to understand how to use the gdb stub, like, are there CLI arguments to start it ? because i searched in the source it doesn't seem like it

PoroCYon:
No, currently it's always-on (which is also a thing that should change, ideally in the settings menu as well), as long as you pass the `ENABLE_GDBSTUB` cmake option.

SeleDreams:
> No, currently it's always-on (which is also a thing that should change, ideally in the settings menu as well), as long as you pass the `ENABLE_GDBSTUB` cmake option.

in this case what port is it using by default ? because i'd like to use it to debug homebrew i'm making but i'm not sure about how to go about it

SeleDreams:
I found the port by looking at the source. 3333.
I feel like one thing that would be cool is the ability to wait for the debugger to attach before launching the nds file. this way it would be easier to catch issues happening at boot time

PoroCYon:
It's... literally the first thing I wrote in the description here: https://github.com/melonDS-emu/melonDS/pull/1583#issue-1494849426

> I feel like one thing that would be cool is the ability to wait for the debugger to attach before launching the nds file. this way it would be easier to catch issues happening at boot time

Ah, alright, I'll put that in as an option when I get around to putting eg. the enable and port numbers in the option UI as well.

patataofcourse:
Something I just remembered- in my experience, reading from the GBA slot (whether ROM or save) always returned 0, any idea why that could be?

SeleDreams:
> It's... literally the first thing I wrote in the description here: [#1583 (comment)](https://github.com/melonDS-emu/melonDS/pull/1583#issue-1494849426)
> 
> > I feel like one thing that would be cool is the ability to wait for the debugger to attach before launching the nds file. this way it would be easier to catch issues happening at boot time
> 
> Ah, alright, I'll put that in as an option when I get around to putting eg. the enable and port numbers in the option UI as well.

don't mind me I spent the time working on some stuff and was half asleep while reading lol

PoroCYon:
> Something I just remembered- in my experience, reading from the GBA slot (whether ROM or save) always returned 0, any idea why that could be?

From GDB or from regular code running in the emulator? The former needs some code duplication for things that don't go through the regular system bus (eg. the ARM9 TCMs), so maybe the GBA ROM is like that too. If the latter, no clue, sorry.

axel7083:
Does anyone was able to connect this to ghidra debugger ?

PoroCYon:
Alright, so, updates! Finally!
I've also added support for listing all execution-mode-dependent CPU registers (with `sp_svc` and so on).

The GDB stub can now also be enabled (and the ports can be configured) in the emu settings window.

@nathanfarlow can you check whether you still get these errors? It looks like I forgot to send acks for the `P` (register set) command, that should be fixed now.

@patataofcourse can you also try the GBA cart thingy again? Because I'm not sure how that would happen (after I had fixed the TCMs not being mapped into memory issue, but that fix was done a long time ago already).

@axel7083 as long as it uses the GDB protocol it should work. Just try it. If it doesn't work, that's probably a bug on my end.

PoroCYon:
> I feel like one thing that would be cool is the ability to wait for the debugger to attach before launching the nds file. this way it would be easier to catch issues happening at boot time

This is implemented now

patataofcourse:
> @patataofcourse can you also try the GBA cart thingy again? Because I'm not sure how that would happen (after I had fixed the TCMs not being mapped into memory issue, but that fix was done a long time ago already).
works now! thank you

jonko0493:
Tested on Windows connecting through VS code and haven't had any issues -- seems to work brilliantly. Thanks for putting this together!

src/ARM.h
```diff
@@ -163,6 +183,20 @@ class ARM
     void (*BusWrite8)(u32 addr, u8 val);
     void (*BusWrite16)(u32 addr, u16 val);
     void (*BusWrite32)(u32 addr, u32 val);
+
+#ifdef GDBSTUB_ENABLED
+    bool is_single_step;
```
RSDuck, line 188: member variables in PascalCase

src/ARM.cpp
```diff
@@ -916,3 +993,220 @@ void ARMv4::FillPipeline()
...
+{
+    using Gdb::Register;
+    ARM* cpu = (ARM*)ud;
+    int r = static_cast<int>(reg);
+
+    if (reg < Register::pc) return cpu->R[r];
+    else if (reg == Register::pc)
+        return cpu->R[r] - ((cpu->CPSR & 0x20) ? 2 : 4);
+    else if (reg == Register::cpsr) return cpu->CPSR;
+    else if (reg == Register::sp_usr || reg == Register::lr_usr) {
```
RSDuck, line 1008: opening curly braces on the next line here and in the rest of the code

src/debug/GdbArch.h
```diff
@@ -0,0 +1,60 @@
...
+	spsr_fiq,
+	spsr_irq,
+	spsr_svc,
+	spsr_abt,
+	spsr_und,
+
+	COUNT
+};
+
+#define GDB_ARCH_N_REG ((int)Register::COUNT)
```
RSDuck, line 55: this would probably better be a constexpr constant

src/debug/GdbProto.cpp
```diff
@@ -0,0 +1,317 @@
...
+ * vAttach;addr
+ * vKill;pid
+ * qRcmd? qSupported?
+ */
+u8 Cmdbuf[GDBPROTO_BUFFER_CAPACITY];
+ssize_t Cmdlen;
+
+namespace Proto {
+
+u8 packetbuf[GDBPROTO_BUFFER_CAPACITY];
```
RSDuck, line 48: global variables in PascalCase

src/ARM.cpp
```diff
@@ -34,6 +35,43 @@
 using Platform::Log;
 using Platform::LogLevel;
 
+#ifdef GDBSTUB_ENABLED
+#define GDB_CHECK_A() do{\
```
RSDuck, line 39: could you make these a member function of ARM which just has an empty body when GDBSTUB_ENABLED is not defined?

src/debug/GdbCmds.cpp
```diff
@@ -0,0 +1,772 @@
...
+	"<feature name=\"org.gnu.gdb.arm.core\">"
+	TARGET_XML__CORE_REGS
+	TARGET_XML__MODE_REGS
+	// 39 regs total
+	"</feature>"
+	"</target>";
+	// TODO: CP15?
+
+
+int do_q_response(int connfd, const u8* query, const char* data, const size_t len) {
```
RSDuck, line 100: functions in PascalCase

src/debug/GdbProto.h
```diff
@@ -0,0 +1,42 @@
...
+#define GDBPROTO_H_
+
+#include <string.h>
+#include <sys/types.h>
+
+#include "GdbStub.h" /* class GdbStub */
+
+namespace Gdb {
+
+#define GDBPROTO_BUFFER_CAPACITY (1024+128)
```
RSDuck, line 12: could probably also be a constexpr

src/debug/GdbStub.h
```diff
@@ -0,0 +1,156 @@
...
+	BreakReq, // "break" command from gdb client
+	Bkpt,
+	Watchpt,
+	BkptInsn, // "bkpt" instruction
+	FaultData, // data abort
+	FaultIAcc, // instruction fetch abort
+	FaultInsn, // illegal instruction
+};
+
+struct StubCallbacks {
```
RSDuck, line 30: can you make this an interface instead (with pure virtual methods instead of function pointers)? Then ARM could just implement that.

RSDuck (commented):
I only marked one instance of some stylistic things

PoroCYon:
@kynex7510 does it work now?

kynex7510:
> @kynex7510 does it work now?

Yup, thanks!

GlowingUmbreon:
I have attempted to get this working with Ghidra with very little success.

Ghidra: 10.3.1
NDS Loader: https://github.com/pedro-javierf/NTRGhidra
GDB: gdb-multiarch 13.1

I connect ghidra to gdb-multiarch and run the following commands:
```
set arch armv5t
target remote 127.0.0.1
```
Then setting ghidra to record this melonDS Inferior.

Continuing and Breaking will work and registers will update but it appears that any attempt to access memory (i.e. scrolling in the memory window) will result in melonDS throwing the following warning: `[GDB] packet with len 1152 can't fit in buffer!`. and Ghidra will fail to read the memory. 
As an experiment I increased `GDBPROTO_BUFFER_CAPACITY` to `1152+128`. The error warning then changed to `[GDB] packet with len 1280 can't fit in buffer!`. So it may possible that Ghidra is sending a incorrect packet size?

Setting `GDBPROTO_BUFFER_CAPACITY` to `8192+128` appears to fix this specific error.

PoroCYon:
Thanks for the troubleshooting, I've pushed the fix. (Weird how GDB and Ghidra behave differently.)

GlowingUmbreon:
When the specified port is already in use there will be no warnings etc. but a bunch of buttons inside of the **Config** toolbar menu will be disabled.

PoroCYon:
That's by design because, when launching multiple melonDS instances, their configuration needs to be synchronized to make local multiplayer possible.

When launching melonDS while letting `nc -l 0.0.0.0 3333` run), a warning is still printed in the log output of melonDS (typically in the terminal/console). I guess you also want something else? Though I'm not sure what the best way of reporting such warnings would be. A popup messagebox? An info sign in the config menu? Neither feels satisfactory and would also be annoying to implement as it adds extra dependencies in the emulator core. (Plus, honestly, I'd expect people who would use the GDB stub to be able to read the emulator logs and figure it out. It's an "expert" feature, after all.)

GlowingUmbreon:
> That's by design because, when launching multiple melonDS instances, their configuration needs to be synchronized to make local multiplayer possible.
> 
> When launching melonDS while letting `nc -l 0.0.0.0 3333` run), a warning is still printed in the log output of melonDS (typically in the terminal/console). I guess you also want something else? Though I'm not sure what the best way of reporting such warnings would be. A popup messagebox? An info sign in the config menu? Neither feels satisfactory and would also be annoying to implement as it adds extra dependencies in the emulator core. (Plus, honestly, I'd expect people who would use the GDB stub to be able to read the emulator logs and figure it out. It's an "expert" feature, after all.)

Your reasoning does make sense, it does feel strange to have a GUI to enable it but then all related information gets dumped into the command line. What I was thinking could work was a status bar in the bottom right (GDB connected to 127.0.0.1, GDB running on port 3333, GDB port in use etc.) like what other emulators do but if it is hard to implement than it is probably fine to just print to command line.

PoroCYon:
A status bar could be a solution, another option could be to have a toolbar menu thingy to open a separate window with the emulator logs (mGBA does this). Though I'm not going to add either to this PR.

GlowingUmbreon:
Yeah this pull request has the base down at least, the rest can be added later. 
Also I have been trying out ghidra with this PR and it works quite well but it appears to have an issue where every time you attempt to pause the program it has a small (10% maybe) chance to get disconnected. I was wondering if there was any diagnostic steps you can think of that I can try so we see if we can find what causes this?

PoroCYon:
There's a number of `Log()` calls in the GDBstub source that are commented out. You can try uncommenting them and see if that gives you enough info to spot the difference.

axel7083:
> I have attempted to get this working with Ghidra with a little success.
> 
> Ghidra: 10.3.1 NDS Loader: https://github.com/pedro-javierf/NTRGhidra GDB: gdb-multiarch 13.1
> 
> I connect ghidra to gdb-multiarch and run the following commands:
> 
> ```
> set arch armv5t
> target remote 127.0.0.1
> ```

I wanna try to replicates, but I cannot get gdb-multiarch to connect to the MelonDS instance.

On windows, using MSYS2 instructions, I created a static build, launch MelonDS, open a game, it runs fine, but whe trying `gdb-multiarch` on either the MSYS2 console, or WSL2-Ubuntu , I always get `Connection timed out.` 
I tried 
- `target remote 0.0.0.0:3333`
- `target remote 127.0.0.1:3333`
- `target remote localhost:3333`

Any ideas ?

PoroCYon:
Did you check the GDB settings in melonDS itself? When running melonDS in a terminal, what's the output you are getting?

turtleisaac:
I built the gdbstub-standalone branch and tried to connect with my debugger, but I get `failed to get reply to handshake packet within timeout of 0.0 seconds` immediately in the terminal output and it doesn't connect. I know an attempt at connecting was clearly made because game execution starts in melonDS, but clearly something is going wrong that prevents it from working.

PoroCYon:
@turtleisaac well, if it's related to the previous comment, could you try doing what I suggested for that post as well?

> Did you check the GDB settings in melonDS itself? When running melonDS in a terminal, what's the output you are getting?
>
> Also, if you compile it yourself, make sure the JIT is disabled.

turtleisaac:
> @turtleisaac well, if it's related to the previous comment, could you try doing what I suggested for that post as well?
> 
> > Did you check the GDB settings in melonDS itself? When running melonDS in a terminal, what's the output you are getting?
> > Also, if you compile it yourself, make sure the JIT is disabled.

definitely not one of my finest GitHub moments, not trying what you recommended the other guy lol...

Anyways, rebuilt with the argument `-DENABLE_JIT=False` and launched from Terminal. Noticed the following in melonDS's output:

```
[GDB] initializing GDB stub for core 9 on port 3333
[GDB] err: can't bind to address <any> and port 3333
[GDB] initializing GDB stub for core 7 on port 3334
```

Tried turning GDB off, closing and reopening, then turning it on, closing and reopening. That solved that. Continuing onwards though and now after choosing a game and having it wait to boot, I see this coming from melonDS upon attempting a connection :

```
Game is now booting
[GDB] unknown command 'Q'!
[GDB] disconnect
[GDB] enter exit: 5
````

**Comment by PoroCYon** [comment]
Ah, I see what's going wrong now. It looks like your GDB is sending a command that my implementation can't parse yet. What GDB version are you using?

Also, since `Q` can be [a number of different query commands](https://sourceware.org/gdb/onlinedocs/gdb/General-Query-Packets.html#General-Query-Packets), I'll need a bit more info: could you please uncomment [this](https://github.com/melonDS-emu/melonDS/blob/30bb773f39b73d34f62eb709c413f7a3b33e7f80/src/debug/GdbProto.cpp#L116) and [this](https://github.com/melonDS-emu/melonDS/blob/30bb773f39b73d34f62eb709c413f7a3b33e7f80/src/debug/GdbStub.cpp#L397) line and send me the output again?

**Comment by axel7083** [comment]
Using `pacman -S mingw-w64-clang-x86_64-gdb-multiarch`[^1], I was able to connect to the debugger once the game is started. Working good !

> I connect ghidra to gdb-multiarch and run the following commands:

However on windows I am having a lot of trouble with connecting ghidra to gdb-multiarch...



[^1]: [mingw-w64-clang-x86_64-gdb-multiarch](https://packages.msys2.org/package/mingw-w64-clang-x86_64-gdb-multiarch?repo=clang64)

**Comment by PoroCYon** [comment]
@axel7083 it looks like there's multiple things going on. The checksum error isn't something that should happen, though I'll need a little bit more info to know what exactly is going wrong again. Can you uncomment [these two lines](https://github.com/melonDS-emu/melonDS/blob/30bb773f39b73d34f62eb709c413f7a3b33e7f80/src/debug/GdbProto.cpp#L145-L171) (at the begin and end of the selected region) too and run it once more?

Though, there's also this part:
```
[GDB] command in: 'qSupported:multiprocess+;swbreak+;hwbreak+;qRelocInsn+;fork-events+;vfork-events+;exec-events+;vContS
upported+;QThreadEvents+;no-resumed+;memory-tagging+;xmlRegisters=i386'
```
It looks like your GDB is running for x86 targets while it should be ARM? You probably also want to add `-iex 'set architecture arm'` to the arguments specified in Ghidra.

**Comment by axel7083** [comment]
> target remote

okey thanks for your answer. I changed the argument to `C:\\msys64\\clang64\\bin\\gdb-multiarch -i mi2 -iex 'set architecture arm'`.

Here are the logs of gfb-mutliarch
```
(gdb)set arch armv5t
warning: A handler for the OS ABI "Windows" is not built into this configuration
of GDB.  Attempting to continue with the default armv5t settings.

The target architecture is set to "armv5t".
(gdb)target remote 
To open a remote debug connection, you need to specify what
serial device is attached to the remote system
(e.g. /dev/ttyS0, /dev/ttya, COM1, etc.).
(gdb)target remote 127.0.0.1:3333
Remote debugging using 127.0.0.1:3333
Ignoring packet error, continuing...
Remote replied unexpectedly to 'vMustReplyEmpty': timeout
```

**MelonDS stdout**
```
...
Game is now booting
[GDB] recv() 701 bytes: '+$qSupported:multiprocess+;swbreak+;hwbreak+;qRelocInsn+;fork-events+;vfork-events+;exec-events
+;vContSupported+;QThreadEvents+;no-resumed+;memory-tagging+;xmlRegisters=i386#77$qSupported:multiprocess+;swbreak+;hwbr
eak+;qRelocInsn+;fork-events+;vfork-events+;exec-events+;vContSupported+;QThreadEvents+;no-resumed+;memory-tagging+;xmlR
egisters=i386#77$qSupported:multiprocess+;swbreak+;hwbreak+;qRelocInsn+;fork-events+;vfork-events+;exec-events+;vContSup
ported+;QThreadEvents+;no-resumed+;memory-tagging+;xmlRegisters=i386#77$qSupported:multiprocess+;swbreak+;hwbreak+;qRelo
cInsn+;fork-events+;vfork-events+;exec-events+;vContSupported+;QThreadEvents+;no-resumed+;memory-tagging+;xmlRegisters=i
386#77' (2b)
[GDB] recv() after skipping: n=700, recv_total=700
[GDB] got pkt, checksum: 77 vs 77
[GDB] command in: 'qSupported:multiprocess+;swbreak+;hwbreak+;qRelocInsn+;fork-events+;vfork-events+;exec-events+;vContS
upported+;QThreadEvents+;no-resumed+;memory-tagging+;xmlRegisters=i386'
[GDB] got pkt, checksum: 77 vs 77
[GDB] BIG MISTAKE: 700 > 525 which shouldn't happen!
[GDB] WUT
[GDB] EOF!
[GDB] disconnect
[GDB] enter exit: 5
```

**Comment by PoroCYon** [comment]
Ok looks like there's something wrong with the command parsing/sequencing code. Working on a fix.

**Comment by axel7083** [comment]
> target remote 127.0.0.1


MelonDS stdout

```
...
Game is now booting
--- dataoff=0
[GDB] recv() 701 bytes: '+$qSupported:multiprocess+;swbreak+;hwbreak+;qRelocInsn+;fork-events+;vfork-events+;exec-events
+;vContSupported+;QThreadEvents+;no-resumed+;memory-tagging+;xmlRegisters=i386#77$qSupported:multiprocess+;swbreak+;hwbr
eak+;qRelocInsn+;fork-events+;vfork-events+;exec-events+;vContSupported+;QThreadEvents+;no-resumed+;memory-tagging+;xmlR
egisters=i386#77$qSupported:multiprocess+;swbreak+;hwbreak+;qRelocInsn+;fork-events+;vfork-events+;exec-events+;vContSup
ported+;QThreadEvents+;no-resumed+;memory-tagging+;xmlRegisters=i386#77$qSupported:multiprocess+;swbreak+;hwbreak+;qRelo
cInsn+;fork-events+;vfork-events+;exec-events+;vContSupported+;QThreadEvents+;no-resumed+;memory-tagging+;xmlRegisters=i
386#77' (2b)
[GDB] recv() after skipping: n=700, recv_total=700
[GDB] got pkt, checksum: 77 vs 77
[GDB] got more: cksumoff=173, recvtotal=700, remain=525
==> $qSupported:multiprocess+;swbreak+;hwbreak+;qRelocInsn+;fork-events+;vfork-events+;exec-events+;vContSupported+;QThr
eadEvents+;no-resumed+;memory-tagging+;xmlRegisters=i386#77$qSupported:multiprocess+;swbreak+;hwbreak+;qRelocInsn+;fork-
events+;vfork-events+;exec-events+;vContSupported+;QThreadEvents+;no-resumed+;memory-tagging+;xmlRegisters=i386#77$qSupp
orted:multiprocess+;swbreak+;hwbreak+;qRelocInsn+;fork-events+;vfork-events+;exec-events+;vContSupported+;QThreadEvents+
;no-resumed+;memory-tagging+;xmlRegisters=i386#77
[GDB] command in: 'qSupported:multiprocess+;swbreak+;hwbreak+;qRelocInsn+;fork-events+;vfork-events+;exec-events+;vContS
upported+;QThreadEvents+;no-resumed+;memory-tagging+;xmlRegisters=i386'
[GDB] send ack
[GDB] send ack
[GDB] send resp: '$PacketSize=47F;qXfer:features:read+;swbreak-;hwbreak+#5d'
--- dataoff=525
--- got preexisting: $qSupported:multiprocess+;swbreak+;hwbreak+;qRelocInsn+;fork-events+;vfork-events+;exec-events+;vCo
ntSupported+;QThreadEvents+;no-resumed+;memory-tagging+;xmlRegisters=i386#77$qSupported:multiprocess+;swbreak+;hwbreak+;
qRelocInsn+;fork-events+;vfork-events+;exec-events+;vContSupported+;QThreadEvents+;no-resumed+;memory-tagging+;xmlRegist
ers=i386#77$qSupported:multiprocess+;swbreak+;hwbreak+;qRelocInsn+;fork-events+;vfork-events+;exec-events+;vContSupporte
d+;QThreadEvents+;no-resumed+;memory-tagging+;xmlRegisters=i386#77
--- datastart=1
--- cksumoff=173
=== cksumoff=173 recv_total=525 datastart=1 dataoff=175
==> $qSupported:multiprocess+;swbreak+;hwbreak+;qRelocInsn+;fork-events+;vfork-events+;exec-events+;vContSupported+;QThr
eadEvents+;no-resumed+;memory-tagging+;xmlRegisters=i386#77$qSupported:multiprocess+;swbreak+;hwbreak+;qRelocInsn+;fork-
events+;vfork-events+;exec-events+;vContSupported+;QThreadEvents+;no-resumed+;memory-tagging+;xmlRegisters=i386#77$qSupp
orted:multiprocess+;swbreak+;hwbreak+;qRelocInsn+;fork-events+;vfork-events+;exec-events+;vContSupported+;QThreadEvents+
;no-resumed+;memory-tagging+;xmlRegisters=i386#77
[GDB] got pkt, checksum: 77 vs 77
[GDB] got more: cksumoff=173, recvtotal=525, remain=350
==> $qSupported:multiprocess+;swbreak+;hwbreak+;qRelocInsn+;fork-events+;vfork-events+;exec-events+;vContSupported+;QThr
eadEvents+;no-resumed+;memory-tagging+;xmlRegisters=i386#77$qSupported:multiprocess+;swbreak+;hwbreak+;qRelocInsn+;fork-
events+;vfork-events+;exec-events+;vContSupported+;QThreadEvents+;no-resumed+;memory-tagging+;xmlRegisters=i386#77
[GDB] command in: 'qSupported:multiprocess+;swbreak+;hwbreak+;qRelocInsn+;fork-events+;vfork-events+;exec-events+;vContS
upported+;QThreadEvents+;no-resumed+;memory-tagging+;xmlRegisters=i386'
[GDB] send ack
[GDB] send ack
[GDB] send resp: '$PacketSize=47F;qXfer:features:read+;swbreak-;hwbreak+#5d'
[GDB] send resp: '$PacketSize=47F;qXfer:features:read+;swbreak-;hwbreak+#5d'
[GDB] send resp: '$PacketSize=47F;qXfer:features:read+;swbreak-;hwbreak+#5d'
--- dataoff=350
--- got preexisting: $qSupported:multiprocess+;swbreak+;hwbreak+;qRelocInsn+;fork-events+;vfork-events+;exec-events+;vCo
ntSupported+;QThreadEvents+;no-resumed+;memory-tagging+;xmlRegisters=i386#77$qSupported:multiprocess+;swbreak+;hwbreak+;
qRelocInsn+;fork-events+;vfork-events+;exec-events+;vContSupported+;QThreadEvents+;no-resumed+;memory-tagging+;xmlRegist
ers=i386#77
--- datastart=1
--- cksumoff=173
=== cksumoff=173 recv_total=350 datastart=1 dataoff=175
==> $qSupported:multiprocess+;swbreak+;hwbreak+;qRelocInsn+;fork-events+;vfork-events+;exec-events+;vContSupported+;QThr
eadEvents+;no-resumed+;memory-tagging+;xmlRegisters=i386#77$qSupported:multiprocess+;swbreak+;hwbreak+;qRelocInsn+;fork-
events+;vfork-events+;exec-events+;vContSupported+;QThreadEvents+;no-resumed+;memory-tagging+;xmlRegisters=i386#77
[GDB] got pkt, checksum: 77 vs 77
[GDB] got more: cksumoff=173, recvtotal=350, remain=175
==> $qSupported:multiprocess+;swbreak+;hwbreak+;qRelocInsn+;fork-events+;vfork-events+;exec-events+;vContSupported+;QThr
eadEvents+;no-resumed+;memory-tagging+;xmlRegisters=i386#77
[GDB] command in: 'qSupported:multiprocess+;swbreak+;hwbreak+;qRelocInsn+;fork-events+;vfork-events+;exec-events+;vContS
upported+;QThreadEvents+;no-resumed+;memory-tagging+;xmlRegisters=i386'
[GDB] send ack
[GDB] send packet ack failed!
[GDB] disconnect
[GDB] enter exit: 5
remapping DTCM ffffffff ffffffff 3000000 2ffc000
PU: region 0 = 04000033 : enabled, 04000000-08000000
PU: region 1 = 0200002B : enabled, 02000000-02400000
PU: region 2 = 027C0023 : enabled, 027C0000-02800000
PU: region 3 = 08000035 : enabled, 08000000-10000000
PU: region 4 = 023C001B : enabled, 023C0000-023C4000
PU: region 5 = 0100002F : enabled, 01000000-02000000
PU: region 6 = FFFF001D : enabled, FFFF0000-FFFF8000
PU: region 7 = 027FF017 : enabled, 027FF000-02800000
remapping DTCM 23c0000 23c4000 ffffffff ffffffff
PU region 0: 04000000-08000000, user=00 priv=07, 15111011/05100011
PU region 1: 02000000-02400000, user=70 priv=77, 15111011/05100011
PU region 2: 027C0000-02800000, user=00 priv=00, 15111011/05100011
PU region 3: 08000000-10000000, user=00 priv=03, 15111011/05100011
PU region 4: 023C0000-023C4000, user=00 priv=03, 15111011/05100011
PU region 5: 01000000-02000000, user=00 priv=07, 15111011/05100011
PU region 6: FFFF0000-FFFF8000, user=50 priv=55, 15111011/05100011
PU region 7: 027FF000-02800000, user=00 priv=03, 15111011/05100011
[GDB] send resp: '$W00#b7'
--- dataoff=175
--- got preexisting: $qSupported:multiprocess+;swbreak+;hwbreak+;qRelocInsn+;fork-events+;vfork-events+;exec-events+;vCo
ntSupported+;QThreadEvents+;no-resumed+;memory-tagging+;xmlRegisters=i386#77
--- datastart=1
--- cksumoff=173
=== cksumoff=173 recv_total=175 datastart=1 dataoff=175
==> $qSupported:multiprocess+;swbreak+;hwbreak+;qRelocInsn+;fork-events+;vfork-events+;exec-events+;vContSupported+;QThr
eadEvents+;no-resumed+;memory-tagging+;xmlRegisters=i386#77
[GDB] got pkt, checksum: 77 vs 77
[GDB] command in: 'qSupported:multiprocess+;swbreak+;hwbreak+;qRelocInsn+;fork-events+;vfork-events+;exec-events+;vContS
upported+;QThreadEvents+;no-resumed+;memory-tagging+;xmlRegisters=i386'
[GDB] send ack
[GDB] send packet ack failed!
[GDB] disconnect
[GDB] enter exit: 5
[GDB] send resp: '$W00#b7'
--- dataoff=0
--- dataoff=0
--- dataoff=0
WIFI: ON
--- dataoff=0
--- dataoff=0
--- dataoff=0
--- dataoff=0
WIFI: OFF
unknown ARM9 IO write32 04001060 00000000 01FFDE38
unknown ARM9 IO write32 04001064 00000000 01FFDE38
unknown ARM9 IO write32 04001068 00000000 01FFDE38
--- dataoff=0
--- dataoff=0
--- dataoff=0
--- dataoff=0
--- dataoff=0
--- dataoff=0
--- dataoff=0
--- dataoff=0
```


gdb-multiarch in ghidra stdout
```
(gdb)set arch armv5t
warning: A handler for the OS ABI "Windows" is not built into this configuration
of GDB.  Attempting to continue with the default armv5t settings.

The target architecture is set to "armv5t".
(gdb)target remote 127.0.0.1:3333
Remote debugging using 127.0.0.1:3333
Remote replied unexpectedly to 'vMustReplyEmpty': PacketSize=47F;qXfer:features:read+;swbreak-;hwbreak+
```

**Comment by PoroCYon** [comment]
Looks like with my fixes, your GDB doesn't seem to be following the GDB protocol properly. It sends four `qSupported` packets and one `vMustReplyEmpty` packet, so my server implementation tries to send four `qSupported` responses, one for each packet. When trying to send the response for the last one, things fail. This seems to be the case because your GDB expects the second response sent by my server implementation to be the response for `vMustReplyEmpty` (i.e. skipping the other three `qSupported` packets somehow). Then your GDB closes the connection, and my implementation subsequently fails with `[GDB] send packet ack failed!`.

That's a lot of words to say, I'm not sure what's wrong, please try a different (version of) GDB.

**Comment by turtleisaac** [comment]
> Ah, I see what's going wrong now. It looks like your GDB is sending a command that my implementation can't parse yet. What GDB version are you using?
> 
> Also, since `Q` can be [a number of different query commands](https://sourceware.org/gdb/onlinedocs/gdb/General-Query-Packets.html#General-Query-Packets), I'll need a bit more info: could you please uncomment [this](https://github.com/melonDS-emu/melonDS/blob/30bb773f39b73d34f62eb709c413f7a3b33e7f80/src/debug/GdbProto.cpp#L116) and [this](https://github.com/melonDS-emu/melonDS/blob/30bb773f39b73d34f62eb709c413f7a3b33e7f80/src/debug/GdbStub.cpp#L397) line and send me the output again?

Here you go:

```
Game is now booting
--- dataoff=0
[GDB] recv() 1 bytes: '+' (2b)
[GDB] recv() 19 bytes: '$QStartNoAckMode#b0' (24)
[GDB] recv() after skipping: n=19, recv_total=19
[GDB] got pkt, checksum: b0 vs b0
[GDB] command in: 'QStartNoAckMode'
[GDB] unknown command 'Q'!
[GDB] send resp: '$#00'
[GDB] send resp: '$#00'
[GDB] send resp: '$#00'
[GDB] disconnect
[GDB] enter exit: 5
```

**Comment by PoroCYon** [comment]
Heh, looks like LLDB is also not *quite* standard-compliant in how it works (it's only supposed to send `QStartNoAckMode` after seeing that a response to `qSupported` advertises that the server does in fact support this, which my implementation currently doesn't). So I've opened an issue [on their side](https://github.com/llvm/llvm-project/issues/64872). Though, in the meantime, I'll see if I can implement no-ack mode anyway.

**Comment by turtleisaac** [comment]
> Heh, looks like LLDB is also not _quite_ standard-compliant in how it works (it's only supposed to send `QStartNoAckMode` after seeing that a response to `qSupported` advertises that the server does in fact support this, which my implementation currently doesn't). So I've opened an issue [on their side](https://github.com/llvm/llvm-project/issues/64872). Though, in the meantime, I'll see if I can implement no-ack mode anyway.
> 
> EDIT: @turtleisaac pushed the 'fix', does it work now?

Pulled, rebuilt and gave it a shot. Unfortunately, melonDS crashes after I attempt the connection.

```
Game is now booting
--- dataoff=0
[GDB] recv() 1 bytes: '+' (2b)
[GDB] recv() 19 bytes: '$QStartNoAckMode#b0' (24)
[GDB] recv() after skipping: n=19, recv_total=19
[GDB] got pkt, checksum: b0 vs b0
[GDB] command in: 'QStartNoAckMode'
[GDB] send ack
[GDB] send ack
[GDB] send resp: '$OK#9a'
[GDB] send resp: '$OK#9a'
[GDB] send resp: '$OK#9a'
--- dataoff=0
[GDB] recv() 111 bytes: 'upported:xmlRegisters=i386,arm,mips,arc;multiprocess+#04+$QThreadSuffixSupported#e4+$QListThreadsInStopReply#21' (75)
zsh: bus error /Users/USER/Documents/melonDS/build/melonDS.app/Contents/MacOS/melonDS
```

Here's the Apple error report if you need it.

```
Process:               melonDS [26020]
Path:                  /Users/USER/Documents/*/melonDS.app/Contents/MacOS/melonDS
Identifier:            net.kuribo64.melonDS
Version:               0.9.5 (0.9.5)
Code Type:             ARM-64 (Native)
Parent Process:        zsh [7392]
Responsible:           Terminal [95029]
User ID:               502

Date/Time:             2023-08-21 14:44:24.3935 -0500
OS Version:            macOS 13.4 (22F66)
Report Version:        12
Anonymous UUID:        41FCB3B0-D5AA-8E08-C9E2-A14766F8C4F2

Sleep/Wake UUID:       F51B6370-E05E-491E-A6AD-5E0A79C3911F

Time Awake Since Boot: 250000 seconds
Time Since Wake:       10168 seconds

System Integrity Protection: enabled

Crashed Thread:        12  EmuThread

Exception Type:        EXC_BAD_ACCESS (SIGBUS)
Exception Codes:       KERN_PROTECTION_FAILURE at 0x0000000103183fe0
Exception Codes:       0x0000000000000002, 0x0000000103183fe0

Termination Reason:    Namespace SIGNAL, Code 10 Bus error: 10
Terminating Process:   exc handler [26020]

VM Region Info: 0x103183fe0 is in 0x103168000-0x103184000;  bytes after start: 114656  bytes before end: 31
      REGION TYPE                    START - END         [ VSIZE] PRT/MAX SHRMOD  REGION DETAIL
      __TEXT                      102f44000-103168000    [ 2192K] r-x/r-x SM=COW  ...MacOS/melonDS
--->  __DATA_CONST                103168000-103184000    [  112K] r--/rw- SM=COW  ...MacOS/melonDS
      __DATA                      103184000-103190000    [   48K] rw-/rw- SM=COW  ...MacOS/melonDS

...

Thread 12 Crashed:: EmuThread
0   libsystem_platform.dylib      	       0x19a096858 _platform_memmove + 536
1   melonDS                       	       0x10304b230 Gdb::Proto::MsgRecv(int, unsigned char*) + 456
2   melonDS                       	       0x10304a1c4 Gdb::GdbStub::Poll(bool) + 260
3   melonDS                       	       0x10304a5a0 Gdb::GdbStub::Enter(bool, Gdb::TgtStatus, unsigned int, bool) + 132
4   melonDS                       	       0x103024094 unsigned int NDS::RunFrame<false, 0>() + 248
5   melonDS                       	       0x102f53f9c EmuThread::run() + 2740
6   QtCore                        	       0x105568b40 QThreadPrivate::start(void*) + 332
7   libsystem_pthread.dylib       	       0x19a067fa8 _pthread_start + 148
8   libsystem_pthread.dylib       	       0x19a062da0 thread_start + 8
```

**Comment by PoroCYon** [comment]
What happens if you swap [these two lines](https://github.com/melonDS-emu/melonDS/blob/73d2c5f438e7296d598b86318603b32673f78675/src/debug/GdbCmds.cpp#L880-L881)?

**Comment by PoroCYon** [comment]
@turtleisaac Ok, I've implemented a few more commands which should be able to keep lldb happy. Can you try now?

**Comment by turtleisaac** [comment]
> @turtleisaac Ok, I've implemented a few more commands which should be able to keep lldb happy. Can you try now?

Still no luck unfortunately. Definitely seems to be a different series of errors tho?

```
Game is now booting
[GDB] inital handshake: didn't receive inital ack!
[GDB] disconnect
[GDB] enter exit: 5
PU: region 0 = 04000033 : enabled, 04000000-08000000
PU: region 1 = 0200002D : enabled, 02000000-02800000
PU: region 2 = 027E0021 : enabled, 027E0000-02800000
PU: region 3 = 08000035 : enabled, 08000000-10000000
PU: region 4 = 027E001B : enabled, 027E0000-027E4000
PU: region 5 = 0100002F : enabled, 01000000-02000000
PU: region 6 = FFFF001D : enabled, FFFF0000-FFFF8000
PU: region 7 = 027FF017 : enabled, 027FF000-02800000
PU region 0: 04000000-08000000, user=00 priv=07, 15111011/05100011
PU region 1: 02000000-02800000, user=70 priv=77, 15111011/05100011
PU region 2: 027E0000-02800000, user=00 priv=00, 15111011/05100011
PU region 3: 08000000-10000000, user=00 priv=03, 15111011/05100011
PU region 4: 027E0000-027E4000, user=00 priv=03, 15111011/05100011
PU region 5: 01000000-02000000, user=00 priv=07, 15111011/05100011
PU region 6: FFFF0000-FFFF8000, user=50 priv=55, 15111011/05100011
PU region 7: 027FF000-02800000, user=00 priv=03, 15111011/05100011
PU: region 1 = 0200002B : enabled, 02000000-02400000
PU region 0: 04000000-08000000, user=00 priv=07, 15111011/05100011
PU region 1: 02000000-02400000, user=70 priv=77, 15111011/05100011
PU region 2: 027E0000-02800000, user=00 priv=00, 15111011/05100011
PU region 3: 08000000-10000000, user=00 priv=03, 15111011/05100011
PU region 4: 027E0000-027E4000, user=00 priv=03, 15111011/05100011
PU region 5: 01000000-02000000, user=00 priv=07, 15111011/05100011
PU region 6: FFFF0000-FFFF8000, user=50 priv=55, 15111011/05100011
PU region 7: 027FF000-02800000, user=00 priv=03, 15111011/05100011
PU: region 2 = 023E0021 : enabled, 023E0000-02400000
PU region 0: 04000000-08000000, user=00 priv=07, 15111011/05100011
PU region 1: 02000000-02400000, user=70 priv=77, 15111011/05100011
PU region 2: 023E0000-02400000, user=00 priv=00, 15111011/05100011
PU region 3: 08000000-10000000, user=00 priv=03, 15111011/05100011
PU region 4: 027E0000-027E4000, user=00 priv=03, 15111011/05100011
PU region 5: 01000000-02000000, user=00 priv=07, 15111011/05100011
PU region 6: FFFF0000-FFFF8000, user=50 priv=55, 15111011/05100011
PU region 7: 027FF000-02800000, user=00 priv=03, 15111011/05100011
PU region 3: 08000000-10000000, user=00 priv=01, 15115011/05100011
unknown ARM9 IO write32 04001060 00000000 01FF8548
unknown ARM9 IO write32 04001064 00000000 01FF8548
unknown ARM9 IO write32 04001068 00000000 01FF8548
[GDB] send ack
[GDB] recv() 19 bytes: '$QStartNoAckMode#b0' (24)
[GDB] recv() after skipping: n=19, recv_total=19
[GDB] got pkt, checksum: b0 vs b0
[GDB] command in: 'QStartNoAckMode'
[GDB] send ack
[GDB] send ack
[GDB] send resp: '$OK#9a'
[GDB] recv() 59 bytes: '$qSupported:xmlRegisters=i386,arm,mips,arc;multiprocess+#04' (24)
[GDB] recv() after skipping: n=59, recv_total=59
[GDB] got pkt, checksum: 04 vs 04
[GDB] command in: 'qSupported:xmlRegisters=i386,arm,mips,arc;multiprocess+'
[GDB] send resp: '$PacketSize=47F;qXfer:features:read+;swbreak-;hwbreak+;QStartNoAckMode+#73'
[GDB] recv() 27 bytes: '+$QThreadSuffixSupported#e4' (2b)
[GDB] recv() after skipping: n=26, recv_total=26
[GDB] got pkt, checksum: e4 vs e4
[GDB] command in: 'QThreadSuffixSupported'
[GDB] unknown subcommand 'ThreadSuffixSupported'!
[GDB] send resp: '$#00'
[GDB] recv() 28 bytes: '+$QListThreadsInStopReply#21' (2b)
[GDB] recv() after skipping: n=27, recv_total=27
[GDB] got pkt, checksum: 21 vs 21
[GDB] command in: 'QListThreadsInStopReply'
[GDB] unknown subcommand 'ListThreadsInStopReply'!
[GDB] send resp: '$#00'
[GDB] recv() 14 bytes: '+$qHostInfo#9b' (2b)
[GDB] recv() after skipping: n=13, recv_total=13
[GDB] got pkt, checksum: 9b vs 9b
[GDB] command in: 'qHostInfo'
[GDB] send resp: '$cputype:12;cpusubtype:7;triple:arm-none-eabi;ostype:none;vendor:none;endian:little;ptrsize:4;#6a'
[GDB] recv() 39 bytes: '+$vCont?#49+$qVAttachOrWaitSupported#38' (2b)
[GDB] recv() after skipping: n=38, recv_total=38
[GDB] got pkt, checksum: 49 vs 49
[GDB] got more: cksumoff=8, recvtotal=38, remain=28
==> +$qVAttachOrWaitSupported#38
[GDB] command in: 'vCont?'
[GDB] send resp: '$vCont;c;s;t#05'
--- got preexisting: +$qVAttachOrWaitSupported#38
--- datastart=2
--- cksumoff=26
=== cksumoff=25 recv_total=27 datastart=2 dataoff=27
==> $qVAttachOrWaitSupported#38
[GDB] got pkt, checksum: 38 vs 38
[GDB] command in: 'qVAttachOrWaitSupported'
[GDB] unknown subcommand 'VAttachOrWaitSupported'!
[GDB] send resp: '$#00'
[GDB] recv() 24 bytes: '+$QEnableErrorStrings#8c' (2b)
[GDB] recv() after skipping: n=23, recv_total=23
[GDB] got pkt, checksum: 8c vs 8c
[GDB] command in: 'QEnableErrorStrings'
[GDB] unknown subcommand 'EnableErrorStrings'!
[GDB] send resp: '$#00'
[GDB] recv() 24 bytes: '+$qProcessInfo#dc+$qC#b4' (2b)
[GDB] recv() after skipping: n=23, recv_total=23
[GDB] got pkt, checksum: dc vs dc
[GDB] got more: cksumoff=14, recvtotal=23, remain=7
==> +$qC#b4
[GDB] command in: 'qProcessInfo'
[GDB] send resp: '$cputype:12;cpusubtype:7;triple:arm-none-eabi;ostype:none;vendor:none;endian:little;ptrsize:4;#6a'
--- got preexisting: +$qC#b4
--- datastart=2
--- cksumoff=5
=== cksumoff=4 recv_total=6 datastart=2 dataoff=6
==> $qC#b4
[GDB] got pkt, checksum: b4 vs b4
[GDB] command in: 'qC'
[GDB] send resp: '$QC1#c5'
[GDB] recv() 17 bytes: '+$qfThreadInfo#bb' (2b)
[GDB] recv() after skipping: n=16, recv_total=16
[GDB] got pkt, checksum: bb vs bb
[GDB] command in: 'qfThreadInfo'
[GDB] send resp: '$m1#9e'
[GDB] recv() 6 bytes: '+$?#3f' (2b)
[GDB] recv() after skipping: n=5, recv_total=5
[GDB] got pkt, checksum: 3f vs 3f
[GDB] command in: '?'
[GDB] send resp: '$S02#b5'
[GDB] recv() 1 bytes: '+' (2b)
[GDB] recv() error -1, errno=35 (Resource temporarily unavailable)
[GDB] EOF!
[GDB] disconnect
[GDB] enter exit: 5```

PoroCYon:
Can you try running the LLDB command `log enable gdb-remote all` before running the `gdb-remote` command, and give me the output? Because the melonDS log sounds like LLDB closed the connection on its side.

Also, the "inital handshake: didn't receive inital ack!" at the beginning sounds cursed (and also contradicting what's being said [here](https://github.com/llvm/llvm-project/issues/64872#issuecomment-1687004115), hooray). Can you also try the same when commenting away [this code](https://github.com/melonDS-emu/melonDS/blob/7305cb3496cfeda902d0766c10f2a1b3ae5adcdf/src/debug/GdbStub.cpp#L305-L318), and then with also [this code](https://github.com/melonDS-emu/melonDS/blob/7305cb3496cfeda902d0766c10f2a1b3ae5adcdf/src/debug/GdbProto.cpp#L73-L74) uncommented?

turtleisaac:
> Can you try running the LLDB command `log enable gdb-remote all` before running the `gdb-remote` command, and give me the output? Because the melonDS log sounds like LLDB closed the connection on its side.
> 
> Also, the "inital handshake: didn't receive inital ack!" at the beginning sounds cursed (and also contradicting what's being said [here](https://github.com/llvm/llvm-project/issues/64872#issuecomment-1687004115), hooray). Can you also try the same when commenting away [this code](https://github.com/melonDS-emu/melonDS/blob/7305cb3496cfeda902d0766c10f2a1b3ae5adcdf/src/debug/GdbStub.cpp#L305-L318), and then with also [this code](https://github.com/melonDS-emu/melonDS/blob/7305cb3496cfeda902d0766c10f2a1b3ae5adcdf/src/debug/GdbProto.cpp#L73-L74) uncommented?

Will do in a sec, but yeah I think it is detaching purposefully. Look at what I sent earlier:

> On the CLion side of things:
> 
> ```
> Process 1 was reported after connecting to 'connect://localhost:3333', but no stop reply packet was received
> Debugger detached
> ```

PoroCYon:
The log output means that it seems to work. But, what you're saying is, it works with terminal LLDB but not CLion's LLDB? That's a bit annoying, and not something I can troubleshoot easily (beyond guessing things, like commenting/uncommenting the code I highlighted).

RSDuck:
I already wanted to merge it a while ago, but I've been kept off by the bug reports in this PR.

PoroCYon:
@axel7083 did you actually manage to get it to work with a different GDB? (or maybe one of the newer commits automagically fixed things.)

turtleisaac:
I've made my own attempts to intercept the communication between CLion and its internal LLDB but that hasn't gone anywhere. However, I did get a different CLion installation on Linux which is actually able to use GDB to connect to melonDS, so the culprit definitely seems to be down to the implementation of their LLDB in terms of what it expects from the connected gdb-server, but the possibility to make it happy does exist.

Anyways, considering the following error message I've still been getting on the Mac side of things:
```
Process 1 was reported after connecting to 'connect://localhost:3333', but no stop reply packet was received
Debugger detached
```

I think it has something to do with [this page](https://sourceware.org/gdb/onlinedocs/gdb/Stop-Reply-Packets.html), but I'm not smart enough (and mostly just don't have enough information about CLion's internal LLDB and the sequence of events that led to that error) to understand what it is saying in relation to what melonDS either is or isn't doing...

PoroCYon:
> I think it has something to do with [this page](https://sourceware.org/gdb/onlinedocs/gdb/Stop-Reply-Packets.html), but I'm not smart enough (and mostly just don't have enough information about CLion's internal LLDB and the sequence of events that led to that error) to understand what it is saying in relation to what melonDS either is or isn't doing...

I don't think it means much. A "stop reply packet" is the packet sent from the gdbstub that tells the debug client why the debugged program is stopped (breakpoint, crash, user hit the 'pause now' button, etc.). Such a packet is supposed to be sent after a `?` or `vCont` command. My gdbstub does implement this (as you can see in the log you just sent me):

```
[GDB] command in: 'vCont;s:0001'
[GDB] send resp: '$E01#a6'
```

As the CLion debugger thingy disconnects immediately after sending a single packet (without requesting the stop reply packet in the first place), it seems to be another meaningless error on their end.

As for the thumb thing, I'm not sure what's up. It should be reported correctly in the CPSR. Maybe LLDB does want the thumb bit set in the program counter, too (unlike gdb, cf [this earlier discussion](https://github.com/melonDS-emu/melonDS/pull/1583?notification_referrer_id=NT_kwDOADGkFLI1MDc5Mjc3Nzk5OjMyNTMyNjg&notifications_query=is%3Aunread#issuecomment-1377370406)). Maybe you could try changing the `2` in [this line](https://github.com/melonDS-emu/melonDS/blob/7305cb3496cfeda902d0766c10f2a1b3ae5adcdf/src/ARM.cpp#L1006) to `1`, but I'm not sure if it'll help.

pleonex:
Thanks a lot for bringing this. It would be very useful for reverse engineering. I decided to give it a try but unfortunately I am couldn't make it on Windows with Ghidra or IDA. I think it's _almost_ working.

## Ghidra
- OS : Windows 11
- Ghida 10.3.2
- gdb-multiarch from msys2: GDB 13.2
- GDB launch command: `C:\\msys64\\mingw64\\bin\\gdb-multiarch.exe -i mi2 -iex 'set architecture arm'`
- Result
  - [x] Registers work
  - [ ] Memory viewing doesn't work (errors parsing hexadecimal output)
  - [ ] Step into: I need to send the command 2-3 times before it goes to the next instruction
  - [x] Breakpoints work
  - [ ] Break execution: it fails to pause after running
  - [x] Continue work
  - [x] Stop: melonDS finishes
- Same results if I run gdb without Ghidra directly (but memory display works)
- Logs: 
  - [ghidra_melonds.txt](https://github.com/melonDS-emu/melonDS/files/12705542/ghidra_melonds.txt)
  - [ghidra_errors.txt](https://github.com/melonDS-emu/melonDS/files/12705543/ghidra_errors.txt)
  - [ghidra_gdb-output.txt](https://github.com/melonDS-emu/melonDS/files/12705544/ghidra_gdb-output.txt)

## gcc-arm-none-eabi-10.3-2021.10
Same issue with step into as Ghidra. Rest seems to work via CLI but for some reason Ghidra rejects this GDB.

## IDA
- IDA Home ARM Version 8.3.230608 Windows x64
- Result:
  - [x] Register work
  - [x] Memory view: work
  - [ ] Step into: IDA throws a popup with the error `Single-stepping a thread failed; possibly stub does not support single-stepping. Try disabling the single-step functionality in debugger-specific options.`
  - [x] Breakpoint: ? can't step or continue but the breakpoint is added in the UI
  - [ ] Continue: fails with `Command "ProcessStart" failed`
  - [x]  Break execution: works if I join the emulator without setting "break on startup"
  - [ ] Stop: melonDS still running even after closing window, ignores control+c 
- Logs: [melonds_idahomearm.txt](https://github.com/melonDS-emu/melonDS/files/12705525/melonds_idahomearm.txt)

PoroCYon:
I'm having trouble reproducing some of these, and if I don't, it's exceptionally hard to even figure out what's going on in the other tools. This difficulty in reproduction might be because I'm using a different platform (Linux, also I don't have IDA). But either way,

* GDB itself seems to work. The only 'real' issue with GDB itself, from your report, seems to be the single-stepping. Right after reset release, the very first instruction seems to be 'duplicated' a few times, but after that, things work normally:
```
0xffff0000 in ?? ()
(gdb)stepi
0xffff0000 in ?? ()
(gdb)
(gdb)stepi
0xffff0000 in ?? ()
(gdb)stepi
0xffff00c0 in ?? ()
(gdb)stepi
0xffff8000 in ?? ()
(gdb)stepi
0xffff8004 in ?? ()
(gdb)stepi
0xffff8008 in ?? ()
(gdb)stepi
0xffff8078 in ?? ()
(gdb)stepi
0xffff807c in ?? ()
(gdb)stepi
0xffff8080 in ?? ()
(gdb)stepi
0xffff8084 in ?? ()
(gdb)stepi
0xffff8084 in ?? ()
(gdb)stepi
0xffff8088 in ?? ()
(gdb)stepi
0xffff808c in ?? ()
(gdb)stepi
0xffff8090 in ?? ()
(gdb)stepi
0xffff8094 in ?? ()
(gdb)stepi
0xffff8098 in ?? ()
(gdb)disassemble $pc,+0x20
Dump of assembler code from 0xffff8098 to 0xffff80b8:
=> 0xffff8098:	ldr	r0, [pc, #148]	@ 0xffff8134
   0xffff809c:	mcr	15, 0, r0, cr6, cr0, {0}
   0xffff80a0:	ldr	r0, [pc, #144]	@ 0xffff8138
   0xffff80a4:	mcr	15, 0, r0, cr6, cr1, {0}
   0xffff80a8:	ldr	r0, [pc, #140]	@ 0xffff813c
   0xffff80ac:	mcr	15, 0, r0, cr6, cr2, {0}
   0xffff80b0:	ldr	r0, [pc, #136]	@ 0xffff8140
   0xffff80b4:	mcr	15, 0, r0, cr6, cr3, {0}
End of assembler dump.
```
* Ghidra does indeed seem to give errors wrt. something about threads not working. Poking around in the [relevant](https://github.com/NationalSecurityAgency/ghidra/blob/650808862311b7c72ab009fc15a38ddbc5fb1ab6/Ghidra/Debug/Debugger-agent-gdb/src/main/java/agent/gdb/model/impl/GdbModelTargetThread.java#L100-L103) Ghidra [sources](https://github.com/NationalSecurityAgency/ghidra/blob/650808862311b7c72ab009fc15a38ddbc5fb1ab6/Ghidra/Debug/Debugger-agent-gdb/src/main/java/agent/gdb/model/impl/GdbModelTargetThreadContainer.java#L130-L158), however, lead to not much results because the Ghidra source code is an impenetrable mess of Enterprise-Grade Java Bullshit. I thus have no idea what it wants or what it's doing (it's even discarding the exception messages, which is *great* for debugging this /s).
* As for the memory errors, I have no idea what's going on here. In the melonDS log, everything seems to be going fine. The Ghidra output complains about an "invalid hex digit 103" (that's the ascii code of `g` in decimal), but this doesn't even appear in the data melonDS' gdbstub is sending. The problem thus lies elsewhere.
* Re: `gcc-arm-none-eabi-10.3-2021.10`: First of all, that's a build from Arm (the company), not from the people who make GDB (10.3 isn't even a real release number in the official version). So that's already starting to be fishy. But I downloaded the source and went ahead anyway, and saw no difference with GDB-multiarch 13.2. So I suppose it's just a matter of Ghidra being annoying.
* I don't have IDA, so I can't debug this. However, `Command "ProcessStart" failed` sounds like IDA is doing something it isn't supposed to (trying to create a new process, instead of using the gdbstub remote), which will naturally fail. This might be because it's either misconfigured, or the implementation is bugged and doesn't support remote gdbstubs.

pleonex:
Thanks for taking a look.

I think I have found the potential issue with the step / continue. IDA uses the command `vCont;s` to step in. The method `Handle_v_Cont` checks the command at `cmd[0]` but that's the semicolon `;`. Changing to `cmd[1]` fixes that issue. 
The following patch fixes it, now it can step, continue (no more _ProcessStart_ failed errors) and stop with IDA.

```diff
diff --git a/src/debug/GdbCmds.cpp b/src/debug/GdbCmds.cpp
index 057502f2..d642ef6d 100644
--- a/src/debug/GdbCmds.cpp
+++ b/src/debug/GdbCmds.cpp
@@ -883,13 +883,13 @@ ExecResult GdbStub::Handle_v_MustReplyEmpty(GdbStub* stub, const u8* cmd, ssize_
 
 ExecResult GdbStub::Handle_v_Cont(GdbStub* stub, const u8* cmd, ssize_t len)
 {
-       if (len < 1)
+       if (len < 2)
        {
                stub->RespStr("E01");
                return ExecResult::Ok;
        }
 
-       switch (cmd[0])
+       switch (cmd[1])
        {
        case 'c':
                stub->RespStr("OK");
```

There is still the issue that sometimes it requires to send several times the _step_ command to advance. I can reproduce that one with GDB directly. It happens at the beginning and randomly from time to time. In the following output it happened at: `0x02004800 `, `0x02004a30`, `0x02004a78` and `0x02004a94`
[doublestepi_gdboutput.txt](https://github.com/melonDS-emu/melonDS/files/12711259/doublestepi_gdboutput.txt)
[doublestepi_melonds.txt](https://github.com/melonDS-emu/melonDS/files/12711260/doublestepi_melonds.txt)

So the last issue before I can easily use it, it's that _step over_ doesn't work. It seems IDA is sending a regular `vCont;s` when we need a step over, so it goes inside the subroutine as expected. Comparing with desmume, in that case IDA first requests to setup a temp breakpoint after the jump and then it sends the command to continue (`s` in its case). So I am guessing something in the initial information melonds sends is configuring IDA differently (desmume doesn't support even half of the commands melonds does). I'll play with it to see if I find the cause.

zurgeg:
For me, GDB (with PEDA) doesn't show as much information. Take for instance an example executable:
```
[----------------------------------registers-----------------------------------]
RAX: 0x555555555175 (<main>:    push   rbp)
RBX: 0x7fffffffe1a8 --> 0x7fffffffe5c1 ("/home/jg/lame-patcher/example/main")
RCX: 0x555555557dd8 --> 0x5555555550f0 (endbr64)
RDX: 0x7fffffffe1b8 --> 0x7fffffffe5e4 ("PWD=/home/jg/lame-patcher/example")
RSI: 0x7fffffffe1a8 --> 0x7fffffffe5c1 ("/home/jg/lame-patcher/example/main")
RDI: 0x1 
RBP: 0x7fffffffe090 --> 0x1 
RSP: 0x7fffffffe090 --> 0x1 
RIP: 0x555555555179 (<main+4>:  lea    rax,[rip+0xea0]        # 0x555555556020)
R8 : 0x0 
R9 : 0x7ffff7fcecd0 (endbr64)
R10: 0x7fffffffddc0 --> 0x800000 
R11: 0x202 
R12: 0x0 
R13: 0x7fffffffe1b8 --> 0x7fffffffe5e4 ("PWD=/home/jg/lame-patcher/example")
R14: 0x7ffff7ffd000 --> 0x7ffff7ffe2d0 --> 0x555555554000 --> 0x10102464c457f 
R15: 0x555555557dd8 --> 0x5555555550f0 (endbr64)
EFLAGS: 0x246 (carry PARITY adjust ZERO sign trap INTERRUPT direction overflow)
[-------------------------------------code-------------------------------------]
   0x555555555174 <callme+43>:  ret
   0x555555555175 <main>:       push   rbp
   0x555555555176 <main+1>:     mov    rbp,rsp
=> 0x555555555179 <main+4>:     lea    rax,[rip+0xea0]        # 0x555555556020
   0x555555555180 <main+11>:    mov    rdi,rax
   0x555555555183 <main+14>:    call   0x555555555030 <puts@plt>
   0x555555555188 <main+19>:    mov    DWORD PTR [rip+0x2e92],0x1        # 0x555555558024 <global_variable>
   0x555555555192 <main+29>:    mov    eax,DWORD PTR [rip+0x2e8c]        # 0x555555558024 <global_variable>
[------------------------------------stack-------------------------------------]
0000| 0x7fffffffe090 --> 0x1 
0008| 0x7fffffffe098 --> 0x7ffff7dd6cd0 (mov    edi,eax)
0016| 0x7fffffffe0a0 --> 0x7fffffffe190 --> 0x7fffffffe198 --> 0x7ffff7f91000 --> 0x7ffff7daf000 --> 0x3010102464c457f 
0024| 0x7fffffffe0a8 --> 0x555555555175 (<main>:        push   rbp)
0032| 0x7fffffffe0b0 --> 0x155554040 
0040| 0x7fffffffe0b8 --> 0x7fffffffe1a8 --> 0x7fffffffe5c1 ("/home/jg/lame-patcher/example/main")
0048| 0x7fffffffe0c0 --> 0x7fffffffe1a8 --> 0x7fffffffe5c1 ("/home/jg/lame-patcher/example/main")
0056| 0x7fffffffe0c8 --> 0x50a969a1adaea7a0 
[------------------------------------------------------------------------------]
Legend: code, data, rodata, value
```
and then on melonDS:
```
0x02083590 in ?? ()
```

Obviously I can just examine PC:
```
gdb-peda$ x/i $pc
=> 0x2083590:   swpb    r5, r5, [r1]
```

Maybe this is just an issue with PEDA or GDB. I don't really know. But yeah, there you go

PoroCYon:
I'm not sure what you're trying to convey. Why is the disassembly x86?

If you want symbols and a usable backtrace in the NDS binary being debugged, you have to load it first:

```
arm-none-eabi-gdb -r 'target remote localhost:3333' path/to/executable.elf
```

zurgeg:
> I'm not sure what you're trying to convey. Why is the disassembly x86?
> 
> If you want symbols and a usable backtrace in the NDS binary being debugged, you have to load it first:
> 
> ```
> arm-none-eabi-gdb -r 'target remote localhost:3333' path/to/executable.elf
> ```

The example executable is x86. I was showing what normally happens on a standard debugging situation.

I'm not sure how to get the ELF file correctly. I attempted to `objcopy` the `arm9.bin` into an ELF file, because when I load it it gives `"/home/jg/pokemon-white-shenanigans/arm9.elf" is not a core dump: file format not recognized`.

But anyway, that's a problem for elsewhere

RSDuck:
I really don't want to let this PR rot, so I'm just going to merge for now.

PoroCYon:
> I'm not sure how to get the ELF file correctly. I attempted to `objcopy` the `arm9.bin` into an ELF file, because when I load it it gives `"/home/jg/pokemon-white-shenanigans/arm9.elf" is not a core dump: file format not recognized`.

Yeah, no, you're supposed to get your ELF file from the compiler you use to compile the source code with. If you aren't debugging something that comes from an ELF file or with source code, there isn't much you can do to get symbols or stack traces.

> I really don't want to let this PR rot, so I'm just going to merge for now.

I guess that's one way of doing it :D

zurgeg:
> > I'm not sure how to get the ELF file correctly. I attempted to `objcopy` the `arm9.bin` into an ELF file, because when I load it it gives `"/home/jg/pokemon-white-shenanigans/arm9.elf" is not a core dump: file format not recognized`.
> 
> Yeah, no, you're supposed to get your ELF file from the compiler you use to compile the source code with. If you aren't debugging something that comes from an ELF file or with source code, there isn't much you can do to get symbols or stack traces.
> 
> > I really don't want to let this PR rot, so I'm just going to merge for now.
> 
> I guess that's one way of doing it :D

Ah okay, I was more wondering about it showing the current instruction and stack, but that's off topic. It's good that this got merged though!

SeleDreams:
i noticed this is supposed to be merged in master but i don't see the option on melonds 0.9.5.
does it mean it's not yet in public releases and i have to build it from source ?

DaisukeDaisuke:
In order to use GDB with melonds, I created a single zip file containing the related libraries for gdb-remote and cross-compiled a standalone gdb_run.bat with Docker
However, in the newest Ghidra, the GUI appears to have very limited functionality with melonds
I hope this is a result of my own compilation fault
It seems that in the latest Ghidra, the target connection has been removed, forcing the installation of a custom protocol into GDB
However, this build disables the cli graphical interface, so you should probably use the one provided

- [x] Connect to gdb
- [x] gdb commands
- [x] Stopping and Restarting
- [x] `info all-registers`
- [ ] See the register in real time on gui
- [ ] GUI Something

https://github.com/DaisukeDaisukeForks/gdb-remote-for-ghidra-windows/releases

PS: The smartest way to do this is to connect to host.docker.internal in docker desktop and debug using pwndbg

PR #1587, opened 2022-12-20
BIOS/NAND UI changes

yeah-its-gloria:
Adds a configuration entry that remembers the last used folder for BIOS files, and adds `.mmc` to the list of valid NAND filename extensions.


Epicpkmn11:
> adds .mmc to the list of valid NAND filename extensions

If you're adding no$gba extensions, might be a good idea to add `.sd` to the SD image extensions too

src/frontend/qt_sdl/EmuSettingsDialog.cpp
```diff
@@ -32,11 +32,21 @@
...
 
-extern std::string EmuDirectory;
 extern bool RunningSomething;
 
 bool EmuSettingsDialog::needsReset = false;
 
+inline void updateLastBIOSFolder(QString& filename)
+{
+    int pos = filename.length() - 1;
+    while (filename[pos] != '/' && filename[pos] != '\\' && pos > 0) pos--;
```
RSDuck, line 42: I think you can just use one of the lastIndexOf overloads.

PR #1607, opened 2023-01-21
Add save state hotkeys

r3claimer:
This adds a basic save / load state hotkey feature. I have only tested this on linux but it should work on all platforms. Includes UI integration to map the hotkey under the general hotkey tab. 

Only adds the option to set a hotkey for state slot 1. But I imagine that someone could add the ability to switch which state slot to use pretty easily.


patataofcourse:
Usually what you see is a hotkey for each state and "restore from newest slot" + "save to oldest slot", i'd say that just a hotkey for the first slot is not very useful

src/frontend/qt_sdl/Config.cpp
```diff
@@ -26,8 +26,8 @@
 namespace Config
 {
 
-int KeyMapping[12];
-int JoyMapping[12];
+int KeyMapping[14];
```
RSDuck, line 29: why are you increasing this?

brooksytech: There are now 14 hotkey variables for both keys and joypads.

brooksytech:
> Usually what you see is a hotkey for each state and "restore from newest slot" + "save to oldest slot", i'd say that just a hotkey for the first slot is not very useful

I wrote this to use as a feature for Jelos, which is a handheld linux distro. Not disagreeing that it would be more useful the way you describe it but it fits our use case where we are limited to only a handful of buttons that arent already mapped to controls.

src/frontend/qt_sdl/Config.cpp
RSDuck: these arrays store the mapping for keyboard keys and gamepad to DS keys (ABXY, Dpad, L, R, Start, Select), not hotkeys.

The reason you changed this was probably, because before your change there were 12 hotkeys as well, though that's a pure coincidence.

The hotkey mapping are stored in HKKeyMapping and HKJoyMapping. They were automatically resized to 14 elements when you added elements to the enum, thanks HK_MAX moving further to the back.

brooksytech: Ah okay. Makes sense. I can revise this.

brooksytech:
@RSDuck I resolved the issues you pointed out. Thanks.

crashGG:
This commit is very practical, most of the time we just need a quick one-click S/L to get through some tricky positions in the stages of the game.

PR #1630, opened 2023-02-26
DSi power button and volume switch support

Ed-1T:
Proposed features:

- Accurate BPTWL power button, volume switch and battery level IRQ behavior.
- Accurate BPTWL read/write behaviors, such as value clamping and auto IRQ flag clearing.
- Simple power button and volume switch hotkey implementations.
- Output volume synchronization with the DSi's volume level.

Notes:

- Hotkey timings are merely an approximation and do not depend on the emulator's speed, but rather on the timing provided by SDL (this is for simplicity's sake and also because we have pretty much zero information on the BPTWL's internal code).
- The power button hardware reset feature currently has some issues as SoftReset can only be triggered from within an instruction. A workaround (as suggested by @RSDuck) was used instead.


src/frontend/qt_sdl/main.cpp
```diff
@@ -3027,7 +3077,9 @@ void MainWindow::onCameraSettingsFinished(int res)
 
 void MainWindow::onOpenAudioSettings()
 {
-    AudioSettingsDialog* dlg = AudioSettingsDialog::openDlg(this);
+    AudioSettingsDialog* dlg = AudioSettingsDialog::openDlg(this, emuThread->emuIsActive());
+    connect(emuThread, &EmuThread::syncVolumeLevel, dlg, AudioSettingsDialog::onSyncVolumeLevel);
```
RSDuck, line 3081: I think you need to explicitly take the address of onSyncVolumeLevel here for the error to go away, though I'm not sure.

RSDuck (approved):
looks good. We're probably going to merge this weird error is sorted out.

RSDuck:
I was a bit confused about you closing the issue, is there any problem or can it be merged?

Ed-1T:
I think Github closed it when I tried updating it from master, still don't know why.
Anyway, there's no problem with the PR, it can be merged.

PR #1639, opened 2023-03-20
Improved settings dialog

yeah-its-gloria:
Adds a new and improved:tm: settings dialog for melonDS.

~~still work in progress so yeah~~


yeah-its-gloria:
once again interest was lost, and I am abandoning these efforts

PR #1640, opened 2023-03-21
Introduce `Platform::Log`

JesseTG:
I'm writing [a new libretro core for melonDS](https://github.com/JesseTG/melonds-ds), and I needed to ensure that melonDS logs to the same place as RetroArch (or whatever other libretro implementation is in use). There's no way to customize how melonDS logs output, so I took matters into my own hands.

Specifically, this pull request:

- Adds a new `Log` function in the `Platform` namespace
- Adds a `LogLevel` enum
- Replaces most calls to `printf` with `Platform::Log` and a `LogLevel` that seemed appropriate in context
- Implements `Platform::Log` in the default frontend
- Does _not_ replace calls to `printf` within embedded dependencies (e.g. teakra)
- Does _not_ change the contents of any log
- Does _not_ use `LogLevel`.

At the moment, all log entries are written via `printf` regardless of the provided `LogLevel`. However, the pieces are in place for proper filtering.

### DIFF
--- a/src/SPI.cpp
+++ b/src/SPI.cpp
@@ -427,7 +427,7 @@ void Reset()
         }
     }
 
-    Log(LogLevel::Debug, "MAC: %02X:%02X:%02X:%02X:%02X:%02X\n",
+    Log(LogLevel::Info, "MAC: %02X:%02X:%02X:%02X:%02X:%02X\n",
            Firmware[0x36], Firmware[0x37], Firmware[0x38],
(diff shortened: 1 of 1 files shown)


src/Platform.h
```diff
@@ -128,6 +128,15 @@ inline bool LocalFileExists(std::string name)
     return true;
 }
 
+enum LogLevel {
```
RSDuck, line 131: braces on the next line

JesseTG: Done.

PR #1643, opened 2023-03-23
Expose SRAM pointers for frontends that manage save data their own way

JesseTG:
I wanted to use libretro's APIs for the [new libretro core I'm writing](https://github.com/JesseTG/melonds-ds) for melonDS, so that it can take advantage of whatever benefits the frontend provides (the details of which are beyond the scope of this post).

In order to do that, I need to return a pointer to the emulated SRAM from [`retro_get_memory_data`](https://github.com/libretro/RetroArch/blob/master/libretro-common/include/libretro.h#L3956-L3958), so that the frontend knows where to load and save the game data. However, the existing APIs in [NDSCart.h](https://github.com/melonDS-emu/melonDS/blob/master/src/NDSCart.h) and [GBACart.h](https://github.com/melonDS-emu/melonDS/blob/master/src/GBACart.h) are insufficient for this purpose.

This pull request exposes a pointer to the SRAM for supported NDS and GBA carts, i.e. subclasses of `NDSCart::CartRetail` and `GBACart::CartRetail` respectively. For other types of cartridges, the returned pointer is `nullptr` and the returned SRAM size is `0`.

I was warned that arbitrary modifications to SRAM (which this PR allows) can cause problems. I've added doc comments to this effect.

I have tested the NDS portion of this PR on my libretro core, and it works like a charm. I have not tested the GBA portion yet.

### DIFF
--- a/src/GBACart.cpp
+++ b/src/GBACart.cpp
@@ -100,6 +100,15 @@ void CartCommon::SRAMWrite(u32 addr, u8 val)
 {
 }
 
+u8* CartCommon::GetSaveMemory() const
+{
+    return nullptr;
+}
+
+u32 CartCommon::GetSaveMemoryLength() const
+{
+    return 0;
+}
 
 CartGame::CartGame(u8* rom, u32 len) : CartCommon()
 {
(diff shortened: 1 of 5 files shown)


CasualPokePlayer:
This would be nice for the current BizHawk fork as a note (I do something [similar here](https://github.com/TASEmulators/BizHawk/blob/10423abdf9575d1cc3119e2d3b1aea1fc5f47dce/waterbox/melon/BizInterface.cpp#L161-L173) so I can bypass melon's APIs for saving (due to BizHawk wanting a get entire save block on demand API). However, I still use the NDS::LoadSave API, as that may do some extra work.

Issue #1653, opened 2023-03-31
Building without Wayland libraries


dakrk:
Ever since DuckStation's OpenGL code was pulled in, I've been required to have Wayland libraries on my system. Ideally this should be able to be toggled with the build system as not all systems use it or even have it, and it's not core to function anyway as libwayland is loaded during runtime and the software continues to work if not found.

nadiaholmquist:
Realistically most Linux systems should have these libraries anyway so it's not really a big deal unless you're running some ultra-minimal customized system where you really don't want any library you don't strictly need for things to work.

If you can think of a way to solve this without breaking Wayland support though then suggestiosn are welcome.

dakrk:
It should just be a CMake option and a few conditions for a `USE_WAYLAND` option or whatever inside `duckstation/gl/context.cpp` and `qt_sdl/CMakeLists.txt` to exclude the Wayland context source file from building, the Wayland CMake module from being processed, the header being included, and the WindowInfo condition that tries to access ContextEGLWayland which won't be present as the header isn't being included. For the most part it should just be `if(` and `#ifdef`s.

nadiaholmquist:
What's the point though? Are there any distros where the Wayland libraries are not available?

dakrk:
Admittedly I don't have too much of a reason to nowadays other than preference and absolutely not liking having a bunch of data on my system that's never *actually* going to be used.

I used to start taking action on that because for the longest time I've had machines with barely any storage, not so good IO speeds both on my disk and my network, and generally just not a lot of resources (however that is thankfully no longer the case). Excluding such things I didn't need although at first may seem negligible, but in the end had some marginal gains which were quite much helpful in my situation. I imagine there are plenty of people still in such situations, and this would help for them.

Wayland support in this software isn't at all rooted too deeply in how it works, and allowing to disabling it should only be a few minor build system changes and preprocessor directives away. I therefore don't see any reason to not make those changes. You can already run the binary produced without having the library anyway, as there's not really any sort of dependency on it as it's loaded during runtime.

PR #1654, opened 2023-04-02
Allow toggling Wayland support

dakrk:
Allow toggling Wayland support, as reported in #1653.

Plenty of software allows doing this, so this project would just be an outlier otherwise.


dakrk:
Yes, the `ENABLE_WAYLAND` option is `ON` by default.

PR #1663, opened 2023-04-09
Discord rpc take 2

EndangeredNayla:
What is different from the #1579 commit. Well this adds gameid support. so discord can have box art key using gameids

This would implement the request from #318, just, uhh, 4 years late


yeah-its-gloria:
Not adding game box art was a conscious choice I made, Discord doesn't give us the storage to store even nearly enough of the box art for the games people would play, and the only emulator I know of that actually shows game box art is Dolphin (only for some games anyway), e.g Citra, Cemu, Ryujinx (although there is Ryujinx/Ryujinx#4391, the way they go about images might be of interest to you), yuzu all don't

EndangeredNayla:
Ive been working on adding support for this i just dont have enough C++ knowledge.   GamesTDB has a endpoint for DS Games

yeah-its-gloria:
Implemented it properly in my own PR 👍

PR #1667, opened 2023-04-15
Support loading Zstandard-compressed ROMs

nadiaholmquist:
This feature adds support for loading individual ROMs compressed using [Zstandard](https://github.com/facebook/zstd).

Zstandard is able to achieve a good compression ratio while also having incredibly fast (de)compression speeds, so no noticeable delay is added when loading even very large compressed ROMs.

This is different from the archive support in that the compressed ROMs are standalone files, rather than archives, making it possible to use them exactly as if they were regular ROMs, while saving a bunch of space on disk. This is supported both for DS and GBA ROMs, though given GBA ROMs' generally small size it's mostly useful for the former.

I'm posting this as a PR rather than directly committing it to see what people think about this feature, it's relatively close to the existing archive feature, so it might be superfluous to have this too, but it's something I'd personally want to use.

To test this, install Zstandard (should be in all Linux package repositories and Homebrew for macOS), and then simply run `zstd YourROM.nds`, and you should be able to run the resulting `YourROM.nds.zst` in melonDS.

### DIFF
--- a/.github/workflows/build-ubuntu-aarch64.yml
+++ b/.github/workflows/build-ubuntu-aarch64.yml
@@ -33,7 +33,7 @@ jobs:
         rm /etc/apt/sources.list
         mv /etc/apt/sources.list{.new,}
         apt update
-        DEBIAN_FRONTEND=noninteractive apt install -y {gcc-10,g++-10,pkg-config}-aarch64-linux-gnu {libsdl2,qtbase5,qtbase5-private,qtmultimedia5,libslirp,libarchive}-dev:arm64 cmake extra-cmake-modules dpkg-dev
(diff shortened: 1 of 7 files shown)


yazinnnn:
hi, i tried to load a .nds.zst file, but the melonDS had crashed

the git commit log 
```
commit ea1755bed0378a2f3783e78c5e63c266583218ce (HEAD -> master, origin/master, origin/HEAD)
Author: RSDuck <RSDuck@users.noreply.github.com>
Date:   Tue Mar 12 09:23:20 2024 +0100

    call Start again NDS object after Reset
    fixes issue where game doesn't properly start after changing settings

.......
```

image


and here is the coredump file
[melon.coredump.zip](https://github.com/melonDS-emu/melonDS/files/14581784/melon.coredump.zip)

Issue #1704, opened 2023-06-18
Allow swapping small screens in hybrid-mode


mrcalin:
It would be so cool to be able to swap the right side small screens when hybrid mode is enabled, independently from which screen is used as the main one.

Golden Sun Dark Dawn, for example, plays some summonings cinematic by using both screens to split the video, so it ends up being played upside down.

Had a look at the current features and just saw someone asking if it is possible here https://melonds.kuribo64.net/board/thread.php?id=1265.

Thanks in advance and congrats for the awesome emulator!

PR #1705, opened 2023-06-21
Add Flipped Hybrid Screen Layout 

### LINKED ISSUE #1658: Allow magnified display in hybrid view mode to move to the right
I am right-handed and am using melonDS on my Steam Deck. Having the touchscreen blown up on the right instead of the left would allow me better touch accuracy (and benefit many right-handed PC handheld users).

HetchBug:
This commit added the "Flipped Hybrid" option to the screen layout options. It is similar to the "Hybrid" option but moves the large screen to the right and the two smaller screens to the left.
This feature will help for people who are right-handed using a touch screen on Steam Decks or similar devices. It works with all four rotations, swap screen option, and touch functionality.

Hybrid Layout:
Screenshot 2023-06-20 233722

Flipped Hybrid Layout:
Screenshot 2023-06-20 233735

Closes #1658

### DIFF
--- a/src/frontend/FrontendUtil.h
+++ b/src/frontend/FrontendUtil.h
@@ -34,6 +34,7 @@ namespace Frontend
 //     1 = vertical
 //     2 = horizontal
 //     3 = hybrid
+//     4 = flipped hybrid
 // * rotation: angle at which the DS screens are presented: 0/1/2/3 = 0/90/180/270
 // * sizing: how the display size is shared between the two screens
 //     0 = even (both screens get same size)
--- a/src/frontend/Util_Video.cpp
+++ b/src/frontend/Util_Video.cpp
@@ -31,13 +31,19 @@ namespace Frontend
 float TopScreenMtx[6];
 float BotScreenMtx[6];
 float HybScreenMtx[6];
+float HybFlippedScreenMtx[6];
 float TouchMtx[6];
 float HybTouchMtx[6];
+float HybFlippedTouchMtx[6];
 bool TopEnable;
 bool BotEnable;
+bool EitherHybEnable;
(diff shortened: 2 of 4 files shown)


RSDuck:
you're duplicating far too much code.

HetchBug:
> you're duplicating far too much code.

I see what you mean. I'll have to retry when I have time. Thanks.

HetchBug:
I just added a new toggle, which takes the place of how "Swap screens" worked with the hybrid layout. Now, the "Swap screens" toggle works the way it does in other layouts. The new toggle changes what the large screen displays (top/bottom). This addresses #1704 and includes hotkey support.

Quick reference for both hybrid layouts:
- **Swap screens**: swap the two small screens with each other.
- **Swap focused hybrid screen**: change large screen to show the top screen/bottom screen.

ttang4299:
Hi, will this be merged for the 1.0 update, whenever that comes?

ttang4299:
@Jaklyy Sorry for the mention, but since you will be taking over soon, I was wondering if this can be merged for the 1.0 update. It's kind of an important feature for me, as otherwise, it wouldn't be ergonomic to use touch controls in some games on my Steam Deck.

Jaklyy:
don't have merge perms, but code looks fine to me? Needs to be updated to stop conflicting with master though.
I will say that i do believe we had plans of our own to eventually rework the way screen layouts work to support more user configurability? so i think in theory something like this would come as a result of that work. Though merging something like this as a stop gap in the meantime would be probably be fine if that rework doesn't end up coming in time for 1.0

ttang4299:
Hey @HetchBug , just wondering, did you ever get a chance to resolve those conflicts?

PR #1709, opened 2023-06-25
Add microphone combobox using SDL

### LINKED ISSUE #1571: [Feature Request] Mic source selection

falsidge:
Closes #1571 
Uses SDL to list devices instead to avoid compatibility issues and possible issue with names being different in Qt.

### DIFF
--- a/src/frontend/qt_sdl/AudioInOut.cpp
+++ b/src/frontend/qt_sdl/AudioInOut.cpp
@@ -133,7 +133,12 @@ void MicOpen()
     whatIwant.channels = 1;
     whatIwant.samples = 1024;
     whatIwant.callback = MicCallback;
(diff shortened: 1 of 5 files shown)


nadiaholmquist:
Haven’t tested but changes look fine to me.

HetchBug:
I tested the changes on Linux (Steam Desk) and I can verify that it lists the correct devices. The input volume was a bit quiet, but it works in games.
Screenshot_20230627_111329

RSDuck:
thanks

PR #1714, opened 2023-07-03
Modest cleanups for DSi_NAND

JesseTG:
My latest chapter in the [melonDS DS saga](https://github.com/JesseTG/melonds-ds) has led me to DSiWare support. Since melonDS doesn't support direct-booting DSiWare, I took an alternative approach; temporarily installing DSiWare and save data for the duration of a session, then exporting it back to the host at the end.

This PR introduces some odds and ends to make that simpler, including:

- A `struct` definition for title metadata (minus the certificates)
- A function to import DSiWare from memory


src/DSi_TMD.h
```diff
@@ -0,0 +1,157 @@
...
+
+    /// Content SHA1 (on decrypted ".app" file)
+    u8 ContentSha1Hash[0x14];
+
+    [[nodiscard]] u32 GetVersion() const noexcept
+    {
+        return (ContentId[0] << 24) | (ContentId[1] << 16) | (ContentId[2] << 8) | ContentId[3];
+    }
+};
+
```
RayyanAnsari, line 49: Might be good to add a static assert for this struct as well.

src/DSi_TMD.h
```diff
@@ -0,0 +1,157 @@
...
+        return (TitleId[4] << 24) | (TitleId[5] << 16) | (TitleId[6] << 8) | TitleId[7];
+    }
+
+    [[nodiscard]] u32 GetIDNoByteswap() const noexcept
+    {
+        return *reinterpret_cast<const u32*>(&TitleId[4]);
+    }
+};
+
+static_assert(sizeof(TitleMetadata) == 0x208, "TitleMetadata is not 520 bytes!");
```
RayyanAnsari, line 153: decimal

src/DSi_TMD.h
```diff
@@ -0,0 +1,157 @@
...
+    u32 SignatureType;
+
+    /// Signature RSA-OpenPGP-SHA1 across 140h..207h
+    u8 Signature[0x100];
+
+    /// Signature padding/alignment (zerofilled)
+    u8 SignatureAlignment[0x3C];
+
+    /// Signature Name "Root-CA00000001-CP00000007", 00h-padded
+    u8 SignatureName[0x40];
```
RayyanAnsari, line 65: should be `char`, isn't it ASCII?

src/DSi_TMD.h
```diff
@@ -0,0 +1,157 @@
...
+    {
+        return *reinterpret_cast<const u32*>(&TitleId[4]);
+    }
+};
+
+static_assert(sizeof(TitleMetadata) == 0x208, "TitleMetadata is not 520 bytes!");
+
+}
+
+#endif //DSI_TMD_H
```
RayyanAnsari, line 157:
```suggestion
#endif // DSI_TMD_H
```

src/DSi_TMD.h
```diff
@@ -0,0 +1,157 @@
...
+    [[nodiscard]] u32 GetVersion() const noexcept
+    {
+        return (ContentId[0] << 24) | (ContentId[1] << 16) | (ContentId[2] << 8) | ContentId[3];
+    }
+};
+
+/// Metadata for a DSiWare title.
+/// Used to install DSiWare titles to NAND.
+/// @see https://problemkaputt.de/gbatek.htm#dsisdmmcdsiwareticketsandtitlemetadata
+struct [[gnu::packed]] TitleMetadata
```
RayyanAnsari, line 53: likewise, see the comment at `TitleMetadataContent`

src/DSi_TMD.h
```diff
@@ -0,0 +1,157 @@
...
+
+namespace DSi_TMD
+{
+
+struct [[gnu::packed]] TitleMetadataContent {
+    /// Content ID (00,00,00,vv) ;lowercase/hex ;"0000000vv.app"
+    u8 ContentId[4];
+
+    /// Content Index (00,00)
+    u16 ContentIndex;
```
RayyanAnsari, line 33: Do we need to comment each and every one of these? Some comments aren't really adding much info.

RayyanAnsari (changes requested):
Skimmed through `DSi_TMD.h` in particular, there are some things that could be improved.

src/DSi_TMD.h
```diff
@@ -0,0 +1,157 @@
...
+
+    /// Content Type (00,01) ;aka DSi .app
+    u16 ContentType;
+
+    /// Content Size (00,00,00,00,00,19,E4,00)
+    u64 ContentSize;
```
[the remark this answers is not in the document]
JesseTG: I was going by the sizes given by the GBATek docs. Let me take another look at the docs.

src/DSi_TMD.h
```diff
@@ -0,0 +1,157 @@
...
+#ifndef DSI_TMD_H
+#define DSI_TMD_H
+
+#include "types.h"
+#include <array>
+
+namespace DSi_TMD
+{
+
+struct [[gnu::packed]] TitleMetadataContent {
```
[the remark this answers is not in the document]
JesseTG: I wanted to to ensure that the fields could be easily accessed while keeping the size exact, but I'll just use member functions instead.

src/DSi_TMD.h
JesseTG: I figured that the `static_assert` for the overall `TitleMetadata` would suffice, but sure. Why not?

src/DSi_TMD.h
JesseTG: Good catch.

src/DSi_TMD.h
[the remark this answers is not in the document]
RayyanAnsari: The sizes are correct. An example of what I meant:
`u8 ContentIndex[2];`
instead of
`u16 ContentIndex;`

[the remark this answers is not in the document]
JesseTG: `ContentIndex` and `ContentType` still logically represent one value each, though. Does that representation offer an advantage I'm not aware of?

src/DSi_TMD.h
```diff
@@ -0,0 +1,146 @@
...
+
+static_assert(sizeof(TitleMetadataContent) == 36, "TitleMetadataContent is not 36 bytes!");
+
+/// Metadata for a DSiWare title.
+/// Used to install DSiWare titles to NAND.
+/// @see https://problemkaputt.de/gbatek.htm#dsisdmmcdsiwareticketsandtitlemetadata
+struct TitleMetadata
+{
+    u32 SignatureType;
+
```
RSDuck, line 58: This looks pretty blown up, I think it would make more sense to only put spaces to group struct items.

JesseTG: I can slim this down.

JesseTG: I slimmed it down.

PR #1717, opened 2023-07-07
Make librt a dependency only on Linux (not just `NOT APPLE`)

JesseTG:
librt was interfering with my ability to compile an Android build of [melonDS DS](https://github.com/JesseTG/melonds-ds). However, I discovered that it doesn't appear to actually be necessary for upstream melonDS. This patch makes life a little simpler for all of us.


JesseTG:
Whoops, guess I was wrong. I'll fix it.

PR #1722, opened 2023-07-14
Allow to build without X11

lanodan:
X11 being pretty much deprecated and absent in some embedded platforms.


RSDuck:
is this trolling?

lanodan:
>is this trolling?

No, it's not. Heck even desktop distros are removing XOrg support lately, libX11 is going away at some point.

jmjoy:
It's great to have `X11` as an option, as emulators such as retroarch, drastic, and ppsspp can run on devices without X-Windows.

PR #1728, opened 2023-07-24
Hopefully fix fastmem on macOS

nadiaholmquist:
This makes two very small changes:

* Add forgotten calls to `JitEnableWrite`/`JitEnableExecute` in `FaultHandler`
* Disable the immediate `munmap` of `MemoryBase` which, while it works on Linux, seems to cause problems on macOS. Additionally this should fix FreeBSD as it seems to behave the same as macOS here.

### DIFF
--- a/src/ARMJIT_Memory.cpp
+++ b/src/ARMJIT_Memory.cpp
@@ -687,7 +687,11 @@ bool FaultHandler(FaultDescription& faultDesc)
             rewriteToSlowPath = !MapAtAddress(faultDesc.EmulatedFaultAddr);
 
         if (rewriteToSlowPath)
+        {
+            ARMJIT::JitEnableWrite();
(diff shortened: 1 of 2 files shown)


RSDuck:
https://github.com/melonDS-emu/melonDS/blob/master/src/ARMJIT_Memory.cpp#L338

This needs to be replaced by mmaping anonymous memory in place again. Otherwise once memory is remapped (e.g. DTCM) the fastmem area will get holely.

RayyanAnsari:
iirc there was also a hack to keep fastmem by default disabled in config.

nadiaholmquist:
> iirc there was also a hack to keep fastmem by default disabled in config.
Just looked, all it does is disable it by default but doesn't keep it off if you turned it on. Should probably still be removed in this PR though.

PR #1795, opened 2023-07-25
Fix 1 Pixel Tall Vertical Slopes Being Assigned the Wrong Increment

Jaklyy:
Fixes anti-aliasing being calculated incorrectly on certain polygon edges, resulting in transparent holes in several objects.
There's a few different ways to fix this properly, so feel free to request a different fix, if desired.
Some examples of objects fixed by this include:
This table:
image image
This mailbox:
image image
This shadow:
image image

### DIFF
--- a/src/GPU3D_Soft.h
+++ b/src/GPU3D_Soft.h
@@ -294,7 +294,7 @@ private:
             // TODO: this is still not perfect (see for example x=169 y=33)
             if (ylen == 0)
                 Increment = 0;
-            else if (ylen == xlen)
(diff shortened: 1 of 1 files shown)


Jaklyy:
converting to draft: may have found an alternative fix that's completely different; needs more digging.

Jaklyy:
verified that 1x1 slopes don't get their aa broken by this change, that was the one thing i was worried about potentially regressing with this pr so it should be fairly safe to merge.
image

PR #1798, opened 2023-07-29
Make the NDS teardown more robust

JesseTG:
This pull request makes `NDS::DeInit` (and its constituent `DeInit` functions) less likely to crash. Specifically, each "destructor" now checks the state of the resources it manages before trying to clean them up, then sets them to `nullptr` after they're deallocated.

Additionally, some classes are refactored to manage their resources with RAII (constructors/destructors) instead of `Init()`/`DeInit()` methods.

I wrote this PR to simplify error handling in melonDS DS. Thanks to this PR I can abort the core after a failed initialization and clean up after myself without triggering a segfault and crashing the libretro frontend (as may happen with the current `DeInit` functions).


src/GPU3D.h
```diff
@@ -137,14 +137,12 @@ void Write32(u32 addr, u32 val);
...
 public:
-    Renderer3D(bool Accelerated);
     virtual ~Renderer3D() {};
 
     Renderer3D(const Renderer3D&) = delete;
     Renderer3D& operator=(const Renderer3D&) = delete;
 
-    virtual bool Init() = 0;
-    virtual void DeInit() = 0;
+    bool IsValid() const noexcept { return Valid; }
```
RSDuck, line 145: I don't like that it's possible to a such an invalid Renderer object lying around.

I'd prefer it if the InitRenderer functions would just be skipped from the frontend side in case Init fails.

JesseTG: This is generally why constructors throw exceptions, but since I don't see any exceptions in the codebase I assumed that melonDS doesn't use them.

Alternatively, I could hide the constructor and change the class to be created with a function that returns a `unique_ptr`.

Are either of these solutions okay?

JesseTG: I have addressed this in my latest series of commits.

RSDuck:
if initialising OpenGL failed, why would you even try to initialise the OpenGL renderer?

JesseTG:
In my case it might not have _failed_ so much as _not happened_ yet. When requesting an OpenGL context with libretro, you might not get it immediately. This should allow me to put off initializing the renderer until it's ready.

RSDuck:
if you have to call the renderer again later once you have the context anyway, why don't you just initalise the renderer then?

JesseTG:
I do. I just added this safeguard because the weird crash was annoying me. It's easier to diagnose a `return false` than an attempt to call a `NULL` function pointer, as the latter won't show up on a stack trace.

RSDuck:
then change it to an assert

JesseTG:
Fair enough. One moment, please.

PR #1800, opened 2023-08-07
Assorted portability enhancements

JesseTG:
This pull request introduces several assorted fixes and features that make it easier to build melonDS on multiple platforms. Most of them have no visible effect, but they make it easier to use it as a library.

- Reduces the severity of some `Platform::Log` calls to `Debug` or `Info`, because they don't represent actual problems like I initially believed.
- Change some `#include`s to use project-relative paths (rather than file-relative) paths. This made it easier to build `LAN_PCap` and `LAN_Socket` into my libretro core.
- Add `Platform` calls for loading, unloading, and fetching symbols from dynamic libraries. `LAN_PCap` uses these. There's [another place in `ARMJIT_Memory.cpp`](https://github.com/melonDS-emu/melonDS/blob/master/src/ARMJIT_Memory.cpp#L756) that could benefit as well, but I'm not ready to make and test the fix yet.
- Add `Platform` calls for all file operations that melonDS uses. My motivation here was to keep track of when emulated filesystems (DSi NAND and SD cards) are modified, so I can ensure they're flushed to disk a short time afterward.
- Apply a fix to `CommonFuncs.cpp` that was preventing it from compiling on Android.
- Add `Platform::EnterGBAMode`, which is called when the emulated DS attempts to enter GBA mode. This is intended to allow a frontend to gracefully stop the emulator and tell the player that GBA mode is unsupported; I introduce this capability into melonDS as well.

### DIFF
--- a/src/Platform.h
+++ b/src/Platform.h
@@ -158,7 +158,7 @@ bool CloseFile(FileHandle* file);
 bool IsEndOfFile(FileHandle* file);
 
 bool FileGetString(char* str, int count, FileHandle* file);
-bool FileSeek(FileHandle* file, s32 offset, FileSeekOrigin origin);
+bool FileSeek(FileHandle* file, s64 offset, FileSeekOrigin origin);
 void FileRewind(FileHandle* file);
 u64 FileRead(void* data, u64 size, u64 count, FileHandle* file);
 bool FlushFile(FileHandle* file);
--- a/src/Platform.h
+++ b/src/Platform.h
@@ -146,90 +146,9 @@ bool GetConfigBool(ConfigEntry entry);
 std::string GetConfigString(ConfigEntry entry);
 bool GetConfigArray(ConfigEntry entry, void* data);
 
-/// The type of file that the provided path points to.
-/// Possible use cases include:
-///     \li Validating a file path before opening it
-///     \li Validating a file before returning its contents
-///     \li Setting buffering/flushing options on some kinds of files
-///     \li Logging file operations
-///
-/// @note All files are binary unless otherwise specified.
-enum class FileType {
-    /// The file type is unspecified.
-    /// It should be opened as a binary file.
-    Generic,
-
-    /// Text file containing Action Replay cheat codes for the current game.
-    /// May be read or written.
-    ARCodeFile,
-
-    /// A ROM image for the NDS, including DSiWare.
-    /// Will not be written.
-    NDSROM,
-
-    /// A dump of an NDS game's SRAM.
-    /// Will be saved and loaded occasionally.
-    NDSSaveFile,
-
-    /// BIOS image for the NDS's ARM7 processor (not the DSi).
-    BIOS7,
-
-    /// BIOS image for the NDS's ARM9 processor (not the DSi).
-    BIOS9,
-
-    /// Firmware image for the NDS (not the DSi).
-    /// @note May be written to by the emulator.
-    Firmware,
-
-    /// A subset of the NDS or DSi firmware containing only the Wi-fi configuration.
-    WifiSettings,
-
-    /// BIOS image for the DSi's ARM7 processor.
-    DSiBIOS7,
-
-    /// BIOS image for the DSi's ARM9 processor.
-    DSiBIOS9,
-
-    /// Firmware image for the DSi.
-    /// @note May be written to by the emulator.
-    DSiFirmware,
-
-    /// An image of the DSi's NAND flash memory.
-    /// Frequently read and written to by the emulator in DSi mode.
-    DSiNANDImage,
-
-    /// An image of a virtual SD card.
-    /// May be frequently read and written to by the emulator in DSi mode
-    /// or when running homebrew.
-    SDCardImage,
-
-    /// Text file containing a list of files and accompanying metadata in the virtual SD card.
-    SDCardIndex,
-
-    /// A GBA ROM.
-    /// Will not be written.
-    GBAROM,
-
-    /// A dump of a GBA game's SRAM.
-    /// May be read or written.
-    GBASaveFile,
-
-    /// An unspecified file being copied to or from an emulated SD card or NAND image.
-    /// Should be opened in binary mode.
-    /// \c Read mode means the file is being copied to the emulated filesystem.
-    /// \c Write mode means the file is being copied to the host.
-    HostFile,
-
-    /// A configuration file for the frontend.
-    /// May be text or binary, depending on the frontend.
-    Config,
-};
-
 /**
  * Denotes how a file will be opened and accessed.
  * Flags may or may not correspond to the operating system's file API.
- * @note This enum does not specify text or binary mode;
- * that should be determined by the \c FileType.
  */
 enum FileMode : unsigned {
     None = 0,
@@ -263,6 +182,15 @@ enum FileMode : unsigned {
     NoCreate = 0b00'10'00,
 
     /**
+     * Opens a file in text mode,
+     * rather than the default binary mode.
+     * Text-mode files may have their line endings converted
+     * to match the operating system,
+     * and may also be line-buffered.
+     */
+    Text = 0b01'00'00,
+
+    /**
      * Opens a file for reading and writing.
      * Equivalent to <tt>Read | Write</tt>.
      */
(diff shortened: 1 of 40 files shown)


src/frontend/qt_sdl/Platform.cpp
```diff
@@ -566,4 +654,26 @@ void Camera_CaptureFrame(int num, u32* frame, int width, int height, bool yuv)
...
+{
+    return (DynamicLibrary*) SDL_LoadObject(lib);
+}
+
+void DynamicLibrary_Unload(DynamicLibrary* lib)
+{
+    SDL_UnloadObject(lib);
+}
+
+void *DynamicLibrary_LoadFunction(DynamicLibrary *lib, const char *name)
```
RSDuck, line 667: `void*`

src/Platform.h
```diff
@@ -201,6 +237,36 @@ void Camera_Start(int num);
...
+ */
+void DynamicLibrary_Unload(DynamicLibrary* lib);
+
+/**
+ * Loads a function from a library.
+ * @param lib The library to load the function from.
+ * @param name The name of the function to load.
+ * @return A pointer to the loaded function, or \c nullptr if the function could not be loaded.
+ */
+void *DynamicLibrary_LoadFunction(DynamicLibrary *lib, const char *name);
```
RSDuck, line 261: same here

RSDuck (commented):
I'm mostly ok with these changes.

Though I don't really such much of a point in replacing the C file functions with basically an exact wrapper. Because any platform which already has a somewhat freely accessible filesystem should have C files. And if not, you probably have to be pretty hacky implementation of this filesystem wrapper where the paths are compared.

I'd rather see this wrapper done all the way, so that e.g. path handling is only left where absolutely necessary (such as fatfs). Otherwise e.g. BIOS there would be just some kind of function which returns a "file handle" for it.

JesseTG:
> I'm mostly ok with these changes.
> 
> Though I don't really such much of a point in replacing the C file functions with basically an exact wrapper. Because any platform which already has a somewhat freely accessible filesystem should have C files. And if not, you probably have to be pretty hacky implementation of this filesystem wrapper where the paths are manually compared etc..

I see two:

1. It's not just about replacing the C functions. Wrapping them is a valid use case. For example, I manually flush files to disk a few frames after they've been written, but I reset that timer if the same file is written again. Here's what I do:

```c++
struct Platform::FileHandle {
    RFILE *file;
    FileType type;
};

static std::unordered_map<Platform::FileHandle*, int> flushTimers;

u64 Platform::FileWrite(const void* data, u64 size, u64 count, FileHandle* file)
{
    if (!file || !data)
        return 0;

    u64 result = filestream_write(file->file, data, size * count);
    switch (file->type) {
        case FileType::DSiNANDImage:
        case FileType::SDCardImage:
        case FileType::SDCardIndex:
        case FileType::Firmware:
        case FileType::DSiFirmware: {
            flushTimers[file] = melonds::config::save::FlushDelay();
        }
            break;
        default:
            break;
    }

    return result;
}

retro::task::TaskSpec melonds::file::FlushTask() noexcept {
    retro::task::TaskSpec task([](retro::task::TaskHandle &task) {
        ZoneScopedN("melonds::fat::FlushTask");
        using namespace melonds::file;
        if (task.IsCancelled()) {
            // If it's time to stop...
            task.Finish();
            return;
        }

        std::vector<Platform::FileHandle*> filesToRemove;

        for (auto& [file, timeUntilFlush] : flushTimers) {
            timeUntilFlush--;
            if (timeUntilFlush == 0) {
                // If the timer has reached 0, flush the file and remove it from the map

                retro_assert(file != nullptr);
                retro_assert(file->file != nullptr);

                libretro_vfs_implementation_file* handle = filestream_get_vfs_handle(file->file);
                const char* original_path = filestream_get_path(file->file);
                retro_assert(handle != nullptr);
                retro_assert(handle->fd >= 0);
                retro_assert(original_path != nullptr);
                // If the above conditions were false, the file should never have left Platform::OpenFile

#ifdef _WIN32
                int syncResult = _commit(handle->fd);
#else
                int syncResult = fsync(handle->fd);
#endif

                if (syncResult == 0) {
                    retro::debug("Flushed file \"%s\" to host disk", original_path);
                } else {
                    int error = errno;
                    if (error == EBADF) {
                        retro::info("File \"%s\" was closed behind our backs, no need to flush it to disk.", original_path);
                    }
                    else {
                        retro::error("Failed to flush \"%s\" to host disk: %s (%x)", original_path, strerror(error), error);
                    }
                }

                // If the file is not open, then it was closed before the flush timer reached 0.
                // In that case it will have been flushed to disk anyway, so we don't need to do anything.

                filesToRemove.push_back(file);
            }
        }

        for (auto& file : filesToRemove) {
            flushTimers.erase(file);
        }
    });

    return task;
}
```

2. [libretro's file system wrappers](https://github.com/libretro/RetroArch/blob/master/libretro-common/vfs/vfs_implementation.c) have platform-specific logic in numerous places. For example, it `mmap`'s open files under the hood if a file is opened with a `RETRO_VFS_FILE_ACCESS_HINT_FREQUENT_ACCESS` flag on supported platforms.

> I'd rather see this wrapper done all the way, so that e.g. path handling is only left where absolutely necessary (such as fatfs). Otherwise e.g. BIOS there would be just some kind of function which returns a "file handle" for it.

What do you have in mind?

nadiaholmquist:
Rather than having a specific function for reporting that the emulated system attempted to enter GBA mode, wouldn't it be better to have a platform function that can be called by the emulator for multiple reasons that could stop emulation? For example it could be that the user shut down the emulated system, attempted to enter GBA mode, or some other thing went wrong internally that the frontend might want to react to?

JesseTG:
> Rather than having a specific function for reporting that the emulated system attempted to enter GBA mode, wouldn't it be better to have a platform function that can be called by the emulator for multiple reasons that could stop emulation? For example it could be that the user shut down the emulated system, attempted to enter GBA mode, or some other thing went wrong internally that the frontend might want to react to?

You mean like `Platform::StopEmu` with an extra argument?

JesseTG:
Writing a generalized `Platform::OpenFile(FileType)` function has proven to be more complicated than expected, because various parts of the code (such as `ARCodeFile` handling) would have to be refactored extensively to accept `FileHandle*`s instead of paths _and_ deal with situations where non-existent files are non-erroneous.

In the interest of getting something out the door, I'm not going to write a generalized `Platform::OpenFile(FileType)` function; I will just stick with the existing `Platform::OpenFile` functions, except they'll contain optional `FileType` args and return `FileHandle*`s. What I may do instead is just add some extra entries to `Platform::GetConfigString` for things like wi-fi settings or firmware backups. That plus the `FileType` should be enough for the core to be ignorant of how a file is loaded, and for a frontend to treat certain file types differently when loading them.

JesseTG:
I've finished my work on this PR and happily await its review.

RSDuck:
I think that the optional file type is pretty pointless. At best it gives nicer logging, but otherwise e.g. in terms of file validation, what's there that the core itself shouldn't handle itself. It's going to process the file anyways.

JesseTG:
> I think that the optional file type is pretty pointless. At best it gives nicer logging, but otherwise e.g. in terms of file validation, what's there that the core itself shouldn't handle itself. It's going to process the file anyways.

I use it to keep track of which files to flush to manually flush, to minimize data loss in case of a crash or a power loss. These are typically long-lived files such as SD card images, the DSi NAND, or GBA save data.

I keep a counter for each file that I'm interested in manually flushing, decrementing it each frame or resetting it if it's written again. Once the timer expires, I call `fsync` (or local equivalent) on the file handle. I don't flush after every write because writes can occur multiple times in one frame, or spread across several frames.

Is there another approach you have in mind?

RSDuck:
yes I know and I think I should have expressed this earlier, but I'm not that much of a fan of what you're doing with the flush logic behind the file interface. It feels kind of out of place to me to put this behind what's effectively a generic file interface afterall.

I'm not sure what would be the better approach to this, though we already have a thread for asynchronous DS savefile writing.

JesseTG:
> yes I know and I think I should have expressed this earlier, but I'm not that much of a fan of what you're doing with the flush logic behind the file interface. It feels kind of out of place to me to put this behind what's effectively a generic file interface afterall.

I would argue that this is a good place for flushing logic _because_ it's generic. melonDS opens and closes files by its own volition, so right now the only way for me to know when to flush a file is by intercepting the open/close calls. Until such time as it's refactored to operate on file handles (or even raw memory) from the outside world, this is the best solution I can think of.

> I'm not sure what would be the better approach to this, though we already have a thread for asynchronous DS savefile writing.

And that's great, but it's not what I need. Nintendo DS SRAM is a solved problem in my case. I'm specifically interested in NAND images, SD card images, and firmware.

PR #1803, opened 2023-08-10
Fix aa being upside down on swapped y-major slopes

Jaklyy:
fixes kyurem's face on the title screen, also fixes crossed quads.
image
fixed
image
console screenshot
image

### DIFF
--- a/src/GPU3D_Soft.cpp
+++ b/src/GPU3D_Soft.cpp
@@ -753,8 +753,8 @@ void SoftRenderer::RenderShadowMaskScanline(RendererPolygon* rp, s32 y)
         interp_start = &rp->SlopeR.Interp;
         interp_end = &rp->SlopeL.Interp;
 
-        rp->SlopeR.EdgeParams_YMajor(&l_edgelen, &l_edgecov);
-        rp->SlopeL.EdgeParams_YMajor(&r_edgelen, &r_edgecov);
+        rp->SlopeR.EdgeParams_YMajor(&l_edgelen, &l_edgecov, 1);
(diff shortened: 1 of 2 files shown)


Jaklyy:
draft pr because it needs more testing, specifically i'd like more than two examples of this bug occurring, i think it's good though? also not sure whether i wanna do a light restructuring of some of the relevant code or if that's just unnecessary to worry about.

Jaklyy:
added improved handling for swapped vertical and x-major slopes.
should all be console accurate (or at least closer in the case of x-major slopes)

src/GPU3D_Soft.cpp
```diff
@@ -753,8 +753,8 @@ void SoftRenderer::RenderShadowMaskScanline(RendererPolygon* rp, s32 y)
         interp_start = &rp->SlopeR.Interp;
         interp_end = &rp->SlopeL.Interp;
 
-        rp->SlopeR.EdgeParams_YMajor(&l_edgelen, &l_edgecov);
-        rp->SlopeL.EdgeParams_YMajor(&r_edgelen, &r_edgecov);
+        rp->SlopeR.EdgeParams(&l_edgelen, &l_edgecov, 1);
```
RSDuck, line 756: please use true and false

Jaklyy:
took the time to rewrite it slightly to hopefully force it to pre-compile the if statements, wasn't sure if they would on its own the way i had done it. feel free to tell me if there's a better way to do this/if you think this is worse from a readability/optimization standpoint

src/GPU3D_Soft.cpp
Jaklyy: done

Issue #963, opened 2021-01-25
[Request] Savestate hotkey


Shideravan:
Would be nice to configure the savestate hotkey. I love to put the savestate and loadstate in a joystick button, which helps a lot in some games to create a fast save and reload them easy from a button press in joystick.

PR #1805, opened 2023-08-12
Save load state shortcuts

hcook95:
Added the ability to customize the shortcuts for saving and loading game states. This is especially helpful for those with keyboards that don't have very accessible function keys. 

Unfortunately, I was unable to get the shortcuts working with joystick hotkeys, as it doesn't seem that QAction shortcuts support input from a controller. Fortunately, many 3rd party tools (e.g. Steam) allow you to map the controller's buttons to keyboard inputs, achieving a similar effect.


nadiaholmquist:
Seems fine to me, though a workaround that only properly works for US-layout keyboards probably won't do, especially with none of the people with access to the repo being from the US :p

There's some nitpicks about code style you should fix to be consistent with the rest of the code ~~even if I do like your code style better~~, but otherwise I can't really think of any issues.

Is there a reason you're using QActions instead of triggering them the same way the other shortcuts work?

hcook95:
The reason I used QActions instead of triggering them the same way as the other hotkeys is because `MainWindow::onSaveState()` and `MainWindow::onLoadState()`  both use QActions. I wanted to limit how much code had to be refactored to support customizable save/load state hotkeys. Additionally, using the QAction allows the shortcut key combo to be displayed next to the action in the menu, which I figured is nice from a usability standpoint.

That being said, it probably is worth the refactor switching from QActions to the normal hotkey method. Doing so should fix both the joystick issues, as well as the weird problems with keys like "Shift+1". I'll work on the refactor and update this pull request with it in the next few weeks.

Also thank you for pointing out my coding style. I'll make sure to fix the formatting of the code to be compliant with the coding style in the contributing guide for my next pull request.

Thanks for the help!

nadiaholmquist:
I do really like how the QActions actually display their shortcut in the UI and how they're integrated with Qt.

If there's a way to trigger a QAction manually then maybe the better approach is to convert the other hotkeys into QActions and just have a mapping of gamepad buttons to QActions somewhere so we can trigger them on gamepad button presses.

hcook95:
Ok, I'll look into integrating QActions into hotkeys. This is my first time using QT (or really any GUI tool), so I'm still trying to learn the ropes. I might reach out if I have any questions.

hcook95:
@nadiaholmquist

I fixed it so that the load/save shortcuts are done the same as the other shortcuts. This allows joystick shortcuts to work with them, while also allowing shortcuts like "Shift+!" to work, (although, I still don't know why it appears as "shift+!" instead of "shift+1". The shortcuts should also be layout-agnostic. 

I also added functionality to have the shortcut appear in the menus. This currently works with the save/load actions and the pause and reset actions.

Finally, I should have my code consistent with the coding standard.

Let me know what you think.

hcook95:
@nadiaholmquist Any update on the status of this pull request?

hcook95:
@nadiaholmquist I think this pull request should be ready to be pulled into the master branch. Could you take a look when you get the chance?

Shideravan:
How are the state of this PR in recent code? Should this need too much refactor to work in recent code?

Shideravan:
Also, this is the first step for solving https://github.com/melonDS-emu/melonDS/issues/963

PR #1814, opened 2023-08-16
added customDNSLookup

TimOelrichs:
dear Arisotura / melonDs team

i added a custom DNS lookup, so that altWFC online feature should work out of the box, without user configuration.
See new options in the wifi settings. maybe you are interested.

thx for your great work with melonDS.

best regard 
tim


src/frontend/qt_sdl/CustomDnsQuery.h
```diff
@@ -0,0 +1,394 @@
+// DNS Query Program on Linux
```
RSDuck, line 1: this code seems to come from somewhere else with no attribution

src/frontend/qt_sdl/Config.cpp
```diff
@@ -347,7 +351,13 @@ ConfigEntry ConfigFile[] =
     {"Camera1_CamDeviceName", 2, &Camera[1].CamDeviceName, (std::string)"", false},
     {"Camera1_XFlip", 1, &Camera[1].XFlip, false, false},
 
-    {"", -1, nullptr, 0, false}
+    {"", -1, nullptr, 0, false},
+
+    {"useCustomDNS", 1, &useCustomDNS, true, true},
```
RSDuck, line 356: you have to add the new settings after the sentinel value indicating the last entry

RSDuck (commented):
Stylistically it's all over the place, take a look at https://github.com/melonDS-emu/melonDS/blob/master/CONTRIBUTING.md

Also what's the deal with the change of the cursor image?

src/frontend/qt_sdl/InputConfig/resources/cursor.png
JesseTG, line 1: Where did you get this cursor? I might want to use it for something myself, but only if I can be sure of the license.

TimOelrichs: I got it from http://www.rw-designer.com/cursor-set/nintendo-ds-stylus and converted it to png.

It's released under the Attribution Required (CC by) [license](http://www.rw-designer.com/licenses).
Name: Sunny Cloray
Email: globalreject@gmail.com

sorry, did not see that before, but it should be fine to use, if the original source is linked.

src/frontend/qt_sdl/CustomDnsQuery.h
TimOelrichs: yes, you're right. i got it from https://www.binarytides.com/dns-query-code-in-c-with-linux-sockets/

i'm sure there are better, shorter and prettier ways to implement this. i just wanted to test things out, without adding more dependencies. Than i got hyped because i got it working. I'm still kind of a noob. sry, for that

src/frontend/qt_sdl/InputConfig/resources/cursor.png
JesseTG: Awesome, thank you.

PR #1816, opened 2023-08-21
Fix Incorrect Polygon Swapping Behavior and Implement Correct Rules for Shifting Right Edges Left

Jaklyy:
Decided to combine these two fixes into one PR, as the latter is dependent on the former to work properly.
Can be split into two if requested.

**Fix Incorrect Polygon Swapping Behavior:**

I'll be honest I basically just yoinked this from NooDS just to see what would happen and it worked.
In short melonDS was (I'm not familiar with this type of math so I'm not gonna try to explain in detail) but it was using the second vertex where it should've used the first.
Changing that essentially got things working flawlessly... once I inverted everything that was determined by the dot.
possible I messed up something when changing that, since I'm not super familiar with what that part of the code is trying to do.
(should it be > instead of >= ? etc.)
But it fixes odd behavior with the second vertex and polygons being swapped where they shouldn't be.
Or in other words it fixes Mario and Luigi: Partners in Time's time travel scene being broken.

**Implement Correct Rules for Shifting Right Edges Left:**

Based on some personal research into swapped polygon behavior done on a 3ds.
Currently melonDS always shifts right vertical edges left, but there are situations where it should not do so.
This PR implements the rules for this and should result in most affected polygons rendering properly.
Fixes a number of objects in the gen 4 Pokemon games.
Some relevant polygons do not render their textures perfectly for reasons I have yet to fully determine.
Such as this one, which should have the topmost pixel be green:
image
Seems like the textures are being applied slightly too high up?

### DIFF
--- a/src/GPU3D.cpp
+++ b/src/GPU3D.cpp
@@ -1049,12 +1049,12 @@ void SubmitPolygon()
     v2 = &TempVertexBuffer[2];
     v3 = &TempVertexBuffer[3];
 
-    normalX = ((s64)(v1->Position[1]-v0->Position[1]) * (v2->Position[3]-v0->Position[3]))
-        - ((s64)(v1->Position[3]-v0->Position[3]) * (v2->Position[1]-v0->Position[1]));
-    normalY = ((s64)(v1->Position[3]-v0->Position[3]) * (v2->Position[0]-v0->Position[0]))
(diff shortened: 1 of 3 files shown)


Jaklyy:
Turns out only the change to how facing view is determined was necessary to fix the incorrect swapping behavior

Issue #1819, opened 2023-08-28
Regression after commit #d69745b3


Brankale:
This commit d69745b3 introduces strange shadow casting in the intro video of `The Legend of Zelda Spirit Tracks`. This behaviour is not present on real hardware.

before
![melonDS 0 9 5](https://github.com/melonDS-emu/melonDS/assets/45881310/c9754245-d934-48f0-b082-3e118bb472fc)

after
![melonDS latest](https://github.com/melonDS-emu/melonDS/assets/45881310/5516a56b-c748-4392-9241-74c4639f08b1)

before
![Immagine 2023-08-28 144428](https://github.com/melonDS-emu/melonDS/assets/45881310/bba3b569-cdbd-4664-8d80-d85ba1fbf23d)

after
![Immagine 2023-08-28 144757](https://github.com/melonDS-emu/melonDS/assets/45881310/39cb54dd-fdec-4db5-9de4-9f50a22fa884)

other images
![Immagine 2023-08-28 145358](https://github.com/melonDS-emu/melonDS/assets/45881310/e3d33192-b277-4ed2-b553-d3340531ec23)

Brankale:
The problem is

``` 
bool facingview = (dot <= 0);
```

This fixes the issue:
```
bool facingview = (dot < 0);
```

This thing was also mentioned in the pull request #1816 as a possible candidate for regressions.

Brankale:
I will not open a pull request since I'm not sure if this change was necessary. @Jaklyy Can you check if this is correct?

Jaklyy:
this would fix the regression, while reintroducing the bug this fix was aiming to resolve
correct fix seems to be to use <= only for determining if a polygon is supposed to be clockwise/counterclockwise
working on a fix right now

PR #1820, opened 2023-08-28
Fix regression caused by change to front face polygon culling

Jaklyy:
Fixes https://github.com/melonDS-emu/melonDS/issues/1819
Maintains the fix to polygon orientation from https://github.com/melonDS-emu/melonDS/commit/d69745b3a82ed75da71e93a0061581367084a6fe
~~It would appear that only the check for a polygon being counter-clockwise is supposed to be <=~~
front facing polygons are supposed to be culled if dot < 0

~~Not sure if I'm supposed to increment the savestate version whenever new things are added to it, or if that's only done on new releases
or if I even need to add that to savestates~~

### DIFF
--- a/src/GPU3D.cpp
+++ b/src/GPU3D.cpp
@@ -558,6 +558,7 @@ void DoSavestate(Savestate* file)
 
         file->Bool32(&poly->FacingView);
         file->Bool32(&poly->Translucent);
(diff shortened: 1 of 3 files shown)


RSDuck:
yeah adding it to the savestates is right, otherwise the case where a game a game doesn't repeatedly swaps polygon buffers wouldn't work.

PR #1835, opened 2023-09-24
Conditionally enable JIT_DEBUGPRINT

JesseTG:
This way, it can be toggled with a compiler flag instead of having to edit the source.


RSDuck:
I prefer it being toggled from source though, if you change the compiler flag you have to recompile everything.

PR #1843, opened 2023-10-03
Refactor DSi_NAND to manage the mount's lifetime with RAII

JesseTG:
This is another small prerequisite PR in my storage refactoring series.

This PR introduces the following small refactorings:

- `DSi_NAND::NANDMount` is introduced, which mounts/unmounts a file system image with fatfs. The image is mounted upon creation and unmounted upon destruction.
- The contents of `diskio.c` are moved out of fatfs and into the main `src` directory, since it consists of application-specific logic and declarations.
- The callbacks registered by `ff_disk_open` now accept `std::function`s instead of plain function pointers, so that they don't need to rely on a particular global `NANDMount`.

Once this PR is merged, next on my list will be to move the NAND's disk image loading logic from the core to the frontend.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -25,6 +25,7 @@ add_library(core STATIC
     DSi_NWifi.cpp
     DSi_SD.cpp
     DSi_SPI_TSC.cpp
+    FATIO.cpp
     FATStorage.cpp
     FIFO.h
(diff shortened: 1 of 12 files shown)


JesseTG:
Actually, hold on. I've made another set of changes on a different branch that includes these commits. I will send in another PR shortly.

PR #1844, opened 2023-10-04
Refactor DSi_NAND

JesseTG:
This is another small prerequisite PR in my storage refactoring series.

This PR introduces the following small refactorings:

- The functionality of the `DSi_NAND` namespace is split into the `NANDImage` and `NANDMount` classes.
- `DSi_NAND::NANDImage` maintains the NAND image's `FileHandle*` and extracts important information that _doesn't_ need the file system to be mounted, such as the console ID or encryption keys. It also manages raw access to the image itself. When this object is deleted, the NAND image file is closed.
- `DSi_NAND::NANDMount` is used to access the actual file system within the NAND image. It mounts the NAND image with fatfs upon creation, and unmounts it upon destruction. It is created with a `NANDImage`, which must stay valid for the life of the `NANDMount`.
- The contents of `diskio.c` are moved out of fatfs and into the main `src` directory, since it consists of application-specific logic and declarations.
- The callbacks registered by `ff_disk_open` now accept `std::function`s instead of plain function pointers, so that they don't need to rely on a particular global `NANDMount`.

Once this PR is merged, next on my list will be to move the NAND's disk image loading logic from the core to the frontend.

### DIFF
--- a/src/DSi.cpp
+++ b/src/DSi.cpp
@@ -528,24 +528,22 @@ void SetupDirectBoot()
             ARM9Write32(0x02FFE000+i, tmp);
         }
 
-        if (DSi_NAND::Init(&DSi::ARM7iBIOS[0x8308]))
+        if (DSi_NAND::NANDMount nand = DSi_NAND::NANDMount(&DSi::ARM7iBIOS[0x8308]))
         {
             DSi_NAND::DSiFirmwareSystemSettings userdata {};
-            DSi_NAND::ReadUserData(userdata);
+            nand.ReadUserData(userdata);
             for (u32 i = 0; i < 0x128; i+=4)
                 ARM9Write32(0x02000400+i, *(u32*)&userdata.Bytes[0x88+i]);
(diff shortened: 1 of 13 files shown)


src/DSi.cpp
```diff
@@ -730,13 +772,51 @@ bool LoadNAND()
...
+                FileRead(tmpbuf, blklen, 1, orig);
+                FileWrite(tmpbuf, blklen, 1, nandfile);
+            }
+            delete[] tmpbuf;
+        }
+
+        Platform::CloseFile(orig);
+        Platform::CloseFile(nandfile);
+
+        nandfile = Platform::OpenLocalFile(instnand, FileMode::ReadWriteExisting);
```
RSDuck, line 808: what's the point of this file handle? It leaks and seems to be never used.

JesseTG: Oops. Good catch. That is indeed a mistake.

PR #1846, opened 2023-10-07
season2 test PR

Arisotura:
branch season2 is the testing ground for new fun networking features such as LAN, netplay, ...

netplay is currently broken, but LAN is mostly functional and testing is welcome


Xatrongamer:
Hey, I tested it for a bit and I immediately found that firmware language settings don't work properly. Is this a known bug? Is there any way to fix it?

PR #1852, opened 2023-10-13
Miscellaneous DSi NAND fixes

JesseTG:
This PR adds a few small fixes to the DSi NAND code that I'd been planning to introduce.

- Replaces some standard I/O calls (`fopen`, etc.) with `Platform::FileHandle` equivalents. I missed a spot when I submitted that PR a few months ago.
- Add some enums for representing regions and languages, and update the appropriate `struct` fields that use them. Upstream melonDS doesn't use these, but I intend to use them in melonDS DS. My use case is twofold:
  - I intend to stop the player from installing DSiWare games from incompatible regions
  - If the player's language is set to something that the current NAND doesn't support, then it will fall back to some other language. (Usually English, but not always.)
- Add an `#include` that should've been there.
- Move DSi NAND userdata patching to the frontend. Being able to control when and how this happens is important to me, because of some additional features I'm planning.

### DIFF
--- a/src/DSi.cpp
+++ b/src/DSi.cpp
@@ -945,7 +945,7 @@ bool LoadNAND()
         NDS::ARM7->JumpTo(bootparams[6]);
     }
 
-    nandmount.PatchUserData();
+    // user data is now expected to be patched by the frontend
 
     return true;
 }
--- a/src/DSi.cpp
+++ b/src/DSi.cpp
@@ -540,8 +540,9 @@ void SetupDirectBoot()
                     ARM9Write32(0x02000400+i, *(u32*)&userdata.Bytes[0x88+i]);
 
                 DSi_NAND::DSiSerialData hwinfoS {};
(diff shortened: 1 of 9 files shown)


Arisotura:
> I intend to stop the player from installing DSiWare games from incompatible regions

maybe don't make this mandatory? like, if they have a custom firmware installed

JesseTG:
> > I intend to stop the player from installing DSiWare games from incompatible regions
> 
> maybe don't make this mandatory? like, if they have a custom firmware installed

I wasn't aware that custom DSi firmware worked in melonDS? But this is a good idea, I'll make it optional at some point.

Arisotura:
> I wasn't aware that custom DSi firmware worked in melonDS? But this is a good idea, I'll make it optional at some point.

I don't know which ones work actually :P but it's still worth taking into account -- just make this a warning

otherwise, I can't really object against any of this PR, looks good to me

PR #1854, opened 2023-10-16
Support emulating R4 Revolution/M3DS Simply cartridges.

asiekierka:
This feature serves mostly the purpose of hardware preservation, with the R4 being a particularly notable aspect of that scene. The emulation is high-level and not the most accurate but, to my understanding, complete; it should be sufficient for all real-world use cases.

Notes:

- R4 emulation is currently enabled by using a dump of the "stage 1" loader.
  - It would be possible to remove this requirement (all the "stage 1" loader does is bootstrap an encrypted .nds file, with its decryption handled by the flashcart's hardware), but it would require some patchwork throughout the ROM loading system that I did not want to introduce as part of this PR.
- The emulated R4 uses the SD card image configured as DLDI.
- Conversely, the emulated R4 is not limited to DLDI emulation - official firmware should operate fully and correctly.
- It would be a good idea to expose configuring the R4's model (R4 Revolution or M3DS Simply) and language - the firmware actually checks this, so for now it is only possible to use English-language firmware designed for the R4. I'm not sure what would be a good place to do so, though, and likewise it could be done as a separate PR.
- The R4 decryption algorithm comes from a lineage of source code initially started by a Japanese developer nicknamed Yasu. While my implementation is expressed in a distinct manner and I do not believe the end result to fall under copyright derivation as such, I feel this lineage should be pointed out explicitly for whoever's reviewing and/or accepting this PR. With Yasu's website not having been updated in fourteen years, and the fact he had to deal with GBAtemp, inquiring about formal approval may be difficult. I hope to send an e-mail inquiry regardless, but I do not expect to receive a response.
  - Note also that devkitPro has bundled a yet distinct implementation of the algorithm since around 2013 for building nds-hb-menu.  

EDIT:

```
Remote-MTA: dns; gate06.netbees.clubbbq.com. (218.219.70.206, the server for
 the domain yasu.hot.co.jp.)
Diagnostic-Code: smtp; 553 sorry, that domain or user isn't in my list of allowed rcpthosts (#5.7.1)
Last-Attempt-Date: Mon, 16 Oct 2023 09:23:32 -0700 (PDT)
```

An attempt was made.

- The "stage 1" loader currently only works when using direct boot; for unknown reasons, the firmware fails to launch it correctly.


src/FATStorage.cpp
```diff
@@ -114,6 +156,10 @@ u32 FATStorage::WriteSectors(u32 start, u32 num, u8* data)
     return WriteSectorsInternal(File, FileSize, start, num, data);
 }
 
+u64 FATStorage::GetSectorCount(void) {
```
RSDuck, line 159: brace needs to be on the next line

src/NDSCartR4.cpp
```diff
@@ -0,0 +1,345 @@
...
+                SDFATEntrySectorGet(FATEntryOffset[rom ? 0 : 1], sector) >> 9,
+                1, Buffer
+            );
+        }
+    }
+}
+
+u32 CartR4::SDFATEntrySectorGet(u32 entry, u32 addr)
+{
+    u8 Buffer[512];
```
RSDuck, line 284: local variables should be in camelCase

RSDuck (commented):
only some stylistic remarks

src/FATStorage.cpp
asiekierka: Fixed!

src/NDSCartR4.cpp
asiekierka: Fixed!

RSDuck:
slightly evil merge, but can you please take a look at the merge? If you think everything is ok, I'm going to merge.

RSDuck:
ok well thanks.

PR #1855, opened 2023-10-18
Fix Certain Slopes Having Bugged Anti-Aliasing + Rewrite Of AA Code

Jaklyy:
This PR contains two commits:

Commit 1: Fixes a bug with certain positive x-major slopes having bugged aa.
This was caused by lack of proper handling for x coordinates which had a different y coordinate calculated than they actually had when following the DS' slope algorithm, resulting in improper coverage being calculated.
The fix was simply to invert the final coverage value when these two values differ.
This seems to match hardware behavior.
Also seems to allow for removing extra handling elsewhere as it should no longer be possible for abnormally high starting coverage values to be calculated.

Commit 2: Attempts to rewrite some of the anti-aliasing code to be simpler. Hope I succeeded, if not then it's easy to revert. (Though I'm keeping most of the changes to y-major aa, because in hindsight I wrote some needlessly gross code when I fixed aa for swapped slopes.)
I did some testing to ensure this new handling doesn't break anything, but it's possible I missed some weird edge case (there's a lot of them.)

For the most part anti-aliasing should be unchanged, aside from some small fixes on certain slopes.

Example bugged slope:
image
Fixed:
image

### DIFF
--- a/src/GPU3D_Soft.cpp
+++ b/src/GPU3D_Soft.cpp
@@ -1063,14 +1063,11 @@ void SoftRenderer::RenderPolygonScanline(RendererPolygon* rp, s32 y)
     if (x < 0) x = 0;
     s32 xlimit;
 
-    s32 xcov = 0;
-
     // part 1: left edge
     edge = yedge | 0x1;
     xlimit = xstart+l_edgelen;
     if (xlimit > xend+1) xlimit = xend+1;
     if (xlimit > 256) xlimit = 256;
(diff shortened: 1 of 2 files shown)


Jaklyy:
i forgot to post an explanation for why i marked this pr as draft:
it turned out that the nds anti-aliasing is *supposed* to do that sometimes, and this pr actually fixes instances that occur on real hardware.
and tbh, im probably just going to put the entire idea on the back burner until i feel ready to engage with rewriting the entire aa algorithm to match the current understanding of how hardware works, so i think im just going to close this pr.

PR #1856, opened 2023-10-18
Slight polish to DMA

JesseTG:
[Someone encountered a crash](https://github.com/JesseTG/melonds-ds/issues/87) in the melonDS DS libretro core, and I narrowed down the cause to an uninitialized variable in `DMA`. That crash occurs when auto-loading a savestate (which occurs within a frame or two of loading the ROM); since you can't manually load savestates that quickly in upstream melonDS, this issue hasn't had any impact there.

This PR makes some minor fixes to `DMA`, including:

- Default-initialize members explicitly
- Mark some methods as const noexcept
- Initialize DMA::MRAMBurstTable to DMATiming::MRAMDummy
- Use the default destructor

~~**Note:** I have not tested this PR yet. My current `HEAD` relies on [this unrelated PR](https://github.com/melonDS-emu/melonDS/pull/1852), so I can't test it with this one until #1852 is merged.~~ Tested it on a separate branch. Works like a charm, resolves the crash I identified.


src/NDS.cpp
```diff
@@ -907,7 +907,12 @@ bool DoSavestate(Savestate* file)
     }
 
     for (int i = 0; i < 8; i++)
-        DMAs[i]->DoSavestate(file);
+    {
+        if (!DMAs[i]->DoSavestate(file))
```
RSDuck, line 911: the DMA objects can't really be null except for OOM.

They could probably be packed directly as values instead of pointers.

JesseTG: How's this?

RSDuck: how's what?

JesseTG: The [two](https://github.com/melonDS-emu/melonDS/pull/1856/commits/66d9573608fa4ab2f3e5070688c622716e52133d) [commits](https://github.com/melonDS-emu/melonDS/pull/1856/commits/3ef38f4247f56e4d6f172b1bd87efe77ee539a50) that follow this one where I address this feedback.

PR #1864, opened 2023-10-24
Protect savestates while the threaded software renderer is running

JesseTG:
In upstream melonDS, the frontend has full control of when it takes savestates, so you can assume that the render thread won't try to run while a savestate is being taken or loaded.

I can't make that assumption, as libretro doesn't let cores control when savestates are taken. End result is [this issue](https://github.com/JesseTG/melonds-ds/issues/81), which this PR aims to fix.


JesseTG:
Okay, I think I've fixed the crash that I cited, _and_ without affecting performance.

Screenshot 2023-10-29 162921

That blue circle represents the only instance of lock contention in the trace file I took.

JesseTG:
I'm about to refactor `GPU` and `GPU3D` to remove global state, so this branch is probably gonna change soon.

JesseTG:
@RSDuck Would you mind taking a look at this PR when you have some time? I recall from our discussions on Discord that you have some reservations about it, but I'm not clear on what the issues are. This is probably one of the last PRs I need before I'm ready to release melonDS DS 1.0.

PR #1867, opened 2023-10-30
RTC revamp

Arisotura:
* rework the RTC to do its own counting instead of just returning the system time
* it's also configurable now (and you can also change it ingame, ie. from the firmware)
* add support for the RTC IRQ
* also add support for the keypad IRQ, and reworked sleep mode some

### DIFF
--- a/src/Platform.h
+++ b/src/Platform.h
@@ -337,6 +337,9 @@ void WriteGBASave(const u8* savedata, u32 savelen, u32 writeoffset, u32 writelen
 /// @param writelen The number of bytes that were written to firmware.
 void WriteFirmware(const SPI_Firmware::Firmware& firmware, u32 writeoffset, u32 writelen);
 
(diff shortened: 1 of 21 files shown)


src/Platform.h
```diff
@@ -337,6 +337,9 @@ void WriteGBASave(const u8* savedata, u32 savelen, u32 writeoffset, u32 writelen
 /// @param writelen The number of bytes that were written to firmware.
 void WriteFirmware(const SPI_Firmware::Firmware& firmware, u32 writeoffset, u32 writelen);
 
+// called when the RTC date/time is changed and the frontend might need to take it into account
+void WriteDateTime(int year, int month, int day, int hour, int minute, int second);
```
JesseTG, line 341: Does `month` start from 0 or 1?

Arisotura: month and day start from 1

PR #1873, opened 2023-11-09
Refactor the GPU to be object-oriented

JesseTG:
All GPU-related global state has been moved into a group of objects.


Wunkolo:
Thank you for this. I've been meaning to do this to add a vulkan renderer.

JesseTG:
> Thank you for this. I've been meaning to do this to add a vulkan renderer.

You're welcome. You might still have to do a bit of work to totally extract the OpenGL compositor from `GPU`, though.

PR #1877, opened 2023-11-12
DSi_SD: implement command 17, 24

asiekierka:
Read/write single block; this will be used by BlocksDS-based homebrew in the near future.

### DIFF
--- a/src/DSi_SD.cpp
+++ b/src/DSi_SD.cpp
@@ -951,6 +951,7 @@ void DSi_MMCStorage::SendCMD(u8 cmd, u32 param)
         Host->SendResponse(CSR, true);
         return;
 
+    case 17: // read single block
     case 18: // read multiple blocks
         //printf("READ_MULTIPLE_BLOCKS addr=%08X size=%08X\n", param, BlockSize);
         RWAddress = param;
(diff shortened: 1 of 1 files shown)


RSDuck:
is it right that to completely take over the behaviour of the block reads in regards to CSR and RWCommand?

asiekierka:
Not at all, but it did just so happen to work. Let's improve it...

asiekierka:
RWCommand was wrong, but I *think* CSR is actually fine; the data block is sent after a response for both single and multiple reads.

I'd like to test this further, but at some point after 0.9.5, my trick of using a dummy NAND filled with zeroes stopped working (I don't own a proper DSi, only a 3DS - I'd like to look into implementing some code to not require a BIOS/firmware/NAND dump for some forms of DSi boot like we do for DS eventually).

src/DSi_SD.cpp
```diff
@@ -973,7 +976,8 @@ void DSi_MMCStorage::SendCMD(u8 cmd, u32 param)
             RWAddress <<= 9;
             BlockSize = 512;
         }
-        RWCommand = 25;
+        if (cmd == 25)
+                RWCommand = 25;
```
RSDuck, line 980:
```suggestion
    RWCommand = 25;
```

src/DSi_SD.cpp
```diff
@@ -959,12 +960,14 @@ void DSi_MMCStorage::SendCMD(u8 cmd, u32 param)
             RWAddress <<= 9;
             BlockSize = 512;
         }
-        RWCommand = 18;
+        if (cmd == 18)
+                RWCommand = 18;
```
RSDuck, line 964:
```suggestion
    RWCommand = 18;
```

RSDuck (commented):
just the spacing

asiekierka:
Solved!

PR #1883, opened 2023-11-20
Completely remove all global state in the core

JesseTG:
This PR is an important step in the [ambitions plans](https://melonds.kuribo64.net/comments.php?id=191) for netplay. It aims to remove all global state in the core (with an exception or two involving the JIT).

The primary benefit is that more than one `NDS` can be created in memory, which is the first step to emulating local wireless over the public Internet. The code should also be more maintainable in the long term, as it's easier to reason about an object when you know what's allowed to change it. That means _no more global variables_, including `Platform::GetConfig`. Those calls will be removed in favor of some members that can be set.

There are also a few bugs that were introduced as part of the refactor of individual components; this PR aims to fix the known ones.

Once this PR is merged, work on new features may resume.


KostaSaizo7:
> With external DS firmware, if I uncheck the "boot game directly", the game icon in the menu is blank and when I press it, it crashes
> 

This is now fixed
image

JesseTG:
This PR has gotten overly complicated, so I'm going to close it and split the work into a few smaller PRs. Nothing will be lost or thrown out, it'll just be resubmitted in an orderly fashion that's less likely to introduce new bugs.

PR #1886, opened 2023-11-24
Move all core types into namespaces

JesseTG:
This PR moves all types in the core to the `melonDS` namespace, to reduce the risk of clashes in the global namespace in melonDS itself and any frontend that uses it as a dependency. It was originally part of [this PR](https://github.com/melonDS-emu/melonDS/pull/1883), but it got too unwieldy. So I'm splitting up into smaller PRs.

This PR looks enormous, but most of the changes are one-line `using` or `namespace` declarations.


src/ARM.cpp
```diff
@@ -104,7 +106,7 @@ const u32 ARM::ConditionTable[16] =
     0x0000  // NE
 };
 
-ARM::ARM(u32 num, ARMJIT::ARMJIT& jit, Melon::GPU& gpu) :
+ARM::ARM(u32 num, ARMJIT& jit, melonDS::GPU& gpu) :
```
RSDuck, line 109: why do you prefix the GPU?

RSDuck (commented):
I'm not happy with how the melonDS namespace creeps into the vendored dependencies, though I don't really know how to make it better either.

src/ARM.cpp
JesseTG: Because having a member with the same name as a type causes C++ to choke, and naming the type as `class GPU` looks weird.

JesseTG:
> I'm not happy with how the melonDS namespace creeps into the vendored dependencies, though I don't really know how to make it better either.

This was primarily to accommodate melonDS's integer `typedef`s. If you'd like, I can replace them (in the dependencies) with standard equivalents.

PR #1887, opened 2023-11-24
Split some types and declarations into their own header files

JesseTG:
This is to break `#include` cycles that crop up in commits I intend to cherry-pick from [this PR](https://github.com/melonDS-emu/melonDS/pull/1883).

### DIFF
--- a/src/Constants.h
+++ b/src/Constants.h
@@ -28,4 +28,27 @@ constexpr u32 ARM7WRAMSize = 0x10000;
 constexpr u32 DSiNWRAMSize = 0x40000;
 constexpr u32 ARM9BIOSLength = 0x1000;
 constexpr u32 ARM7BIOSLength = 0x4000;
+
+// matching NDMA modes for DSi
+constexpr u32 NDMAModes[] =
+{
+    // ARM9
+
+    0x10, // immediate
+    0x06, // VBlank
+    0x07, // HBlank
+    0x08, // scanline start
+    0x09, // mainmem FIFO
+    0x04, // DS cart slot
+    0xFF, // GBA cart slot
+    0x0A, // GX FIFO
+    0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF,
+
+    // ARM7
+
+    0x30, // immediate
+    0x26, // VBlank
+    0x24, // DS cart slot
+    0xFF, // wifi / GBA cart slot (TODO)
+};
 }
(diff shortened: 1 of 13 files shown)


src/Constants.h
```diff
@@ -0,0 +1,54 @@
...
+{
+constexpr u32 MainRAMMaxSize = 0x1000000;
+constexpr u32 SharedWRAMSize = 0x8000;
+constexpr u32 ARM7WRAMSize = 0x10000;
+constexpr u32 NWRAMSize = 0x40000;
+constexpr u32 ARM9BIOSSize = 0x1000;
+constexpr u32 ARM7BIOSSize = 0x4000;
+
+// matching NDMA modes for DSi
+constexpr u32 NDMAModes[] =
```
RSDuck, line 33: does this need to go into this file?

Otherwise I'd move it nearer to where it's used and rename this file to something more specifc than Constants.h

JesseTG: What, `NDMAModes`? That particular declaration was pulled from a later commit in #1883, and the intent was to resolve an `#include` cycle that cropped up. Right now `NDS.cpp` includes `DSi.h`, but `DSi.cpp` includes `NDS.h`; if I don't move that declaration here, then `DSi.cpp` may not have access to `NDS`'s complete declarations. That's not an issue now, but it will be once I cherry-pick the next few commits.

RSDuck: why did you move those out of the source file in the first place?

JesseTG: To break `#include` cycles. You can reject this PR if you'd like, but the changes will probably have to be introduced later.

RSDuck: https://github.com/search?q=repo%3AmelonDS-emu%2FmelonDS%20NDMAModes&type=code

can't you just leave the constant array in the cpp file where it came from? It doesn't need to be moved inside the class.

JesseTG: I'll see what I can do. If I can't escape doing that, I'll document why.

JesseTG:
This was one of the aforementioned `#include` cycles, by the way.
Screenshot 2023-11-25 134658

RSDuck:
ehm yeah I see, you moved the parts of NDS.cpp which used `NDMAModes` into DSi.cpp. So why not move `NDMAModes` with it?

JesseTG:
> ehm yeah I see, you moved the parts of NDS.cpp which used `NDMAModes` into DSi.cpp. So why not move `NDMAModes` with it?

I don't actually remember. I'm creating a new diff that's easier to review and merge; I'll see what I can do about this.

Issue #664, opened 2020-06-28
Minor software renderer bugs


Hydr8gon:
It seems I've fallen down the rabbit hole of 3D accuracy. During my testing, I found some issues with melonDS' software renderer:

![edge](https://user-images.githubusercontent.com/3057954/85939930-4999e500-b8e7-11ea-93f0-f89a0ddfd89c.png)
With edge marking and anti-aliasing enabled, perfectly horizontal edges are drawn with 100% opacity, instead of the 50% that they should be. All non-horizontal edges, including perfectly vertical edges, seem to be fine.

![line](https://user-images.githubusercontent.com/3057954/85939934-4ef72f80-b8e7-11ea-97ae-bce88d86b2ca.png)
Polygons perpendicular to the screen don't render properly. On the polygon's sides, it seems like it's drawing both sides next to each other? On hardware, one side is behind the other, and the rendered polygon is only one pixel wide. The results are the same regardless of anti-aliasing or edge marking.

![line-2](https://user-images.githubusercontent.com/3057954/85940257-5c151e00-b8e9-11ea-88e2-42b6ad1392df.png)
On the bottom of the polygon, it seems to be drawing from the red vertex to the blue vertex, when on hardware it draws from the red vertex to the green vertex. The result is a line half the length that it should be, with the wrong colors.

Here's the test program I made. It's just a modified version of that libnds example, but I added some features: hold A to zoom in, hold B to zoom out, hold L to disable anti-aliasing, hold R to enable edge marking, and hold X to enable wireframe.
[Simple_Tri.zip](https://github.com/Arisotura/melonDS/files/4841813/Simple_Tri.zip)

PR #1892, opened 2023-11-27
When the first vertex is clipped against the last vertex, make the new vertex the last vertex

Jaklyy:
Implements a theory on clipping I had.
Fixes horizontal line polygons rendering inaccurately when the first vertex is clipped.
Can't really give a solid reason *why* this is a thing. It just seems to be the simplest and least arbitrary explanation for horizontal line polygon behavior.
Don't ask why the last vertex being clipped against the first vertex doesn't make a new first vertex, it just doesn't seem to.
Maybe the clipping order just decides to clip vtx 1 against the final vertex last instead of first for some reason?
Hopefully the specifics don't matter!

Fixes the last bug mentioned in https://github.com/melonDS-emu/melonDS/issues/664
Fixes some tests in https://github.com/Jaklyy/polyrastertest

### DIFF
--- a/src/GPU3D.cpp
+++ b/src/GPU3D.cpp
@@ -769,6 +769,8 @@ int ClipAgainstPlane(const GPU3D& gpu, Vertex* vertices, int nverts, int clipsta
         temp[1] = vertices[1];
     }
 
+    Vertex holdit;
+    bool heldit = false;
     for (int i = clipstart; i < nverts; i++)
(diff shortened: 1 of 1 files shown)


Jaklyy:
this pr is definitely fixing it with the correct idea but the way it's going about it is probably wrong.
will do a deeper look into how clipping works eventually.

PR #1895, opened 2023-11-29
Clean up the 3D renderer for enhanced flexibility

JesseTG:
This PR introduces some minor cleanups involving `GPU3D` and the accompanying 3D renderers:

- Various renderer-related resources are kept within `std::unique_ptr` instead of plain pointers, reducing the risk of memory leaks.
- `GPU3D` no longer assumes that it's using any particular renderer. This ideally means that you can introduce new renderers without having to modify the rest of the core.
- Move the `GLCompositor` to be a member of `GLRenderer`, so that `GPU3D` doesn't need to care if it's using OpenGL.
- Remove the `RenderSettings` `struct` in favor of getters/setters for the members that they tweak.
- Make the 2D renderer easier to swap out.
- Add `constexpr` to a couple of places to facilitate compile-time optimizations where possible.

I've already tested this and it works like a charm, even when adjusting video settings at runtime.

### DIFF
--- a/src/GPU_OpenGL.cpp
+++ b/src/GPU_OpenGL.cpp
@@ -50,7 +50,7 @@ std::optional<GLCompositor> GLCompositor::New() noexcept
         // if linking the shaders together failed.
         return std::nullopt;
 
-    return std::make_optional<GLCompositor>(CompShader);
+    return { GLCompositor(CompShader) };
 }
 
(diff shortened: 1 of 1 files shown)


JesseTG:
Uh-oh. I broke the build again. I'll fix it.

JesseTG:
Okay, fixed.

PR #1896, opened 2023-11-30
Implementing the use of the touchscreen as camera movement

vitor251093:
Basically, this changes allows a user to use keys (or an analog joystick) to move the camera by using the touchscreen.

This method, ofc, is only useful for games which allow you to use the touchscreen to move the camera, such as Kingdom Hearts 358/2 Days.

There is still room for improvement. The movement could be smoother, and, when using an analog stick, the camera movement could move more or less depending on how tilted is the analog stick. Still, this implementation is already useful as it is.


src/SPI.cpp
```diff
@@ -462,6 +462,30 @@ void TSC::SetTouchCoords(u16 x, u16 y)
     NDS.KeyInput &= ~(1 << (16+6));
 }
 
+void TSC::MoveTouchCoords(u16 x, u16 y) // 0 -> negative, 1 -> neutral, 2 -> positive
```
JesseTG, line 465: Do you think you could make this an `enum` instead of a series of magic numbers? I could easily see some poor bastard trying to pass mouse coordinates in here, wondering why nothing works.

vitor251093: Sure. Any preferable place where I should place the enum?

JesseTG: `SPI.h` would be my first instinct.

vitor251093: What do you think @JesseTG ?

src/DSi_SPI_TSC.cpp
```diff
@@ -121,6 +121,11 @@ void DSi_TSC::SetTouchCoords(u16 x, u16 y)
     }
 }
 
+void DSi_TSC::MoveTouchCoords(SPITouchScreenMovement x, SPITouchScreenMovement y)
```
JesseTG, line 124: Is there a reason that `DSi_TSC`'s override of this method does nothing?

src/SPI.cpp
vitor251093: That's correct

src/DSi_SPI_TSC.cpp
vitor251093: Oh, sorry, I forgot to mention. I've only tested the touchscreen camera controls with the DS mode, but not with the DSi mode. Since the implementation of SetTouchCoords was different for each of them, I decided to include that override to avoid unexpected behaviours. At the time I had implemented that only for KH Days, so there was no use in implementing that function either.

I can make some tests with the DSi mode to implement that function, but that may take a little while.

vitor251093:
> why does this touch the core at all, looks to me like something which could be implemented on the frontend side only?

Because the input only allow me to update an existing coordinate within the touchscreen, requiring me to know the previous value, while TouchX and TouchY are protected values.

So I either needed to add a function to SPI which updates TouchX and TouchY based on the movement, or I would need to make TouchX and TouchY public.

CasualPokePlayer:
> With that, obtaining TouchX and TouchY would actually be a problem, so I would need to maintain two new variables at NDS.h/NDS.cpp, or main.cpp, instead.

It wouldn't really be that hard to just mask out stuff you don't need, a simple ~0x8000 and checking if the value is 0x700 (and probably checking if y == 0xFFF) would work (or maybe a getter would do that for you)

vitor251093:
> > or I would need to make TouchX and TouchY public (so I could extract that logic to NDS.cpp).
> 
> Is there a reason you can't add `GetTouchX()` and `GetTouchY()` to expose these members without letting them be modified arbitrarily?

I will try to do that, following @CasualPokePlayer 's suggestion. Seems be an approach will fit all the needs we talked about.

vitor251093:
Alright, just finished implementing the changes.

The SPIs now were only changed to add the getters, and the logic that converts inputs into touchscreen movements is now contained within `NDS::SetTouchKeyMask`.

Also now it can also detect every scenario when the coordinates becomes out of bounds, which is then considered as the touchscreen being released, so the movement can restart on the next frame starting from the middle of the touchscreen.

vitor251093:
I just realized there was a conflict. Fixed it

PR #1898, opened 2023-12-01
Refactor how save data (including SD cards) is initialized

JesseTG:
This PR makes some changes to most processes that involve saving and loading data (not just SRAM). Highlights include:

- Add a `struct` that can be used to configure an `NDS` or `DSi` with the required system files. I intend to add other non-sysfile parameters (e.g. audio interpolation) in the future.
- Use the aforementioned `NDSArgs`/`DSiArgs` to initialize the ROMs, save data, firmware, SD cards, and NAND image within the `NDS`/`DSi` constructors as warranted
- Clean up `FATStorage` to make it rely more on RAII
- Change the names of some related methods for consistency (e.g. from `Load` or `Insert` to `Set`)
- Move the DSi's SD card and NAND into `DSi_SDMMCStorage` and give it ownership
- Return the ejected cart in `NDS::EjectCart` and associated methods
- Allow the DSi SD card and NAND to be set through `DSi` itself
- Fix a bug in `NANDImage`'s move assignment operator
- Don't reload the SD card when resetting the game; now the frontend needs to do that itself. (This goes for the DSi and homebrew SD cards.)
- Allow homebrew SD cards to be set at runtime. (You'll need to cast the `CartCommon` pointer returned from `NDSCartSlot::GetCart`.)
- Remove `Platform::ConfigEntry`'s associated with SD cards. The `Config` variables themselves are untouched.
- Store `DSi_SDDevice` and cart ROM/SRAM data behind `unique_ptr`s instead of raw pointers, to reduce the risk of memory leaks or double-frees.
- Embed SD cards, firmware, and NAND images as objects directly to reduce pointer-chasing and the risk of memory leaks or double-frees. (These types are all move-constructible and move-assignable.)
- Allow `NDSCart`s and `GBACart`s to be initialized with save data or SD cards as appropriate. (They can still be set at runtime later.)
- Add overloads to `ParseROM` that accept movable references to ROM data and SRAM, so that the buffer doesn't need to be copied.
- Use `std::unique_ptr` more liberally in the parts of the frontend that load ROMs and SRAM, to reduce the risk of memory leaks and double-frees.
- Remove a considerable amount of redundant code in `ROMManager`.
- Centralize the init/update of the `NDS`/`DSi` in `EmuThread`, so that the console is recreated only if necessary. Currently, an existing console is only thrown out if switching between NDS mode and DSi mode, although all settings that require a reboot still force one.


JesseTG:
Hold on, I broke firmware saving.

PR #1903, opened 2023-12-05
Fix `SPU` not compiling in environments without a `constexpr` `std::cos`

JesseTG:
It turns out that `std::cos` isn't `constexpr` like I thought, [and won't be until C++26](https://en.cppreference.com/w/cpp/numeric/math/cos). So I took a `constexpr`-friendly cosine implementation from a blog and used that.


RSDuck:
I hate C++.

JesseTG:
You and me both.

PR #1907, opened 2023-12-06
Exclude JIT-related declarations more aggressively

JesseTG:
This reduces the risk of linker errors occurring when building on ARM without the JIT.

### DIFF
--- a/src/ARM.cpp
+++ b/src/ARM.cpp
@@ -17,6 +17,7 @@
 */
 
 #include <stdio.h>
+#include <assert.h>
 #include <algorithm>
 #include "NDS.h"
 #include "DSi.h"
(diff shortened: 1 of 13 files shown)


Adamillo:
That makes me curious. Does this mean that someone could now theoretically use melonDS on ARMv7 aka 32-bit ARM with no JIT?

Adamillo:
I see. The reason I ask is that on platforms like x86_32 you can use melonDS with the JIT disabled. I'd imagine that you'd be able to do the same with ARM32 now that the linker errors are fixed

RSDuck:
> That makes me curious. Does this mean that someone could now theoretically use melonDS on ARMv7 aka 32-bit ARM with no JIT?

you could already do that before. Besides maybe some small slip up the JIT is fully disableable

JesseTG:
And this PR fixes some sources of small slip-ups.

PR #1909, opened 2023-12-07
Sprinkle `const` around where appropriate

JesseTG:
- This will make it easier to use `NDS` objects in `const` contexts (e.g. `const` parameters or methods)

### DIFF
--- a/src/DSi_DSP.h
+++ b/src/DSi_DSP.h
@@ -41,7 +41,7 @@ public:
     void DSPCatchUpU32(u32 _);
 
     // SCFG_RST bit0
-    bool IsRstReleased();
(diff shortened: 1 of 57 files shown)


src/DSi_DSP.h
```diff
@@ -64,7 +64,7 @@ class DSi_DSP
     void IrqRep1();
     void IrqRep2();
     void IrqSem();
-    u16 DSPRead16(u32 addr);
+    u16 DSPRead16(u32 addr) const;
```
RSDuck, line 67: I'm not familiar with the DSP address space, but MMIO where reading triggers an action is a thing.

PR #1916, opened 2023-12-15
Encapsulate both network implementations into objects

JesseTG:
This PR wraps `LAN_Socket` and `LAN_PCap` in an object-oriented interface to simplify management of their resources. Instead of having to ensure that frontends call `Init()` and `DeInit()` in the right order, they just have to hang on to their `LAN_Socket` or `LAN_PCap` objects until they're done.

Although these classes are part of the frontend, [I build them into melonDS DS](https://github.com/JesseTG/melonds-ds/blob/main/src/libretro/CMakeLists.txt#L135-L148). [Rafael does so](https://github.com/rafaelvcaetano/melonDS-android-lib/blob/a26a30f19a1d0317af82ebf8074740a61cf3584b/src/android/CMakeLists.txt#L5-L6) in his Android port, too.

### DIFF
--- a/src/frontend/qt_sdl/LAN_PCap.cpp
+++ b/src/frontend/qt_sdl/LAN_PCap.cpp
@@ -387,7 +387,7 @@ LAN_PCap::LAN_PCap(const PCap& pcap, const AdapterData& data, pcap_t* adapter) n
 
 LAN_PCap::~LAN_PCap() noexcept
 {
-    if (PCapAdapter)
+    if (PCapAdapter && pcap_close)
     {
         pcap_close(PCapAdapter);
         PCapAdapter = nullptr;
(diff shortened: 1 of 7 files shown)


RSDuck:
I'm not really sold on managing global resources in objects.

JesseTG:
The only thing that's intrinsically global is the PCap DLL, which is loaded at runtime. Even then, you can still have multiple handles to a DLL; the OS just refcounts them.

JesseTG:
> I'm not really sold on managing global resources in objects.

Thinking about this position some more, I would argue that the network implementations _shouldn't_ be global state, at least not in my case. Since I (and a few other frontends) load melonDS through a dynamic library, any network implementation state is only global to the extent that the DLL itself is.

melonDS DS, like all libretro cores, is ostensibly loaded and unloaded as players decide to start and stop games. However, it's not safe to assume that global destructors will be called when the DLL is unloaded. It's not even safe to assume that constructors will be called a second time when the DLL is loaded again, because it may not have actually been unloaded to begin with. Improperly-destroyed global state is a common source of crashes with libretro cores.

This isn't a big concern for standalone melonDS, as global variables are either cleaned up or made irrelevant when the app exits. The same is not true in my case.

PR #1936, opened 2024-01-01
Add support for generating frame dumps

Jaklyy:
code is _very_ low quality right now
but given it's \*mostly* functional, I wanted to publish it

Frame dumps can be generated under the "System" tab and can be run via this rom:
https://github.com/Jaklyy/ndsframereplay


src/GPU3D.cpp
```diff
@@ -546,7 +550,331 @@ void GPU3D::SetEnabled(bool geometry, bool rendering) noexcept
...
+    
+    // save these values here because its way easier this way.
+
+    // i hate this
+    u32 tocopy = ((u64)&FDT.SwapBuffer + sizeof(FDT.SwapBuffer)) - (u64)&FDT.ZDotDisp_Track;
+    printf("%li", tocopy);
+    memcpy(&FDIS, &FDT, tocopy);
+    
+    // i hate this even more
+    memcpy(FDIS.NormLightVec_Track, FDT.NormLightVec_Track, sizeof(FDT.NormLightVec_Track));
```
RSDuck, line 573: you can probably pack things into structs and then just use assignment

src/GPU3D.cpp
```diff
@@ -2903,6 +3332,7 @@ void GPU3D::Write32(u32 addr, u32 val) noexcept
     if (addr >= 0x04000400 && addr < 0x04000440)
     {
         WriteToGXFIFO(val);
+        NewWriteFD(addr, val);
```
RSDuck, line 3335: I'm not sure about this, because it would change the order a bit. But it might be slightly faster to only store commands batched in GPU3D::Run()

Jaklyy: i dont *think* that would change the order? it might also be easier to work with for some more complex features im planning on.
would still require tracking zero dot disp register writes separately though, but i think the order would still work out fine(?)
also would require saving command data to file as 16 bit "id' (256 gx cmds, +1 extra), 32 bit "write" (and probably some manual batching of cmds?)
could get away with still using 8 bits for id if i either redirect one of the nops to another cmd, or repurpose the box/vector tests ids (since i dont think those do anything relevant for this purpose).
though this whole system for logging and storing writes goes out the window if i implement support for 2d engine gfx

src/GPU3D.cpp
```diff
@@ -546,7 +550,171 @@ void GPU3D::SetEnabled(bool geometry, bool rendering) noexcept
...
+    FD->Cmd[FD->NumCmds++] = cmd;
+
+    for (int i = 0; i < CmdNumParams[cmd]; i++)
+        FD->Params[FD->NumParams++] = param[i];
+}
+
+void GPU3D::StartFrameDump()
+{
+    NDS.GPU.QueueFrameDump = false;
+    FD = std::make_unique<FrameDump>();
```
JesseTG, line 567: If `make_unique` fails due to running out of memory then it won't return `nullptr`. It'll throw an exception. If we're out of memory there's not much to do besides crash anyway.

src/GPU3D.h
```diff
@@ -80,6 +81,58 @@ struct Polygon
 
 };
 
+struct FrameDump
```
JesseTG, line 84: Are you _only_ dumping the 3D renderer's state? Or do you intend to include the 2D renderers as well?

RSDuck (commented):
it would probably be a good idea to move the frame capture data and related functions into a separate object in a separate file.

src/GPU3D.cpp
```diff
@@ -546,7 +550,171 @@ void GPU3D::SetEnabled(bool geometry, bool rendering) noexcept
...
+    FD->ClearDepOff = RenderClearAttr2;
+    FD->FogColor = RenderFogColor;
+    FD->FogOffset = RenderFogOffset;
+    for (int i = 0; i < 32; i++)
+    {
+        FD->FogTable[i] = RenderFogDensityTable[i+1];
+        FD->ToonTable[i] = RenderToonTable[i];
+    }
+
+    Platform::MakeLocalDirectory("framedumps");
```
JesseTG, line 637: Instead of hardcoding a name, I'd make this method take the name as an argument. That goes for both the directory and the file itself.

src/GPU3D.cpp
Jaklyy: Is there a way to handle this without crashing or do we just accept our fate?

src/GPU3D.h
Jaklyy: Currently just 3d renderer, i do intend to leave room for the 2d renderer to be included in the future, but as of now, no

src/GPU3D.cpp
```diff
@@ -1721,14 +1889,18 @@ void GPU3D::ExecuteCommand() noexcept
 
         /*printf("[GXS:%08X] 0x%02X,  0x%08X", GXStat, entry.Command, entry.Param);*/
 
+
+
         switch (entry.Command)
         {
         case 0x10: // matrix mode
+            NewWriteFD(entry.Command, &entry.Param);
```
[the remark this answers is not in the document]
Jaklyy: I dont do it for every command, i exclude nops, and box and vector tests.
Should probably move it to before anyway, and filter them out before writing to file. if just because i might reuse the system for a 3d debugger later on

src/GPU3D.cpp
Jaklyy: Good call, though I'll first need to figure out how to properly handle opening/browsing arbitrarily named files in the file browser of the frame dump player

src/frontend/qt_sdl/CMakeLists.txt
```diff
@@ -40,6 +40,7 @@ set(SOURCES_QT_SDL
     SaveManager.cpp
     CameraManager.cpp
     AudioInOut.cpp
+    ../../FrameDump.cpp
```
JesseTG, line 43: Refresh my memory; what difficulty were you having compiling this file into the `core` target again?

src/frontend/qt_sdl/Platform.cpp
```diff
@@ -271,6 +271,23 @@ FileHandle* OpenFile(const std::string& path, FileMode mode)
     }
 }
 
+void MakeLocalDirectory(const std::string& dir)
+{
```
JesseTG, line 275: To match the behavior of `OpenLocalFile`, this should use absolute paths as-is.

src/FrameDump.h
```diff
@@ -0,0 +1,77 @@
...
+#ifndef FRAMEDUMP_H
+#define FRAMEDUMP_H
+
+#include <vector>
+#include "GPU.h"
+
+namespace melonDS
+{
+
+class FrameDump
```
JesseTG, line 28: This is intended for a homebrew test ROM you wrote, right? I'd add a link to it in the comments for the benefit of future readers.

src/FrameDump.cpp
```diff
@@ -0,0 +1,209 @@
...
+    SwapBuffer = GPU.GPU3D.FlushAttributes;
+
+    // they should already be empty, but just in case.
+    Cmds.clear();
+    Params.clear();
+}
+
+bool FrameDump::FinFrameDump()
+{
+    std::string filename = ROMManager::GetFrameDumpName();
```
JesseTG, line 100: This name should be passed in as an argument. The `FrameDump` doesn't need to know anything about the frontend, it just needs to be told where to write the file.

src/FrameDump.cpp
```diff
@@ -0,0 +1,209 @@
...
+
+    // save final state vars to file
+    Platform::FileWrite(&GPU.GPU3D.RenderDispCnt, 1, 2, file); // only write two bytes, (only 16 bits are used, and thus only those are stored for frame dumps)
+    Platform::FileWrite(GPU.GPU3D.RenderEdgeTable, 1, sizeof(GPU.GPU3D.RenderEdgeTable), file);
+    Platform::FileWrite(&GPU.GPU3D.RenderAlphaRef, 1, sizeof(GPU.GPU3D.RenderAlphaRef), file);
+    Platform::FileWrite(&GPU.GPU3D.RenderClearAttr1, 1, sizeof(GPU.GPU3D.RenderClearAttr1), file);
+    Platform::FileWrite(&GPU.GPU3D.RenderClearAttr2, 1, sizeof(GPU.GPU3D.RenderClearAttr2), file);
+    Platform::FileWrite(&GPU.GPU3D.RenderFogColor, 1, sizeof(GPU.GPU3D.RenderFogColor), file);
+    Platform::FileWrite(&GPU.GPU3D.RenderFogOffset, 1, 2, file); // only write two bytes
+    Platform::FileWrite(&GPU.GPU3D.RenderFogDensityTable[1], 1, sizeof(GPU.GPU3D.RenderFogDensityTable[0])*32, file); // only write the 32 "real" entries of the density table
```
JesseTG, line 113: Did you mean to start from `RenderFogDensityTable[0]`?

src/FrameDump.cpp
```diff
@@ -0,0 +1,209 @@
...
+    // only save vram if the bank is enabled and allocated to be used as texture image/texture palette.
+    if ((GPU.VRAMCNT[0] & 0x83) == 0x83) Platform::FileWrite(GPU.VRAM_A, 1, sizeof(GPU.VRAM_A), file);
+    if ((GPU.VRAMCNT[1] & 0x83) == 0x83) Platform::FileWrite(GPU.VRAM_B, 1, sizeof(GPU.VRAM_B), file);
+    if ((GPU.VRAMCNT[2] & 0x87) == 0x83) Platform::FileWrite(GPU.VRAM_C, 1, sizeof(GPU.VRAM_C), file);
+    if ((GPU.VRAMCNT[3] & 0x87) == 0x83) Platform::FileWrite(GPU.VRAM_D, 1, sizeof(GPU.VRAM_D), file);
+    if ((GPU.VRAMCNT[4] & 0x87) == 0x83) Platform::FileWrite(GPU.VRAM_E, 1, sizeof(GPU.VRAM_E), file);
+    if ((GPU.VRAMCNT[5] & 0x87) == 0x83) Platform::FileWrite(GPU.VRAM_F, 1, sizeof(GPU.VRAM_F), file);
+    if ((GPU.VRAMCNT[6] & 0x87) == 0x83) Platform::FileWrite(GPU.VRAM_G, 1, sizeof(GPU.VRAM_G), file);
+    
+    u8* tempcmds = (u8*)malloc(Cmds.size() * sizeof(u8));
```
JesseTG, line 158: Personally, I'd use a `std::vector` here instead of `malloc`, possibly initialized to a known size.

src/GPU3D.cpp
```diff
@@ -569,7 +533,17 @@ void GPU3D::SetEnabled(bool geometry, bool rendering) noexcept
...
 
-
+inline void GPU3D::BeginWriteFD(u16 cmd, u32* param)
+{
+    if (NDS.GPU.FD != nullptr)
+    {
+        if (NDS.GPU.FD->FDWrite(cmd, param))
+        {
+            NDS.GPU.FD = nullptr;
+            // add an osd notif here?
```
JesseTG, line 543: I think the frontend would be a better place for an OSD notification. The frontend is calling this method in the first place, right?

src/frontend/qt_sdl/CMakeLists.txt
Jaklyy: results in frontend files (specifically ROMManager.h) that were included in framedump.cpp being unable to find their includes
need a better solution to this problem

src/FrameDump.cpp
Jaklyy: nope, entries 0 and 33 are duplicates of 1 and 32

src/frontend/qt_sdl/Platform.cpp
Jaklyy: good catch

src/FrameDump.cpp
```diff
@@ -0,0 +1,209 @@
...
+    }
+    Platform::FileWrite(&numcmd, 1, sizeof(numcmd), file);
+    Platform::FileWrite(&numparam, 1, sizeof(numparam), file);
+    Platform::FileWrite(tempcmds, 1, numcmd, file);
+    Platform::FileWrite(tempparams, sizeof(Params[0]), numparam, file);
+    free(tempcmds);
+    free(tempparams);
+    Platform::CloseFile(file);
+
+    return false;
```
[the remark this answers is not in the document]
Jaklyy: fair enough

src/FrameDump.cpp
Jaklyy: i figured malloc would work just fine for this situation, since i don't need to do any complex management of the memory, so the potential for a memory leak is low, and the upper bounds of how much memory i'd need in this situation are known, so the ability to dynamically resize seemed unneeded.
Though a dynamically sized array would be slightly more efficient for memory usage since im not guaranteed to be using the full size of the allocated memory, but i figured that hyper optimizing for that didn't really matter.
I can change it if you feel it'd be better though.

src/FrameDump.cpp
JesseTG: I'd note that fact in a comment, then, just to ensure that someone trying to be helpful in the future doesn't undo it. My rule of thumb for comments; if it looks weird, comment it. Starting anything from index 1 looks weird.

src/GPU3D.cpp
JesseTG: Probably just have to accept your fate. On paper, you could handle it with a `try`/`catch` statement like any other exception. But in practice, I'm not sure. Here's why:

Some operating systems (such as Linux) over-commit memory, i.e. they may give you an address space but not allocate memory until you try to use it. A consequence of this is that it's possible to get a technically-valid pointer that causes a crash when you try to use it. In that case, the OS lied to you about the memory it could offer.

`bad_alloc` can supposedly be handled like an ordinary exception, but I don't know how well it plays with over-committed memory. There's probably a trick you could do with [the implicit second template argument](https://en.cppreference.com/w/cpp/memory/unique_ptr), but I don't think it's worth it in this case.

RSDuck: we don't really handle failed allocations anywhere except for the rom in memory. I don't think it's a state worth recovering from for most applications (besides saving work if  possible).

scurest:
I'm interested in this.

AFAICT there is no magic number or version number in the ndsfd file? I think those would be good to have.

A magic number would also be good because the [PNG spec](https://w3c.github.io/PNG-spec/#sec-defining-private-chunks) wants private chunks to have "additional identifying information [...] stored at the beginning of the chunk data to further reduce the risk of conflicts."

Also shouldn't the last byte in the PNG chunk type `ndSF` be lowercase, ie it should be safe to copy? Similar to `tEXt`?

Jaklyy:
Y'know its funny you mention that because I legit added a version number and magic number to the frame dump header just today, I just haven't pushed it yet.
also I guess I misread the "safe to copy" flag as _uppercase_ last letter-- good catch. will fix

Jaklyy:
this pr is just broken now and needs to be completely rewritten from scratch most likely.

PR #1937, opened 2024-01-02
Reset the JIT when loading savestate

warmenhoven:
The effect of this change is simply to call JitEnableWrite(), which is necessary on apple silicon


RSDuck:
hm looks like the call of ResetBlockCache in DoSavestate is older than the ARMJIT::Reset function.

JesseTG:
> hm looks like the call of ResetBlockCache in DoSavestate is older than the ARMJIT::Reset function.

What does that imply?

RSDuck:
that is the reason DoSavestate didn't use Reset before the PR.

PR #1947, opened 2024-01-09
Add interrupt instruction for games to print to terminal

abaresk:
If the game executes `svc 0xFC`, the contents of the null-terminated string addressed in register `r0` will be printed to the terminal.

For consistency, this is the same mechanism that is used in Desmume ([source](https://github.com/TASEmulators/desmume/blob/b06537cf51b5ad5dae70378e9fb995b9a4dcafc0/desmume/src/thumb_instructions.cpp#L1018)).

### DIFF
--- a/src/ARMInterpreter.cpp
+++ b/src/ARMInterpreter.cpp
@@ -278,6 +278,20 @@ void A_SVC(ARM* cpu)
 
 void T_SVC(ARM* cpu)
 {
+    // Print from game. Execute `svc 0xFC` with the null-terminated string address in `r0`.
+    if ((cpu->CurInstr & 0xFF) == 0xFC)
+    {
+        u32 addr = cpu->R[0];
+        for (;;)
+        {
+            u32 c;
+            cpu->DataRead8(addr++, &c);
+            if (!c) break;
+            printf("%c", c);
+        }
+        return;
+    }
+
     u32 oldcpsr = cpu->CPSR;
     cpu->CPSR &= ~0xBF;
     cpu->CPSR |= 0x93;


src/ARMInterpreter.cpp
```diff
@@ -278,6 +278,20 @@ void A_SVC(ARM* cpu)
...
+    // Print from game. Execute `svc 0xFC` with the null-terminated string address in `r0`.
+    if ((cpu->CurInstr & 0xFF) == 0xFC)
+    {
+        u32 addr = cpu->R[0];
+        for (;;)
+        {
+            u32 c;
+            cpu->DataRead8(addr++, &c);
+            if (!c) break;
+            printf("%c", c);
```
JesseTG, line 290: Please use `Platform::Log` instead of `printf`. Then the output can go wherever the frontend wants it to instead of always `stdout`.

abaresk: done

JesseTG: Great, thank you.

windwakr:
Is there actually any homebrew out there that uses this over nocash prints?

abaresk:
Just following up, are there any concerns with merging in this feature?

abaresk:
Sorry to keep bugging, is it okay to merge this?

Gericom:
I would suggest that if this feature will be added it should be an opt-in feature to avoid clashing with possible other uses of this swi. E.g. someone can map vectors to itcm and have their own swi handler. It's rare, but possible (for example in gbarunner3). Personally I think the existing way to print with no$ IO registers is less intrusive in that regard.

src/ARMInterpreter.cpp
```diff
@@ -266,6 +266,22 @@ void A_MRC(ARM* cpu)
 
 void A_SVC(ARM* cpu)
 {
+    // Print from game. Execute `svc 0xFC` with the null-terminated string address in `r0`.
+    if ((cpu->CurInstr & 0xFF) == 0xFC && cpu->NDS.GetDebugPrint())
```
abaresk, line 270: I tried `((cpu->CurInstr >> 16) & 0xFF)` as suggested, but this didn't work in practice. The interrupt instruction (`swi 0xFC`) was compiled to the following:

```
   8:   ef0000fc        svc     0x000000fc
```

GalaxyShard:
I fixed up the merge conflicts for my own use; I'm not sure whether or not to make a seperate PR but feel free to take the code: https://github.com/GalaxyShard/melonDS/tree/svc-print

I also changed the 0xFC to 0xFC0000 for arm, consistent with desmume's apparent behavior (and I may change the symbols named "DebugPrint" to something more descriptive).

abaresk:
Merge conflicts have been addressed. I also added some notes on this feature in the 'Emu settings' dialog.

Issue #380, opened 2019-02-16
Game crashes after second cutscene in The World Ends With You


5kyl4rk:
When playing the game for the first time, you have to do a quick tutorial fight, after which a cutscene will play.  A little while after it plays the game will crash and the music will get stuck.  Before it does crash, it appears that the top and bottom screens are slightly off sync (if that helps).  The only way around it is to import a save from either DraStic or DeSmuME from after that cutscene.

Ammako:
Oh wow, how did I not see this issue sooner?

Supposedly this game spams the B8 command as some weird form of anti-piracy, it caused game crashes on older flashcarts or those with slow enough SD cards. There was a recent-ish commit which made fixes to cart timings, and it may have fixed this bug (it's not too unreasonable to assume that this weird from of AP could also trip up emulators which don't properly emulate this.)

If you can compile melonDS yourself you may be able to try it out and see if it works. Otherwise I might still find the time to set up the build environment for it on Windows so I can do it myself, but setting up Code Blocks and SDL2 for compiling melonDS has been a royal pain in the past, so I'm not looking forward to having to do that.

Arisotura:
command B8 is for getting the cart's chip ID. games typically check it against the value stored in RAM every once in a while, to check whether the cart was ejected.

I don't quite see how that would break TWEWY unless it expects a specific chip ID value.

Ammako:
As far as flashcarts go, supposedly B8 is used to stop SD traffic to save battery life, but causes the SD to lockup if spammed too much, and I guess faster SD cards are able to handle it without locking up.

https://gbatemp.net/threads/the-world-ends-with-you-u-crash-in-beginning-cutscene.222788/#post-2780571

As far as emulator goes, who knows. This is assuming that this is even the same thing that causes the crash in melonDS.

Arisotura:
have a savefile close to the crashing spot?

Ammako:
It happens before you get to save for the first time. It only takes a few minutes to get there though (unless you want a save state, I can do that.)

Arisotura:
savestate w/ latest melonDS would be cool, yea

Ammako:
[twewy.zip](https://github.com/Arisotura/melonDS/files/3124262/twewy.zip)

Arisotura:
thx

Arisotura:
loading that savestate causes a flood of 'unknown ARM9 write 00000xxx', do you also get those?

Ammako:
[...]

Issue #359, opened 2019-01-07
Sonic Chronicles - The Dark Brotherhood (Europe) freezes up on the first stage


benderscruffy:
using 0.7.3
at the start of the first stage the game freezes up

benderscruffy:
using 0.8.1
still happens

Arisotura:
v0rp

Arisotura:
going to test it out. if it's not another timing issue, it's going to be interesting.

Arisotura:
welp, that's what I feared. it works on old versions (0.7 and older). it's a big fat stinky timing issue.

Arisotura:
looking back on it (also, related thread: http://melonds.kuribo64.net/board/thread.php?pid=2010#2010)

it actually started breaking in 0.7.3, not 0.7.1, so that's interesting

Arisotura:
as I feared: this is broken by commit 669247e8c81f1480dea66fab8d9d0b0f9413a6ea

the one about using absolute timestamps

gonna try and figure out why.

RSDuck:
this is likely a result of inaccurate arm9 timing

works with the arm9timing hack branch

atomicwrites:
There is something problematic about the tutorials in this game, which seem to be pre-rendered videos. I was originally using DeSmuMe and it would crash in the POW actions tutorial, about 5 minutes into the game, then tried mellonDS and it crashes on the motion tutorial which is the first thing that happens. I built the `arm9timinghack` branch and it now got past the first few tutorials but crashes on the POW tutorial like DeSmuMe. I dont really see any info about it, what does `arm9timinghack` do? When it crashes, it starts spaming the console with 

```
unknown arm9 read32 E597C038 | E597C03C 00000000
unknown arm9 read32 E597C03C | E597C03C 00000000
unknown arm9 read32 E597C040 | E597C040 00000000
unknown arm9 read32 E597C044 | E597C044 00000000
unknown arm9 read32 E597C048 | E597C048 00000000
unknown arm9 read32 E597C04C | E597C04C 00000000
unknown arm9 read32 E597C050 | E597C050 00000000
unknown arm9 read32 E597C054 | E597C054 00000000
unknown arm9 read32 E597C058 | E597C058 00000000
unknown arm9 read32 E597C05C | E597C05C 00000000
unknown arm9 read32 E597C060 | E597C060 00000000
unknown arm9 read32 E597C064 | E597C064 00000000
[...]

PR #1955, opened 2024-01-18
CP15 Cache Emulation Work

DesperateProgrammer:
The instruction and data cache register were not yet implemented in MelonDS and would act different against the hardware by just reading/writing them.

They have no impact on the - not yet implemented - data cache nor on the instruction cache until their implementation progresses.


src/ARM.h
```diff
@@ -322,6 +322,7 @@ class ARMv5 : public ARM
     u32 RNGSeed;
 
     u32 DTCMSetting, ITCMSetting;
+    u32 DCacheLockDown, ICacheLockDown;
```
JesseTG, line 325: Don't forget to add these to the savestate (and to increment the savestate version).

src/CP15.cpp
JesseTG, line 1: Thank you. Could you please increment the major version number in `Savestate.h` as well? It's our only way to detect savestate format changes at runtime; if it's not updated, then loading old states may crash (or worse, silently break the emulated console).

DesperateProgrammer:
I'll add some more changes to the CP15 Caching Instructions and Handling here.

DesperateProgrammer:
The last commit was a bit bigger / combined, as adding DCache required other things to not break some applications.

Instruction cache is working except:
 - Prefetch via CP15 (todo)
 - CacheLockdown has no effect yet
 - Timings on cache-fill are not correct yet
Data cache is working except:
- Timings on cache fill are not correct yet
- Timings on Cache hit are not correct yet
- CacheLockdown has no effect yet
- WriteBack is not implemented yet (it writes through and invalidates the cache line atm)

CP15 Registers Operations:
- Cache Clean has no effect yet (needs write back mechanism)
- Cache Clean & Invalidate will now also invalidate the cache (was missing)
- Cache Debug Index Register was added
- ICache/DCache Tag Registers were added
- ICache/DCache Data Registers are not done yet
- Privileged/User Checks are missing

DesperateProgrammer:
Update on the status

Instruction cache is working except:
-     Prefetch via CP15 (todo)
-     Timings on cache-fill are not correct yet

Data cache is working except:
-     Timings on cache fill are not correct yet
-     WriteBack is not implemented yet (it writes through)

CP15 Registers Operations:
-     Cache Clean has no effect yet (needs write back mechanism)
-     Cache Clean & Invalidate will now also invalidate the cache (was missing)
-     Cache Debug Index Register is working
-     ICache/DCache Tag Registers are working
-     ICache/DCache Data Registers are working
-     Privileged/User Checks are missing

JIT:
- Caches are now disabled when JIT is enabled and no longer breaks JIT.

DesperateProgrammer:
This turned out to be quite some rework for the CP15.
The Cache is now working including the write-back mode. Cache Debug registers, BIST Cache Test register
CP15 Registers and Functions received some documenting comments.
Many CP15 related calls received the const-specifier for its arguments when they are not changed in the call.

src/CP15.cpp
```diff
@@ -332,153 +366,596 @@ u32 ARMv5::RandomLineIndex()
...
+u32 ARMv5::ICacheLookup(const u32 addr)
 {
-    u32 tag = addr & 0xFFFFF800;
-    u32 id = (addr >> 5) & 0x3F;
+    const u32 tag = (addr & ~(ICACHE_LINELENGTH - 1));
+    const u32 id = ((addr >> ICACHE_LINELENGTH_LOG2) & (ICACHE_LINESPERSET-1)) << ICACHE_SETS_LOG2;
 
-    id <<= 2;
-    if (ICacheTags[id+0] == tag)
+    for (int set=0;set<ICACHE_SETS;set++)
```
RSDuck, line 374: can you add spaces after the semicolons in all the for loops?

src/CP15.cpp
```diff
@@ -332,153 +366,596 @@ u32 ARMv5::RandomLineIndex()
...
+            u8 minSet = ICacheLockDown & (ICACHE_SETS-1);
+            line = line | minSet;
+        }
+    }
+
+#else
+    // Do the same as above but instead of using if-else 
+    // utilize the && and || operators to skip parts of the operations
+    // With the order of comparison we can put the most likely path
+    // checked first
```
RSDuck, line 453: I'm not really sure whether this abusing shortcircuit evaluation really gives better code. If you want to steer the compiler using likely and unlikely markers could be used for the ifs.

src/CP15.cpp
```diff
@@ -332,153 +366,596 @@ u32 ARMv5::RandomLineIndex()
...
+    #if !DISABLE_CACHEWRITEBACK
+        // Before we fill the cacheline, we need to write back dirty content
+        // Datacycles will be incremented by the required cycles to do so
+        DCacheClearByASetAndWay(line & (DCACHE_SETS-1), line >> DCACHE_SETS_LOG2);
+    #endif
+    //Log(LogLevel::Debug,"DCache miss, load @ %08x\n", tag);
+    for (int i = 0; i < DCACHE_LINELENGTH; i+=sizeof(u32))
+    {
+        if (tag+i < ITCMSize)
+        {
```
RSDuck, line 635: oh so you can actually put ITCM and DTCM into cache?

src/CP15.cpp
```diff
@@ -332,153 +366,596 @@ u32 ARMv5::RandomLineIndex()
...
+            u32 *cacheLine = (u32 *)&DCache[(id+set) << DCACHE_LINELENGTH_LOG2];
+            cacheLine[(addr & (DCACHE_LINELENGTH-1)) >> 2] = val;
+            DataCycles = 1;
+            #if !DISABLE_CACHEWRITEBACK
+                if (PU_Map[addr >> CP15_MAP_ENTRYSIZE_LOG2] & CP15_MAP_DCACHEWRITEBACK)
+                {
+                    if (addr & (DCACHE_LINELENGTH / 2))
+                    {
+                        DCacheTags[id+set] |= CACHE_FLAG_DIRTY_UPPERHALF ;
```
RSDuck, line 676: please remove the spaces before semicolon

src/CP15.cpp
```diff
@@ -306,7 +340,7 @@ void ARMv5::UpdateRegionTimings(u32 addrstart, u32 addrend)
             MemTimings[i][0] = bustimings[2] << NDS.ARM9ClockShift;
         }
 
-        if (pu & 0x10)
+        if (pu & CP15_MAP_DCACHEABLE)
         {
             MemTimings[i][1] = kDataCacheTiming;
```
RSDuck, line 345: these values will only ever be used when JIT is enabled, right?

RSDuck (approved):

src/CP15.cpp
DesperateProgrammer: Right now, yes. They are used in ARMv5::CodeRead32, when the JIT is not enabled.

src/CP15.cpp
DesperateProgrammer: I am not entirely sure (I think ) I had a block diagram showing the access passing the write buffer before the TCMs but can't seem to find that diagram. If that is the case, the TCM access will go through the cache when enabled.
However my decision to place the read from TCMs here was to keep the access flow within the line fills in sync to the DataReadXX functions.

src/CP15.cpp
DesperateProgrammer: I have removed this and added the [[(un)likely]] attributes

KostaSaizo7:
Fixes https://github.com/melonDS-emu/melonDS/issues/380

KostaSaizo7:
Unfortunately https://github.com/melonDS-emu/melonDS/issues/359 is still occuring

Jaklyy:
seems like disabling the pu via the control reg causes melonds to hang under this pr?
at least reproduceable under this test rom:
[zboxtesttest.zip](https://github.com/melonDS-emu/melonDS/files/15212734/zboxtesttest.zip)

DesperateProgrammer:
Thanks for the report and sample Jakly,

I'll check that as soon i got some time.

src/CP15.cpp
```diff
@@ -332,153 +366,557 @@ u32 ARMv5::RandomLineIndex()
...
 
-    ICacheTags[line] = tag;
+    ICacheTags[line] = tag | (line & (ICACHE_SETS-1)) | CACHE_FLAG_VALID;
 
     // ouch :/
     //printf("cache miss %08X: %d/%d\n", addr, NDS::ARM9MemTimings[addr >> 14][2], NDS::ARM9MemTimings[addr >> 14][3]);
-    CodeCycles = (NDS.ARM9MemTimings[addr >> 14][2] + (NDS.ARM9MemTimings[addr >> 14][3] * 7)) << NDS.ARM9ClockShift;
-    CurICacheLine = ptr;
+    //                      first N32                                  remaining S32
+    CodeCycles = (NDS.ARM9MemTimings[tag >> 14][2] + (NDS.ARM9MemTimings[tag >> 14][3] * ((DCACHE_LINELENGTH / 4) - 1))) << NDS.ARM9ClockShift;
```
Jaklyy, line 459: minor error: DCACHE_LINELENGTH should be ICACHE_LINELENGTH

Arisotura:
sorry for responding so late, but, given renewed interest in cache emulation, I might snatch this

Arisotura:
I merged it in a separate branch for now. did the bare minimum so it would compile. it seems to run, which is nice.

I'll be reviewing it and possibly applying some fixes with guidance from our local ARM9 expert Jakly.

PR #1956, opened 2024-01-19
Add support for using a portable directory without special build flags.

Steveice10:
Replaces the `PORTABLE` build option with runtime support for picking either a portable or system directory. This allows using melonDS in portable mode on any OS from the release builds instead of having to create your own self-build for each version.

The logic works as follows:
* First, check for the "portable" directory next to the executable (for macOS, next to the app bundle if applicable).
* If the directory does not exist, use the standard path based on OS:
  * Windows: Use the emulator executable's directory as before.
  * Other: Use Qt to determine the appropriate config location.

Three other changes were made to support this:
* Path initialization was moved to a `PathInit` function in `Platform` to keep things tidy.
* Some code that re-calculated the config path in `OpenLocalFile` was removed as it does not appear to be necessary. `EmuDirectory` should already be the correct path in either case.
* The `Platform::Init` call was moved after the construction of `MelonApplication` as `QCoreApplication::applicationDirPath()` requires it to be constructed first.


nadiaholmquist:
Doesn't this break making non-portable builds on Windows? That is something that is intentionally supported because I'd like to make an installer as an option eventually.

Steveice10:
> Doesn't this break making non-portable builds on Windows? That is something that is intentionally supported because I'd like to make an installer as an option eventually.

Windows never supported non-portable builds, the condition was `#if defined(__WIN32__) || defined(PORTABLE)`. If you want to support non-portable builds on Windows the logic under the new `#if defined(__WIN32__)` can be changed.

nadiaholmquist:
Must've been changed sometime along the way then, because when I added that I swear <del>correctly behaving</del> non-portable Windows builds did work.

nadiaholmquist:
Nice, looks good to go to me now.

Steveice10:
> Must've been changed sometime along the way then, because when I added that I swear ~correctly behaving~ non-portable Windows builds did work.

I took a second look at it and I think by some odd logic it was working actually. That second `PORTABLE` condition check was basically discarding the `EmuDirectory` regardless of whatever the `Init` function determined for the paths for some reason.

The new commit should help. It re-adds the `PORTABLE` build flag for Windows to allow turning off the portable behavior where it stores its data in the same directory as the executable by default. Since the Windows default is to store in the same directory as the executable with no indicator file or directory required, I couldn't think of a way to allow this other than with a build flag. May be worth revisiting the Windows behavior down the line, but for now this PR improves things on Linux and macOS while essentially maintaining the status quo on Windows.

Issue #1966, opened 2024-01-28
Lighting inaccuracy on pkmn:ss title screen


Jaklyy:
Lugia appears to be rendered as too dark, with incorrect shading(?)
melonDS:
![image](https://github.com/melonDS-emu/melonDS/assets/102590697/f40843d3-c3a1-41a8-ad46-64e8fb07fc6f)
console: (3ds captured using the frame dump player's "reconstructed full color" screen shot feature)
![image](https://github.com/melonDS-emu/melonDS/assets/102590697/85e1ab0d-a925-4574-b9d9-96283aac2eeb)
rounding up the ambient lighting (by adding 0x10 before shifting right by 5) gets us closer, but ever so slightly off.
![image](https://github.com/melonDS-emu/melonDS/assets/102590697/b25ee1a1-7971-4c45-97ae-3ff850ed5466)

PR #1967, opened 2024-01-29
Improve calculation of light colors

Jaklyy:
Edit: This pr improves the algorithm for diffuse lighting to be (near?) perfect, thanks to discoveries by asuza.
This also changes the way fractional color precision is handled to be maintained until all lights are calculated.

Previous text:
Fixes https://github.com/melonDS-emu/melonDS/issues/1966
Results in lighting being noticeably brighter when multiple lights are enabled.
Colors seem to match hardware better.
Doesn't appear to be perfect, though it's possible those imperfections are due to inaccurate interpolation rather than problems with the lighting calc itself.
Definitely needs some test rom verification, though.

Before:
image
After:
image
HW Capture (frame dump "reconstructed full color" screenshot):
image


Jaklyy:
messing with a test rom reveals that there's definitely still something ever so slightly off in the color calcs
but it seems to overall be much closer to hardware

Jaklyy:
Emissive and Ambient lighting are pixel perfect.
Diffuse light *should* be pixel perfect (minus overflow behavior)
Specular still needs work.

Issue #1143, opened 2021-06-29
Rendering Issue (Homebrew)


BrandonKaim:
If you try running this on melonDS https://github.com/devkitPro/nds-examples/tree/master/Graphics/gl2d/dual_screen it only renders the bottom screen while the top is left completely blank, on real hardware both screens display shapes

RSDuck:
the issue is direct boot, it works if the homebrew is booted through the firmware menu.

The firmware seems to initialise the polygon attr to disable the zero dot check which when enabled prevents the drawing on the topscreen which is only done using dots.

RSDuck:
the issue lies in the zero-dot W check, which clips away all the dots which make up the graphics on the top screen.

PR #1968, opened 2024-01-30
Default zero dot display register to the 24 bit integer limit

Jaklyy:
fixes: https://github.com/melonDS-emu/melonDS/issues/1143
fixes a fail on one of Striker's line test roms

tested via this test rom installed to nand as DSiware on a new3DS: [testdepth.zip](https://github.com/melonDS-emu/melonDS/files/14103323/testdepth.zip)

default behavior appears to be that it renders all the way up to the 24 bit integer limit (at which point it overflows to 0)

### DIFF
--- a/src/GPU3D.h
+++ b/src/GPU3D.h
@@ -197,7 +197,7 @@ public:
 
     FIFO<CmdFIFOEntry, 64> CmdStallQueue {};
 
-    u32 ZeroDotWLimit = 0;
+    u32 ZeroDotWLimit = 0xFFFFFF;
 
     u32 GXStat = 0;
 
--- a/src/GPU3D.cpp
+++ b/src/GPU3D.cpp
@@ -191,7 +191,7 @@ void GPU3D::Reset() noexcept
 
     CmdStallQueue.Clear();
 
-    ZeroDotWLimit = 0; // CHECKME
+    ZeroDotWLimit = 0xFFFFFF;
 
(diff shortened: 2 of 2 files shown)


src/GPU3D.h
```diff
@@ -197,7 +197,7 @@ class GPU3D
 
     FIFO<CmdFIFOEntry, 64> CmdStallQueue {};
 
-    u32 ZeroDotWLimit = 0;
+    u32 ZeroDotWLimit = 0xFFFFFF;
```
RSDuck, line 200: doesn't this need to be changed in Reset?

Jaklyy: i did change it there, i just now also changed the variable's default value to look pretty/avoid confusion.

Jaklyy: though looking at it now no other variable gets initialized in gpu3d.h to its actual default as stated in the reset function
so maybe i should revert that change to avoid inconsistency?

RSDuck: there isn't really a reason for the variables to be initialised in the header, except to prevent UB in case they are read before Reset is called (though that shouldn't happen if that makes any sense).

Jaklyy: yeah, so it's mostly just personal preference what value they get initialized to
i wont worry about it
if someone wants it changed back for consistency with the other values being 0 initialized I can revert that change easily

PR #1979, opened 2024-02-18
[FIX] fix malloc on OpenBSD targets

Izder456:
over on the [ports@ mailing lists](https://marc.info/?t=170813350900001&r=1&w=2), It was requested to push this patch upstream to fix building on OpenBSD targets.

TL;DR:

OpenBSD doesn't have `malloc.h`, our `malloc()` function is in `stdlib.h` 

[see `malloc(3)`](http://man.openbsd.org/malloc)

trivial patch, Makes our port nicer to handle.

thanks.

### DIFF
--- a/src/frontend/duckstation/gl/context.cpp
+++ b/src/frontend/duckstation/gl/context.cpp
@@ -3,11 +3,7 @@
 #include "loader.h"
 #include <cstdlib>
 #include <cstring>
-#ifdef __APPLE__
 #include <stdlib.h>
-#else
-#include <malloc.h>
-#endif
 Log_SetChannel(GL::Context);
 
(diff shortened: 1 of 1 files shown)


bentley:
stdlib.h is the correct header on Linux too. So the commit d1f30ca3ec720b493cc105d7ada9bdf5b772b438 should be dropped.

nadiaholmquist:
> stdlib.h is the correct header on Linux too.

If that is the case, could you revert that commit and I'll merge this?

Izder456:
I can revert that commit so MAINTAINER should not need to. give me a second.

Izder456:
commit d1f30ca has been reverted.

OK?

PR #1994, opened 2024-03-06
Misc fixes to shadow/masks and the <= depth test

Jaklyy:
~~Small fix.~~
~~Fixes swapped spans not overwriting/not being overwritten by another polygon via "less than" depth test mode when their z values are equal.~~

~~Examples:~~
image
image

repurposing this pr for my misc fixes because there's a lot of them and they all overlap in some way

Changes:
New: swapped spans have their back facing & front facing flags inverted.
New: xmajor top edges over xmajor bottom edges/ymajor left edge over ymajor right edges invert the back facing flag's role in the <= depth test.
Rework: edge flags now track xmajor top/bottom, ymajor left/right instead of top/bottom/left/right edge
Fix: shadow masks dont care about their alpha value unless it's 0 (and even then it's only used to determine edge fill)
New: begin support for scfg raster rev bit (currently: alters stencil buffer clear behavior, partially fixes a bug with compressed textures)
Fix: stencil buffer is only cleared when starting a mask polygon, and only if a shadow polygon has been rendered after the last mask
Fix: opaque shadow polygons don't write/push down the depth/attr buffer
Cleanup: Remove a bunch of useless variables from the shadow mask function.
Fix: shadow and shadow mask polygons don't have a <= depth test
Fix: shadow polygons dont render when wireframe
Fix: translucent pixels dont update the back facing flag
Fix: always push down the depth buffer, even when it 'should' be useless, because a niche quirk with shadow masks can result in it being used anyway
Cleanup: restructure shadow/mask handling to trust that the 2nd bit cannot be set without the first bit also being set.
Fix: opaque shadows need a different opaque poly id to render (not sure about translucent id?)
Fix: shadow, shadow mask, and decal polygons all count as translucent for poly sorting when a translucent texture mode is applied.
Fix: pixels with translucency from textures should be filled
New: implement the interaction between edge marking and edge flags
New: implement the bug that causes 4x4 compressed textures to be 1/64th darker than they should otherwise be in modes that use interpolated colors
Fix: when using 4x4 compressed textures, the interpolated colors do not discard color precision down to 5 bit, they stay 6 bit

### DIFF
--- a/src/GPU3D.cpp
+++ b/src/GPU3D.cpp
@@ -569,20 +569,7 @@ void GPU3D::DoSavestate(Savestate* file) noexcept
     }
 
     // save any renderer state that can persist through frames
-    if (softRenderer)
-    {
-        file->VarArray(softRenderer->StencilBuffer, sizeof(SoftRenderer::StencilBuffer));
-        file->VarArray(softRenderer->ShadowRendered, sizeof(SoftRenderer::ShadowRendered));
-        file->VarArray(softRenderer->ShadowRenderedi, sizeof(SoftRenderer::ShadowRenderedi));
-    }
-    else
-    {
-        u8 ph[256*2] {};
-        // placeholders for renderers that don't support the same variables
-        file->VarArray(ph, sizeof(SoftRenderer::StencilBuffer));
-        file->VarArray(ph, sizeof(SoftRenderer::ShadowRendered));
-        file->VarArray(ph, sizeof(SoftRenderer::ShadowRenderedi));
-    }
+    CurrentRenderer->DoSavestate(file);
 
     if (softRenderer && softRenderer->IsThreaded())
     {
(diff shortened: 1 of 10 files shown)


src/DSi.h
```diff
@@ -129,6 +129,8 @@ class DSi final : public NDS
     void ARM7IOWrite16(u32 addr, u16 val) override;
     void ARM7IOWrite32(u32 addr, u32 val) override;
 
+    bool GetSCFGRasterBit() override;
```
JesseTG, line 132: Could you mark this method as `const` so that it can be used with `const` references?

JesseTG (changes requested):

src/DSi.h
Jaklyy: it is const now

Jaklyy:
this pr is full of good work but might be worth rewriting to implement individual fixes in separate prs instead

Issue #1120, opened 2021-06-05
Input reaction time altered by some settings


Tatsuya79:
I used the new framestep hotkey to do some tests: 
keeping a direction pushed down then counting how many frame steps it takes to see a move on screen.

In Mario Kart DS on the title screen:
-3 frames to see the selection moved in software
-9 frames in gl
-5 frames in gl with audio sync off

I wonder if there's something wrong here.

tested on win7 x64

Ryunam:
This confirms that the regression initially spotted in the libretro core actually derives from upstream code. See here for the issue report that was posted in the libretro core repo: https://github.com/libretro/melonDS/issues/100

Based on my tests, I believe that these 2 frames of latency started to appear with the "upstream changes" that were referenced in this commit, dating back to January-February (exact timing could be different though): https://github.com/libretro/melonDS/commit/25ec9c05e58944631e5c59711d0e3948c7b4f0ba

RSDuck:
the regression was introduced in https://github.com/Arisotura/melonDS/commit/a3f4aaf50314f6ab528afb8dbcef88cfb012e765

Ryunam:
Hi, thank you for identifying the commit where this issue started. As it stands right now, unfortunately using the GL renderer in both the libretro core and standalone yields a noticeably unresponsive feel, with the input being particularly sluggish compared to software. I hope a workaround is possible and/or planned in the near future.

Immersion95:
Any news on this matter ?

Immersion95:
I'm sorry to bump this but it has been 2 years that the OpenGL core has 2 frames of input lag compared to Software. While I do appreciate all the other effort being done, I think this is an important matter.

Would it be possible to retract this commit https://github.com/melonDS-emu/melonDS/commit/a3f4aaf50314f6ab528afb8dbcef88cfb012e765 ?

RSDuck:
I found the actual cause of the issue a while back and fixed it on the compute shader renderer branch (for both the OpenGL renderer and the compute shader renderer).

https://github.com/melonDS-emu/melonDS/commit/a3f4aaf50314f6ab528afb8dbcef88cfb012e765 only removed something which accidentally made it better

Immersion95:
> I found the actual cause of the issue a while back and fixed it on the compute shader renderer branch (for both the OpenGL renderer and the compute shader renderer).
> 
[...]

PR #1995, opened 2024-03-09
Fix the 2 frames of input lag on the GL renderer

Immersion95:
I just backported [the fix that was present on the compute shader renderer branch](https://github.com/melonDS-emu/melonDS/commit/d9c08341203f7d5d7ac78761f007b470fca25f58), none of this is my work. I built it and it seems to work just fine.

I know that it is planned on the big refactoring but the input lag is present on the GL renderer since 2021, hope this is merged 😊.

"Don't double buffer 3D framebuffers for the GL Renderer - Looks like leftovers from when 3D+2D composition was done in the frontend".

The fix is from @RSDuck - Thanks a lot man ! By the way I'm French and it's "Ne jamais toucher la lentille." - "Lentile" doesn't exist 😭.

Closes https://github.com/melonDS-emu/melonDS/issues/1120


Immersion95:
Already implemented in https://github.com/melonDS-emu/melonDS/pull/2041 :)

PR #1999, opened 2024-03-16
ssize_t is not defined in GdbStub.h or GdbCmds.h

RealAstolfo:
i had to add sys/types.h and patch my gentoo ebuilds for a successful compilation.

### DIFF
--- a/src/debug/GdbStub.h
+++ b/src/debug/GdbStub.h
@@ -3,6 +3,7 @@
 #define GDBSTUB_H_
 
 #include <stddef.h>
+#include <sys/types.h>
 #include <map>
 #include <vector>
 


RSDuck:
https://en.cppreference.com/w/cpp/types/size_t

> Defined in header [<cstddef>](https://en.cppreference.com/w/cpp/header/cstddef)

?

CasualPokePlayer:
ssize_t is the non-standard signed variant of size_t

RealAstolfo:
> ssize_t is the non-standard signed variant of size_t. It would typically also only be defined on Unix-y platforms (on Windows you'd have to define it yourself, probably to intptr_t since that's the same in that context)

exactly this. only POSIX systems have ssize_t, the windows equivelent is actually SSIZE_T



> https://en.cppreference.com/w/cpp/types/size_t
> 
> > Defined in header [<cstddef>](https://en.cppreference.com/w/cpp/header/cstddef)
> 
> ?

as for this, that is C++, this is a C header. including <cstddef> in a C file is a good way to make the compiler switch to C++ mode imo

RealAstolfo:
i just ammended the changes to something more appropriate for C++, which should still fix the cited build issues on ssize_t on my platform

src/debug/GdbCmds.h
```diff
@@ -1,4 +1,4 @@
-
+#include <cstddef>
```
CasualPokePlayer, line 1: GdbCmds.h is only every supposed to be included in gdb stub, so the include here is pointless.

RealAstolfo: got it. ill remove that part of the patch (on the other hand, what a weird way to include code)

RealAstolfo:
Seems like the patch should be finalized, so heres the review.

What i did: 
added `#include <sys/types.h>` to src/debug/GdbStub.h

Why:
on this gentoo system (specifically the musl libc), the program would fail to compile with errors saying ssize_t is not declared/defined.

What it accomplished:
with -DGDB_ENABLE, the gentoo system now successfully compiles the program,
and now the program is more compliant to posix.

PR #2011, opened 2024-04-06
sort of hack to fix extremely cursed nonsense with the way a polygon's orientation is determined

Jaklyy:
so sometimes a polygon is 'normal' and is neither front or back facing
but under specific circumstances a polygon that *should* have a completely identical dot ends up facing *both* front and backfacing
this fixes this to work like on hardware maybe i think
this is almost certainly __not__ the correct fix.
but i haven't found any cases that are wrong so it's good enough for me(?)
but i guarantee you there's loads of them and i just haven't found them yet.

### DIFF
--- a/src/GPU3D.cpp
+++ b/src/GPU3D.cpp
@@ -1013,6 +1013,18 @@ void GPU3D::SubmitPolygon() noexcept
             return;
         }
     }
+    // dot == 0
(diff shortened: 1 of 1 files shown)


Jaklyy:
Thought of a (mostly) more sensible solution, closing.

PR #2028, opened 2024-05-05
Implement toggling bit 15 of the cp15 control register

Jaklyy:
implements disabling switching to/from thumb mode via ldm ldr and pop for the arm 9 via control register
note: not implemented for the jit currently.


CalebPancake:
ummm
what

PR #2043, opened 2024-05-13
Fedora build

IvanVeloz:
Hi everyone,

I would like to add build instructions for Fedora. This was tested on the current Fedora 40.

Best regards,
Ivan Veloz


ankon:
Haven't checked whether these are transitive, but I also needed 
* qt5-qtsvg-devel
* libarchive-devel
* enet-devel

Issue #2047, opened 2024-05-16
Compute renderer displays garbage at high IRs


FireNX70:
Resolutions higher than 6x are very prone to displaying garbage. 6x will display garbage in certain spots, although Zelda Spirit Tracks triggers this pretty easily. 5x seems fine. Tested on a 1080Ti + Win10. Tested with https://github.com/melonDS-emu/melonDS/commit/a72b79a55ad2d61811af11b1b911f6af863f66c2, so https://github.com/melonDS-emu/melonDS/commit/5df83c97c766bff3da8ba5a1504a6a5974467133 doesn't fix it.

Some examples
<img width="424" alt="pkmn_bw_6x" src="https://github.com/melonDS-emu/melonDS/assets/23368648/1db1a9f9-2a0e-44f9-95a4-777280cde169"> Pokemon BW Driftveil Bridge 6x
<img width="424" alt="pkmn_pt_spot_5x" src="https://github.com/melonDS-emu/melonDS/assets/23368648/8ab57000-33a6-4769-855f-3f61597c73f9"> Pokemon Platinum 5x
<img width="423" alt="pkmn_pt_spot_6x" src="https://github.com/melonDS-emu/melonDS/assets/23368648/940e17a0-47ec-446c-b504-af6b085c8977"> Pokemon Platinum 6x
<img width="424" alt="pkmn_pt_next_to_spot_6x" src="https://github.com/melonDS-emu/melonDS/assets/23368648/cf890204-9381-4690-98fb-4291f6fdfc14"> Pokemon Platinum 6x
<img width="422" alt="pkmn_pt_next_to_spot_7x" src="https://github.com/melonDS-emu/melonDS/assets/23368648/8d8b220b-de9b-4044-9844-91c310cc4af7"> Pokemon Platinum 7x
<img width="421" alt="tloz_ph_6x" src="https://github.com/melonDS-emu/melonDS/assets/23368648/8c922d04-1289-4e43-8096-b8736481e77c"> Zelda Phantom Hourglass 6x
<img width="423" alt="tloz_ph_7x" src="https://github.com/melonDS-emu/melonDS/assets/23368648/61409057-4428-4658-a4de-00b630d09f7b"> Zelda Phantom Hourglass 7x
<img width="423" alt="tloz_st_5x" src="https://github.com/melonDS-emu/melonDS/assets/23368648/88125b21-d121-4416-b7c4-96f374aa1fe7"> Zelda Spirit Tracks 5x
[...]

PR #2065, opened 2024-06-13
Implement tile size scaling for the compute renderer

FireNX70:
Based on @RSDuck's comments in https://github.com/melonDS-emu/melonDS/issues/2047.

Right now it scales starting at 4.5x (effectively 5x). Based on some more recent testing, it looks like it should start scaling at 3.25x (effectively 4x). [See this gist for more information.](https://gist.github.com/FireNX70/513e22e542dd5016de6ed1ce1397e592) It also contains notes on the rationale behind the values for `ClearCoarseBinMaskLocalSize`.

The value used for `CoarseTileCountY` when using 32x32 tiles was chosen to match the base resolution's vertical resolution of 192 (6 32x32 tiles vertically per coarse tile means coarse tiles end up being 256x192).


CasualPokePlayer:
It seems more that the files got converted from LF to CRLF for some reason (text editor/IDE goof?)

FireNX70:
You're right, looks like VSCodium got me. Should be fairly easy to fix.

FireNX70:
Ok, looks like it's fixed but I'm not sure the line endings for the template I added to Utils.h aren't messed up. Thanks for the tip @CasualPokePlayer.

OpenSauce04:
Just tested this on my end, and it indeed seems like this PR makes the issue disappear completely. Is there any particular holdup for this being merged?

FireNX70:
Tile scaling alone will never fix the problem completely. At scales 13x and higher it would need tiles larger than 32x32. This isn't feasible because there's a limit of 1024 on the product of the local sizes in compute shaders. Implementations are free to raise this limit, but nVidia (which is the one implementation that has the problem with UB to begin with) doesn't, so we're limited to 32x32 tiles (the tile size is used as the x and y local sizes in at least one shader).

As for 4x and 8x, it's what I wrote in the OP. My branch scales on a factor of 4.5, it looks like more aggressive scaling is needed. Specifically, using a factor of 3.25 (scaling at 4x and 7x, instead of 5x and 9x) seems like it should be fine. 13x and higher will still be a problem, so I was unsure if it was even worth changing and I decided to wait for @RSDuck's input.

FYI, the worst-case scenario I used for testing was MKDS's Desert Hills fire balls (near the pyramid, during laps 2 and 3). You have to try and get them as close to the camera as possible.

extherian:
I found this implementation flawless in Phantom Hourglass and Spirit Tracks, at any resolution from 1x to 16x. A far as I can tell, it completely fixes this issue. What's the holdup on getting it merged? Is this planned to be part of the November v1.0 release?

Tagger102:
Hopefully this will be officially merged into main! It fixes all issues I noticed testing on my RTX 3090, currently using 1.0RC with this fix merged into it.

user18081972:
Why not merge this for 1.0? Would save alot of issues being opened about this.
It has seemingly no downsides (other then it maybe not being the prettiest solution code wise). And I would love not having to keep merging this in myself and compiling from source.

Arisotura:
we were looking at it for the 1.0 release, but Jakly reported that it breaks the renderer for her

so I think this will have to wait longer... sorry :(

Jaklyy:
pr seems to work fine on windows 10 (though i can seemingly no longer reproduce any issue in the first place there?) at least with amd pro drivers (because windows decided i need those for some reason??)
pr seems to be completely broken on linux though. the renderer just does not work under any circumstances there.

FireNX70:
I've done a quick test on a live Debian 12 using nouveau and it worked fine. I'm pretty sure I tested llvmpipe and nVidia's proprietary driver back when I worked on this. I also tested a Vega iGPU on Windows 10 back then, and it worked fine. It's gotta be a Radeon+Linux thing. I might be able to test on a GCN card in a couple of days.
@Jaklyy I assume you're using the regular mesa driver stack, with radeonsi as the OpenGL implementation. Is that correct?

Addendum:
It's really strange that it would break completely regardless of resolution. There's two changes that affect behavior no matter your settings: 57d20751fcd1ff52fe9f8890d565c65c5b6abc13 and 6f8ce9fe897fab0277754209122df4181fb4a177.
The first one was replacing a macro which looked wrong with the one that seemed right; if anything, it's surprising it worked fine before.
The second one changed the default local_size for ClearCoarseBinMask from 32 to 64. I had to change it to 48 at higher resolutions, so I thought I might as well make it 64 at the lower ones, which should be a bit more efficient on GCN while leaving performance untouched on nVidia and RDNA.
Anyway, you can always try reverting either (or both) and see how it goes.

Edit2:
Scratch that, it's broken. I was on the wrong branch, which incidentally confirms it's something in this PR that breaks it.
It breaks catastrophically enough that it borked my KDE session (but only the second time I launched it?).

RSDuck:
I thinking about calculating the area of all polygons on the cpu in beforehand to adjust all buffers. That doesn't solve the issue of index overflow, for this I'd might borrow something from this PR.

FireNX70:
Should be fixed now. @Jaklyy please retest.

So it turns out it was the barriers (https://github.com/melonDS-emu/melonDS/commit/57d20751fcd1ff52fe9f8890d565c65c5b6abc13). GL_SHADER_STORAGE_BUFFER was defined as 37074 (0b1001000011010010). The only valid flags in there are GL_ELEMENT_ARRAY_BARRIER_BIT (2), GL_COMMAND_BARRIER_BIT (64), GL_PIXEL_BUFFER_BARRIER_BIT (128) and GL_ATOMIC_COUNTER_BARRIER_BIT (4096). Ultimately, it turned out to be GL_COMMAND_BARRIER_BIT that made it work. I kept GL_SHADER_STORAGE_BARRIER_BIT in just in case, since that works fine everywhere else it seems.

Jaklyy:
seems to work fine on my end now

Arisotura:
so, shall we merge this? does it get the Seal of Approval(tm)?

PR #2074, opened 2024-06-20
Fix XML information in GdbCmds.cpp and add support for the 'vCont;' command.

enler:
I previously encountered some issues when trying to debug with GDB provided by IDA Pro while connecting to melonDS. I made corrections to address these issues, and after the fixes, debugging in IDA Pro works as shown in the image below. Additionally, I noticed some bugs in the GDB-related configurations in the .toml file, but I am unsure how to fix them properly.
screenshot 2024-06-20 234217

### DIFF
--- a/src/debug/GdbCmds.cpp
+++ b/src/debug/GdbCmds.cpp
@@ -46,7 +46,7 @@ const char* TARGET_INFO_ARM9 = "cputype:12;cpusubtype:7;triple:arm-none-eabi;ost
 	"<reg name=\"r11\" bitsize=\"32\" type=\"uint32\"/>" \
 	"<reg name=\"r12\" bitsize=\"32\" type=\"uint32\"/>" \
 	"<reg name=\"sp\" bitsize=\"32\" type=\"data_ptr\"/>" \
-	"<reg name=\"lr\" bitsize=\"32\" type=\"code_ptr\"/>" \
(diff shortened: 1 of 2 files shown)


src/debug/GdbCmds.cpp
```diff
@@ -46,7 +46,7 @@ const char* TARGET_INFO_ARM9 = "cputype:12;cpusubtype:7;triple:arm-none-eabi;ost
 	"<reg name=\"r11\" bitsize=\"32\" type=\"uint32\"/>" \
 	"<reg name=\"r12\" bitsize=\"32\" type=\"uint32\"/>" \
 	"<reg name=\"sp\" bitsize=\"32\" type=\"data_ptr\"/>" \
-	"<reg name=\"lr\" bitsize=\"32\" type=\"code_ptr\"/>" \
```
RSDuck, line 49: lr definitely points to code (most of the time)?

enler: The situation is like this: when using IDA Pro for debugging, its view will follow the first register of type code_ptr. In the previous code, it would follow the lr register, causing issues with the view display during debugging.

RSDuck: That does sound kind of stupid, though I guess you could move pc up?

PR #2084, opened 2024-07-03
Fix horizontal mosaic on sprites

AntonioND:
The code itself is correct, but the variable being read is the Y mosaic
coordinate, not the X coordinate.


AntonioND:
Test ROM: [effects_mosaic.nds.zip](https://github.com/user-attachments/files/16090660/effects_mosaic.nds.zip)

You can see that the mosaic of the sprite is the same in both coordinates.

PR #2088, opened 2024-07-07
Fix build for clang

Patater:
Remove declaring POSIX_SOURCE manually. Let the compiler take care of this on its own. Additionally, include a missing header file for the definition of struct sockaddr_in, which we need before we can allocate the struct. With these fixes, we can build with Clang 18.1.5, although there are many warnings.


Patater:
Build failed on Windows. I've updated the commit to restore the `_POSIX_SOURCE` defines when building for Windows.

Patater:
Rebased on latest `master`

PR #2103, opened 2024-07-18
Expose some fields of `DSi_BPTWL` needed to support direct boot

JesseTG:
This PR exposes the registers defined in `DSi_BPTWL`, so that direct-boot can be used for DSi games. See https://github.com/JesseTG/melonds-ds/issues/148 for details on why I'd like to be able to do this.


Arisotura:
why do we need to turn Registers into a std::array

JesseTG:
> why do we need to turn Registers into a std::array

`std::array` is more flexible than a plain C array without any overhead (since it's a plain wrapper). But I may have gotten a little trigger-happy; I can undo that part if you'd like.

PR #2127, opened 2024-08-06
pointless micro-optimizations

Jaklyy:
nets us like
1-2 fps maybe?


asiekierka:
@RSDuck Do you think this can be closed now? The POPCNT intrinsic was merged, the for loop replacement was shot down, and I'm not sure I see a point in the multiplication/shift replacements.

PR #2138, opened 2024-08-18
Add an About dialog with build info

nadiaholmquist:
This adds an About dialog showing the versions as well as credits to the melonDS team and build information.

The build information is specified at configure time using either environment variables or CMake options, to make it ideal for packaging and for GitHub Actions builds.

Here's how it currently looks:
image

Feedback on the UI as well as what build info it would be useful to include is very welcome.


Mrlinkwii:
image
looks very nice 

how dose a master build look like , i assume the branch , will just mention master

nadiaholmquist:
Yeah, it should just say `master` there. The information comes from whatever the GitHub Actions job is supplied with.

PR #2146, opened 2024-08-29
Add packet capture for multiplayer traffic

liclac:
This adds support for logging network traffic in tcpdump (`.pcap`) format. You can enable this in `Config -> Emu -> Devtools`.

This seems to work fine when capturing traffic between two multiplayer instances, verified by booting two instances to firmware, opening Pictochat and comparing the logs to promiscuous mode captures of real hardware doing the same.

However, when a game connects to the internet, you'll currently get a partial log - the `Platform::MP_Send*` hooks are called for outgoing packets, but the `Platform::MP_Recv*` ones aren't called for the incoming ones. The `Platform::Net_Recv*` hooks *are* called for them, but they're just passed the source/destination IP, packet type and contents, not the full WiFi frame, which isn't enough to log them.

I don't understand the WiFi emulation code nearly well enough to figure out quite what the right behaviour or thing to do here is, so hopefully somebody who does could help me out here.

You can reproduce this very easily by enabling packet capture, booting up any game with WFC support and tapping Nintendo WFC Settings -> Nintendo WiFi Connection Settings -> Connection 1 -> Test Connection. If you open the pcap file in Wireshark, you'll see an outgoing DNS query for `conntest.nintendowifi.net`, followed by a HTTP GET request, but the responses are missing.

Anyway, this is what `Config -> Emu -> Devtools` looks like now:
image

(The `.pcap` files for secondary multiplayer instances are named `myrom.2.pcap`, rather than `myrom.pcap.2` as is the case for other files, simply so that they show up in eg. Wireshark's file picker and have the right extension when opened by an external program. I can change this for consistency if wanted.)


lnee94:
If you fork and run the github actions this a post the builds it genrates here (note im on linux) I can test it

lnee94:
I tested and reslsed that I need more attention

liclac:
I'm not sure how to do that, but happy to do so if you could tell me how?

But if you're on Linux, would it not be easier to download the archive of this branch and build it from source?

PR #2159, opened 2024-09-26
implement 3 configurable and toggleable framerate targets

Jaklyy:
This pr allows for configuring the framerate target and adds support for two other framerate targets, a "fastforward" and "slowmo" target which can be enabled via either a toggle or holding a button.
this allows for supporting a more accurate framerate target and allows for users to slow down the speed of gameplay if they so desire


SomeGuyDoinWork:
why bother project dead none of your other good pr have been merged :/

any chance you could publish a release with all your fixes?

Issue #1945, opened 2024-01-08
[Regression] RTC does not handle 24h mode anymore


Real96:
Seems like melonDS atm does not have the 24h mode for rtc, only the 12h one. Maybe it should be good to leave the choice to the user with an setting option(?).

[This](https://www.reddit.com/r/pokemonrng/comments/okiby3/need_help_with_bw_rng_on_melonds/) is the issue i'm facing: gen5 pokemon games don't generate correct initial seeds if the time is set after 12:00 pm. The initial seed is rtc dependant.

The recent rtc commits seem to have caused it. It was working fine before that commit.

CasualPokePlayer:
This isn't a regression per se, it was only ever ""correct"" on BizHawk once due to its changes, since it took all the changes from upstream (since they make the BizHawk changes obsolete) it "regressed" for BizHawk specifically.

To be clear, what the actual issue here is strange. By default, melonDS is in 12 hour mode. Pokemon games seem to expect 24 hour mode. Reportedly, even booting into firmware and changing the time in firmware did not result in the "correct" seeds, indicating that either the firmware and/or Pokemon does not bother switching to 24 hour mode, or they do but that switch isn't happening correctly for some reason, or the AM/PM flag is simply not being reported correctly in 24 hour mode even when it should be.

The last doesn't seem right looking at the code, so I could guess it's either the first or second possibilities.

Arisotura:
we've been talking about this and apparently the DS just uses 24-hour mode always from what it seems, so we might as well just make that the default

BlueSwordM:
Has there been any progress on the issue or has it been fixed on Git already?

kaitouctr:
I just tried a small bit of Gen 5 RNG (grabbing the initial seed via both needles and the memory viewer) and it seems like the RNG is now working properly again? I set the emulated time to 01/01/2000 15:00:00 then I do the usual process of determining the seed I am landing on and it looks the RNG is generating the correct seeds again as PokeFinder was able to find the seed I landed on.

I tested on melonDS's built in firmware, a DS firmware dumped from a DSi and a DS Lite firmware and all three are generating the correct initial seeds.

CasualPokePlayer:
The PM flag is not the issue in the first place. The problem is whether 12 hour mode or 24 hour mode is being used. The default in melonDS upstream is still using 12 hour mode, not 24 hour mode. If 24 hour mode was saved at some point somehow for you, you'd also not experience issues.

[...]

PR #2166, opened 2024-10-21
fix: set default mode to 24 hours

kaitouctr:
Here is my hacky fix for #1945 

This pull request should allow the RTC to always be initialised in 24 hours mode.


Arisotura:
looks mergeable to me. that would put that issue to rest

PR #2167, opened 2024-10-23
Fix gdb break on start & gdb ports not closing after restarting/crashing

### LINKED ISSUE #2007: bug(DevTools): break on startup has no effect
## Bug

When selecting `break on startup` the game does not break at startup

https://github.com/melonDS-emu/melonDS/assets/42176370/7ff3224e-f5bf-4940-a3f3-3afd5a90e7a4

## Reproduce

Tested with master https://github.com/melonDS-emu/melonDS/commit/31a7f53282040c7df93204d647de470390b96c4b 

- (1) git checkout 
- (2) follow build instruction standalone windows
- (3) start emulator
- (4) change settings for AMR9
- (5) restart the emulator
- (6) boot a game
- (7) no break

GalaxyShard:
Closes #2007

This PR fixes two bugs: gdb ports not being freed (resulting in errors when trying to debug, see below) and Break On Startup not having any effect.

Prior to this, restarting melonDS with GDB enabled may produce the following error:

```
[GDB] err: can't bind to address <any> and port 7000
```

This PR fixes the error by enabling SO_REUSEADDR on the socket opened for the GDB stub. The error no longer occurs on my end with this patch.

### DIFF
--- a/src/ARM.cpp
+++ b/src/ARM.cpp
@@ -110,6 +110,7 @@ const u32 ARM::ConditionTable[16] =
 ARM::ARM(u32 num, bool jit, std::optional<GDBArgs> gdb, melonDS::NDS& nds) :
 #ifdef GDBSTUB_ENABLED
     GdbStub(this, gdb ? (num ? gdb->PortARM7 : gdb->PortARM9) : 0),
+    BreakOnStartup(gdb ? (num ? gdb->ARM7BreakOnStartup : gdb->ARM9BreakOnStartup) : false),
 #endif
     Num(num), // well uh
     NDS(nds)
--- a/src/debug/GdbStub.cpp
+++ b/src/debug/GdbStub.cpp
@@ -101,6 +101,11 @@ bool GdbStub::Init()
 		Log(LogLevel::Error, "[GDB] err: can't create a socket fd\n");
 		goto err;
 	}
+	{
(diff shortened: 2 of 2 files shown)


Arisotura:
`D:/a/melonDS/melonDS/src/debug/GdbStub.cpp: In member function 'bool Gdb::GdbStub::Init()':
Error: D:/a/melonDS/melonDS/src/debug/GdbStub.cpp:107:62: error: cannot convert 'int*' to 'const char*'
  107 |                 setsockopt(SockFd, SOL_SOCKET, SO_REUSEADDR, &enable, sizeof(enable));`

just fix that error and this will be good to merge :)

GalaxyShard:
Should be fixed now. Looks like Windows takes a `const char*` even though it is still read as an integer according to the [docs](https://learn.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-setsockopt). I don't have a windows machine to test it on, however.

Arisotura:
`Error: D:/a/melonDS/melonDS/src/debug/GdbStub.cpp:107:62: error: invalid conversion from 'void*' to 'const char*' [-fpermissive]`

looks like it's gonna be an ass about this

just cast it to (const char*) and it should work

RSDuck:
I can't believe I missed that one while fixing gdb issues. Break on start did work for me, but I didn't test turning off, so that was probably unitialised memory :/

PR #2168, opened 2024-10-23
Implement BizHawk's touch input interpolation changes in the core

CasualPokePlayer:
"Touch input interpolation" here is rather simple: it does a simple linear interpolation against the touch values between the last frame's touch position and the current frame's touch position, using the current frame cycle count. In case the last frame wasn't touching anything, it just touches immediately.

The API presented here was originally intended as a way to try to avoid merge conflicts. It is intended that before RunFrame(), the user would call MoveTouch(), then call TouchScreen() once RunFrame() returns.

Alternatively, TouchScreen() could be solely called before RunFrame() (this would probably be what the libretro port does, maybe some other ports). This however will run into issues for some games which do not like immediate changes to touch input every frame, rather needing touch input to go slowly across the frame, as touch input gets polled multiple times in a frame (see https://github.com/TASEmulators/BizHawk/issues/3397).

melonDS Qt doesn't do this however, instead just calling TouchScreen() within mouse event handlers, i.e. within the UI thread, not the emulator thread. This does end up making it (at least seemingly) avoid issues with games not liking exactly per frame touch changes, but it is just a giant race condition and probably should be changed.

### DIFF
--- a/src/DSi_SPI_TSC.cpp
+++ b/src/DSi_SPI_TSC.cpp
@@ -88,6 +88,8 @@ void DSi_TSC::SetTouchCoords(u16 x, u16 y)
     TouchX = x;
     TouchY = y;
 
+    DeltaX = DeltaY = 0;
+
     u8 oldpress = Bank3Regs[0x0E] & 0x01;
 
     if (y == 0xFFF)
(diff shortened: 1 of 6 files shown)


JesseTG:
> Alternatively, this newer API doesn't have to be used: TouchScreen() could be solely called before RunFrame() (this would probably be what the libretro port does, maybe some other ports). This however will run into issues for some games which do not like immediate changes to touch input every frame, rather needing touch input to go slowly across the frame, as touch input gets polled multiple times in a frame (see [TASEmulators/BizHawk#3397](https://github.com/TASEmulators/BizHawk/issues/3397)).

Can you suggest such a problematic game? I'd like to compare the results before integrating this change and after.

CasualPokePlayer:
> Can you suggest such a problematic game? I'd like to compare the results before integrating this change and after.

Just see the linked issue, which names such a game and provides SaveRAM for the problematic section.

PR #2177, opened 2024-10-28
Disable options when renderer requirements aren't met

Mrcubix:
More complicated that i would had hoped for it to be.

Disable OpenGL options (partially or completely) when available OpenGL version doesn't meet requirements.

the `getsupportedRenderers()` function could be moved to another file & / or class, lemme knows which one.

Also fixed VSYNC being disabled until an OpenGL option is used & the dialog is re-opened, it should now update properly.

### DIFF
--- a/src/frontend/qt_sdl/VideoSettingsDialog.cpp
+++ b/src/frontend/qt_sdl/VideoSettingsDialog.cpp
@@ -41,13 +41,77 @@ void VideoSettingsDialog::setEnabled()
 {
     auto& cfg = emuInstance->getGlobalConfig();
     int renderer = cfg.GetInt("3D.Renderer");
+    int ogldisplay = cfg.GetBool("Screen.UseGL");
+
+    int supportedRenderer = getsupportedRenderers();
+
+    bool base_gl = supportedRenderer > renderer3D_Software;
+    bool compute_gl = supportedRenderer == renderer3D_OpenGLCompute;
+
+    if (!compute_gl)
+    {
+        ui->rb3DCompute->setEnabled(false);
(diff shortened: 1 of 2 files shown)


src/frontend/qt_sdl/VideoSettingsDialog.cpp
```diff
@@ -41,21 +41,93 @@ void VideoSettingsDialog::setEnabled()
...
+        if (glContext)
+        {
+            const char* gl_version_str = reinterpret_cast<const char*>(glGetString(GL_VERSION));
+
+            if (gl_version_str)
+            {
+                int gl_version = 0;
+                
+                // A proper version string or object isn't provided, so we have to parse it ourselves
+                if (isdigit(gl_version_str[0]) && isdigit(gl_version_str[2]))
```
asiekierka, line 103: Should this check the length of the string before accessing its third byte?

src/frontend/qt_sdl/VideoSettingsDialog.h
```diff
@@ -74,9 +74,12 @@ private slots:
 private:
     void setVsyncControlEnable(bool hasOGL);
     void setEnabled();
+    int getsupportedRenderers();
 
     Ui::VideoSettingsDialog* ui;
     EmuInstance* emuInstance;
+    bool base_gl;
```
asiekierka, line 81: These should probably be `baseGl` and `computeGl`, to match the camel case used throughout the codebase in such cases.

Mrcubix: I have seen multiple example of this case being used in the same codebase.

PR #2181, opened 2024-10-30
fix framerate not adjusting when it should slow/speedup due to vcount writes

Jaklyy:
my bad

### DIFF
--- a/src/frontend/qt_sdl/EmuInstance.cpp
+++ b/src/frontend/qt_sdl/EmuInstance.cpp
@@ -93,25 +93,25 @@ EmuInstance::EmuInstance(int inst) : deleting(false),
     if (val == 0.0)
     {
         Platform::Log(Platform::LogLevel::Error, "Target FPS in config invalid\n");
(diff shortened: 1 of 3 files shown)


src/frontend/qt_sdl/EmuInstance.cpp
```diff
@@ -93,25 +93,25 @@ EmuInstance::EmuInstance(int inst) : deleting(false),
     if (val == 0.0)
     {
         Platform::Log(Platform::LogLevel::Error, "Target FPS in config invalid\n");
-        targetFPS = 1.0 / 60.0;
```
RSDuck, line 96: if this variable contains the inverse of the framerate it should be called targetFrametime

RSDuck (approved):

src/frontend/qt_sdl/EmuInstance.cpp
RSDuck: ah no it is now right, it is FPS now and used to be frametime, everything's alright then

Issue #74, opened 2017-06-20
Tony Hawk's Motion (USA) missing motion pak


benderscruffy:
cannot play the game as it is mising the motion pak
![tony hawk s motion usa -0](https://user-images.githubusercontent.com/2696404/27332362-234ef4c0-55f4-11e7-8a92-9db7d1e2a3ca.png)

xperia64:
The Motion Pack used in this game contains a [Kionix 3-axis accelerometer](http://kionixfs.kionix.com/en/datasheet/KXSC4-2050%20Specifications%20Rev%203.pdf), and an [ATTiny13](http://ww1.microchip.com/downloads/en/devicedoc/doc2535.pdf) MCU.

For the following, I will be calling the VDD35 pin of the GBA slot Pin 1, and GND Pin 32:

- Pins 14 and 15 (AD8 and AD9) are pulled to ground on this pack, such that a 16-bit read of GBA ROM space should return 0xFCFF
  - This should make the game recognize the Pack up until the menus
- Pins 22 and 23 (A16 and A17) output 2-bit packets of data from the ATTiny13 to the NDS
- Pin 30 (/CS2) is used to clock the state of the ATTiny13
  - When the NDS reads from GBA SRAM space, the lower 2 bits of the data read are the 2-bit packet sent by the ATTiny13
  - When the SRAM space read completes and /CS2 is deasserted/pulsed, the ATTiny13 advances state and queues up the next 2 bits 
    - I'm not quite clear under what conditions /CS2 is asserted/deasserted by the DS, but doing multiple reads sequential reads in my homebrew did seem to clock the /CS2 line as needed

The protocol used by the ATTiny13 is a simple 17-state loop, where each state is responsible for outputting 2 bits:
- 0b00
- 0xFF over 4 states (big endian, upper 2 bits first; so (0xFF>>6)&3, then (0xFF>>4)&3, etc)
- 8-bit x-axis acceleration over 4 states
- 8-bit y-axis acceleration over 4 states
- 8-bit z-axis acceleration over 4 states

To convert an input acceleration in g's to the 8-bit units of this pack, use the following equation:
output_accel = min(0xFE, max(0, int((input_accel * 0.660V/g+1.65V) * 256 / 3.3V)))
These values were obtained from the datasheets of the chips.

The output on each accelerometer axis must be clamped between 0x0 and 0xFE inclusive, as 0xFF is likely reserved for synchronization.

Issue #755, opened 2020-09-12
Sonic Chronicles crashing on the first tutorial video


AbandonedWitchStarlight:
Sonic Chronicles: The Dark Brotherhood crashes on the first tutorial video where you touch the screen with touch points, and it happens after the first dialogue scene with Sonic and Tails.

Buenia0:
Hello, AbadonedWitchStarlight. My name is Buenia0, and I'm the author of a pull request to melonDS that adds support for the Guitar Grip and 2 other DS addons. Unfortunately, that pull request is still in the review stage, but if it gets merged, I will add support for the other addons as well.

Here's a link to that pull request:

https://github.com/Arisotura/melonDS/pull/719

Thanks for bringing this up!

PR #2183, opened 2024-10-31
Slot-2 Motion Pak, Guitar Grip emulation

asiekierka:
This PR adds emulation of the following Slot-2 devices:

* The homebrew Motion Pak, supporting X/Y/Z acceleration and Z rotation measurements. I have validated, with the help of a friend, the X/Y/Z acceleration support; I do not have any hardware which supports the Z rotation, however, so that's a bit of a guess.
  * The Motion Pak is mapped to accelerometer/gyroscope readouts from the connected SDL game controller; think DualSense or Switch Pro Controller.
* The official Motion Pack (sic!), as bundled with Tony Hawk's Motion, as described in https://github.com/melonDS-emu/melonDS/issues/74 . There's really no reason to use that one - it omits the gyroscope, has only 8 bits and not 12 bits of accuracy, and for some bizarre reason the game itself supports the homebrew pak too (!?). I have not tested it, as I don't own the game.
* The Guitar Grip. This is a "low-hanging fruit" with regards to adding other devices that support external inputs, such as the Paddle and Piano Controller. Tested, but only with a homebrew test ROM.

Closes https://github.com/melonDS-emu/melonDS/issues/74 and https://github.com/melonDS-emu/melonDS/issues/73 . Related to https://github.com/melonDS-emu/melonDS/issues/755 . Testing welcome.

### DIFF
--- a/src/GBACartMotionPak.cpp
+++ b/src/GBACartMotionPak.cpp
@@ -113,7 +113,8 @@ u8 CartMotionPakHomebrew::SRAMRead(u32 addr)
         return 0;
     case 8:
         // Read Z rotation
-        ShiftVal = RotationToMotionPak(Platform::Addon_MotionQuery(Platform::MotionRotationZ, UserData)) << 4;
+        // CHECKME: This is a guess, compare with real hardware
+        ShiftVal = RotationToMotionPak(-Platform::Addon_MotionQuery(Platform::MotionRotationZ, UserData)) << 4;
         return 0;
     case 10:
         // Identify cart
--- a/src/GBACartMotionPak.cpp
+++ b/src/GBACartMotionPak.cpp
@@ -161,7 +161,7 @@ void CartMotionPakRetail::DoSavestate(Savestate* file)
 
 u16 CartMotionPakRetail::ROMRead(u32 addr) const
 {
-    // CHECKME: Retail cartridge seems to return 0x7C7C.
+    // A9-A8 is pulled low on a real Motion Pack.
     return 0xFCFF;
 }
 
--- a/src/GBACartMotionPak.cpp
+++ b/src/GBACartMotionPak.cpp
@@ -1,0 +1,134 @@
+/*
+    Copyright 2016-2024 melonDS team
+
+    This file is part of melonDS.
+
+    melonDS is free software: you can redistribute it and/or modify it under
+    the terms of the GNU General Public License as published by the Free
+    Software Foundation, either version 3 of the License, or (at your option)
+    any later version.
+
+    melonDS is distributed in the hope that it will be useful, but WITHOUT ANY
+    WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+    FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License along
+    with melonDS. If not, see http://www.gnu.org/licenses/.
+*/
+
+#include <assert.h>
+#include "NDS.h"
+#include "GBACart.h"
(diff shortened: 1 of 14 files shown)


src/GBACartMotionPak.cpp
```diff
@@ -0,0 +1,196 @@
...
+void CartMotionPakRetail::DoSavestate(Savestate* file)
+{
+    CartCommon::DoSavestate(file);
+    file->Var8(&Value);
+    file->Var8(&Step);
+}
+
+u16 CartMotionPakRetail::ROMRead(u32 addr) const
+{
+    // CHECKME: Retail cartridge seems to return 0x7C7C.
```
xperia64, line 164: Per my comment on the issue, as long as bits 9 downto 8 are 0 (pulled down), the rest of the bits are "don't care"/open bus.

The value it actually returns may not be consistent due to open bus.

src/GBACartMotionPak.cpp
```diff
@@ -0,0 +1,196 @@
...
+        // Read Y acceleration
+        ShiftVal = AccelerationToMotionPak(Platform::Addon_MotionQuery(Platform::MotionAccelerationY, UserData)) << 4;
+        return 0;
+    case 6:
+        // Read Z acceleration
+        ShiftVal = AccelerationToMotionPak(Platform::Addon_MotionQuery(Platform::MotionAccelerationZ, UserData)) << 4;
+        return 0;
+    case 8:
+        // Read Z rotation
+        // CHECKME: This is a guess, compare with real hardware
```
xperia64, line 116: You may have a hard time here as I'm not entirely convinced the homebrew Slot2 motion pack was ever actually sold/made available. The only product I can find with a motion pack is supposedly the Neoflash "R6Gold".

All of my knowledge of the homebrew cart's spec comes from libnds.

asiekierka: > You may have a hard time here as I'm not entirely convinced the homebrew Slot2 motion pack was ever actually sold/made available.

Absolutely was. https://web.archive.org/web/20080509173200/http://www.ndsmotion.com/ has a list of stores and photos of the packaging.

There were six products in total which featured Kionix's technology, five of which were for the DS family of consoles; I maintain a list here: https://wiki.asie.pl/doku.php?id=notes:flashcart:ds_motion_card

I tested with an MK6-Motion myself, which happens to feature both the accelerometer and gyroscope chip; since the chips themselves and thus values output should be identical between the Card and a Pak, I deemed the emulation to be accurate enough for running homebrew (which is the important thing for me from a preservation perspective), but I can't verify fine details like the Pak's open bus behaviour or any additional SPI/ROM commands the Card might have.

> All of my knowledge of the homebrew cart's spec comes from libnds.

There's some additional bits of information in a README bundled with the original DS Motion Card/Pak driver source code on said website.

src/GBACartMotionPak.cpp
asiekierka: Thanks for the clarification!

src/GBACartMotionPak.cpp
asiekierka: (Either way, I don't expect this to be resolved as part of this PR, and the detection mechanism used by all homebrew for the Pak doesn't rely on open bus behaviour.)

asiekierka:
PR updated.

asiekierka:
PR rebased.

BueniaDev:
Thanks so much for implementing those specific add-ons, @asiekierka! I mean, I was going to implement those add-ons myself, but it looks like you just saved me quite a bit of work, so...yeah, thanks!

PR #2187, opened 2024-11-01
Use Qt 6 by default on all platforms and update build instructions

nadiaholmquist:
* Switch to using Qt 6 everywhere by default. `USE_QT6` always defaults to enabled
* Use UCRT on Windows
* Move build instructions to BUILD.md and
  * Add packages for Ubuntu 24.04 and Fedora
  * Switch Linux and Windows to Qt 6
  * Clean up Windows build instructions and mention ARM64 support
  * Mention the Nix flake


Mrlinkwii:
image

works fine on ubuntu 22.04

Issue #2096, opened 2024-07-12
Reset crashes MelonDS


ZLGNF:
Reset is causing crashes, sometimes immediately, sometimes you have to do something (before or after using Reset) that causes the emulator to pause (like going to the Inputs for example)

Streem22:
Having the same issue, did you fix it ?

RSDuck:
I cannot reproduce. Can you please send your config file (old ini and new toml?).

PR #2199, opened 2024-11-13
EmuInstanceAudio.cpp: Dont open new mic device on audioEnable()

SachinVin:
Fixes memory leak when resetting the emu
resetting the emu calls `audioEnable()`, creating a new `micDevice` every time without deleting the old one.
Maybe (partly) responsible for #2096


Arisotura:
I think the point was to avoid keeping the mic open when the emulator isn't running.

if that causes a memory leak, we ought to fix that, tho

PR #2200, opened 2024-11-15
add devtools options to break on data errors with GDB connected + flag to access full 8mb main memory in DS mode

joshua-smith-12:
I found these useful when I was debugging issues with a ROM which uses more than 4mb memory under specific optimization levels (at higher optimization levels it fits fine, but with optimizations disabled it needs more than 4mb total).

- the GDB `BreakOnError` flag can help to track down errors related to memory access/allocation
- the `Debug.ExtendedMemory` flag allows access to the full 8mb main memory range, instead of mirroring the lower 4mb to the upper 4mb

no retail DS provides access to the upper 4mb, but per gbatek there were hardware debugger machines which can use the full 8mb. this PR does not implement everything covered by the debugger machine, only allows access to more memory for debugging purposes. this does mean that the behaviour of the emulator with the 8mb flag enabled is a bit far from retail.

### DIFF
--- a/src/NDS.h
+++ b/src/NDS.h
@@ -469,6 +469,8 @@ public: // TODO: Encapsulate the rest of these members
     virtual void ARM7IOWrite16(u32 addr, u16 val);
     virtual void ARM7IOWrite32(u32 addr, u32 val);
 
+    void SetMemoryExtension(bool enabled);
+
 #ifdef JIT_ENABLED
     [[nodiscard]] bool IsJITEnabled() const noexcept { return EnableJIT; }
     void SetJITArgs(std::optional<JITArgs> args) noexcept;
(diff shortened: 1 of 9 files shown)


src/NDS.h
```diff
@@ -469,6 +469,8 @@ class NDS
     virtual void ARM7IOWrite16(u32 addr, u16 val);
     virtual void ARM7IOWrite32(u32 addr, u32 val);
 
+    void SetMemoryExtension(bool enabled);
```
asiekierka, line 472: Could this flag be extended to also work on TWL/DSi? Its debug units have 32MB of RAM as opposed to 16MB of RAM; it's available at 0xC000000 ~ 0xDFFFFFF.

PR #2221, opened 2024-11-26
Add support for using the solar sensor without requiring a Boktai ROM

JesseTG:
This pull request adds an API to the emulator that creates a stub ROM with just enough information to convince Lunar Knights that a solar sensor cart is inserted, allowing players to use the solar sensor without requiring a ROM from the Boktai trilogy. I have tested it in melonDS DS, and it works like a charm. The compressed logo data that's normally embedded in NDS and GBA carts is copied from the loaded NDS ROM in `NDS::Start` if the Boktai stub is used.

~~This PR (as of this writing) does not add this capability to the UI, but I can do so if you want me to.~~


JesseTG:
Whoops, I'd forgotten that I already added UI support. In that case, this PR is ready to merge.

PR #2234, opened 2024-12-13
ci: don't double zip mac .app

### LINKED ISSUE #2232: "macOS-universal.zip" is double-compressed
when unzipping macOS-universal.zip this just contains another similarly named zip file. the app is only inside the zip file contained in the first zip. why?

briaguya0:
resolves https://github.com/melonDS-emu/melonDS/issues/2232

https://github.com/actions/upload-artifact always zips whatever is put in `path` (doesn't matter if it's a file or directory or wildcard)

this gets a little complicated when dealing with `.app`s on macOS because `.app`s are just directories

this means if you just had

```yml
      - name: Upload artifact
        uses: actions/upload-artifact@v4
        with:
          name: macOS-universal
          path: melonDS.app
```

`upload-artifact` would treat `melonDS.app` as a directory and create a zip with the *contents* of the directory, so when someone unzips it they'd just get a `Contents` directory.

zipping the `.app` fixes that issue, but leads to the reported problem of double zips

this PR works around both issues by creating a new directory, moving the `.app` into that directory, and setting that directory as the `path` for the `upload-artifact` action

tested on my fork: https://github.com/briaguya-ai/melonDS/pull/1

### DIFF
--- a/.github/workflows/build-macos.yml
+++ b/.github/workflows/build-macos.yml
@@ -78,12 +78,13 @@ jobs:
           cp -a arm64/melonDS.app melonDS.app
           cp melonDS melonDS.app/Contents/MacOS/melonDS
           codesign -s - --deep melonDS.app
-          zip -r -y macOS-universal.zip melonDS.app
+          mkdir macOS-universal
+          mv melonDS.app macOS-universal
       - name: Upload artifact
         uses: actions/upload-artifact@v4
         with:
           name: macOS-universal
-          path: macOS-universal.zip
+          path: macOS-universal
 #     - name: Clean up architecture-specific artifacts
 #       uses: geekyeggo/delete-artifact@v4
 #       with:


nadiaholmquist:
I tested this myself earlier and the executable in the bundle did not end up marked as executable. Are you sure it actually is?

briaguya-ai:
> I tested this myself earlier and the executable in the bundle did not end up marked as executable.

interesting, i'm on linux (don't have a mac) so i didn't notice that (i just checked the contents of the zip file, didn't think to check perms)

based on https://github.com/actions/upload-artifact?tab=readme-ov-file#permission-loss it seems double zipping is the best way to preserve perms

> If you must preserve permissions, you can `tar` all of your files together before artifact upload. Post download, the `tar` file will maintain file permissions and case sensitivity.

 i'll go ahead and close this PR

nadiaholmquist:
This is now fixed properly by 5421e687fd16000c16371aed8968e540cfaad1f4.

PR #2245, opened 2024-12-23
Fix gdbstub not activating until the console is reset

ZNixian:
The check for initialising the gdbstub depending on whether the JIT was enabled or not was the wrong way around: previously, it would only enable the gdbstub if the JIT was enabled.

The stub started working again if you reset the console, as NDS::SetGdbArgs didn't have any such check and it was called by EmuInstance::updateConsole.


RSDuck:
Yeah that makes sense, thanks!

PR #2258, opened 2025-01-15
build: de-vendor xxHash dependency

Tachi107:
Hi!

This patch adds a small CMake Find Module which uses pkg-config to find the xxHash library.

De-vendoring dependencies (when it makes sense) is required for inclusion in some Linux distributions, like Debian.

Note that I've also added the dependency to the vcpkg manifest, but didn't test on Windows.

Bye :)


asiekierka:
From the Discord:

> this is kind of concerning to me now considering bizhawk port does not use cmake at all (it cant) and it's expected that the core have everything pinned down and not rely on distro deps

In other words, there are downstream projects of relevance to the melonDS community which rely on xxHash being vendored. Could an alternate solution be made where the distro-provided version of xxHash is *preferred*, but a fallback is provided by melonDS?

Paging @CasualPokePlayer

Tachi107:
Hi, thanks for the reply!

There are projects which build melonds without its build system? How? Can you point me to them, so that I can better understand the situation?

CasualPokePlayer:
In general it's best for an emulation core to not have external dependencies outside of language standard libaries, as to ease portability. Any external dependencies like xxhash can be vendored to have the same result. The emulation core can be seen as a platform agnostic library (with even the more platform specific parts of the core, such as the JIT and OpenGL renderer, are optional components). This differs from the entire project in general, including its frontend, as such is implicitly platform specific (and ports of melonDS with different frontends have their own handling of such, more or less independent of the core).

BizHawk's port of melonDS can be found [here](https://github.com/TASEmulators/BizHawk/tree/a2cf4e0/waterbox/melon). It's ported under a special method, called "waterbox." This is a very sandboxed like environment. For the purposes of porting, it can be understood as special Linux version where you do not have any system libraries to rely on outside of language standard libraries (dynamic loading is not possible in the box), everything is statically linked, including the C and C++ standard libraries (musl and libc++ in this case).

In practice too, even if this PR goes forward, it would be trivial for BizHawk to just vendor xxhash regardless and add an include path. Although, such would also be easy enough to do for melonDS anyways, leaving vendoring optional for build environments where there is no xxhash system library.

Tachi107:
> In general it's best for an emulation core to not have external 
> dependencies outside of language standard libaries, as to ease 
> portability.

I do agree with this, but also acknowledge that it's pretty much 
impossible with languages like C or C++ (e.g., crypto).

> Any external dependencies like xxhash can be vendored to have the same 
> result.

I don't really agree with this. In my experience, vendoring is 
a shortcut to proper dependency management.

> The emulation core can be seen as a platform agnostic library (with 
> even the more platform specific parts of the core, such as the JIT and 
> OpenGL renderer, are optional components).  This differs from the 
> entire project in general, including its frontend, as such is 
> implicitly platform specific (and ports of melonDS with different 
> frontends have their own handling of such, more or less independent of 
> the core).

Got it.

> BizHawk's port of melonDS can be found 
> [here](https://github.com/TASEmulators/BizHawk/tree/a2cf4e0/waterbox/melon).  
> It's ported under a special method, called "waterbox." This is a very 
> sandboxed like environment. For the purposes of porting, it can be 
> understood as special Linux version where you do not have any system 
> libraries to rely on (dynamic loading is not possible in the box), 
> everything is statically linked, including the C and C++ standard 
> libraries (musl and libc++ in this case).

Interesting project! Thanks for sharing :)

> In practice too, even if this PR goes forward, it would be trivial for 
> BizHawk to just vendor xxhash regardless and add an include path.  
> Although, such would also be easy enough to do for melonDS anyways, 
> leaving vendoring optional for build environments where there is no 
> xxhash system library.

I don't want to make things harder for other downstream projects; after 
all, I'm a downstream as well!

Still, I don't fully understand what the requirements of BizHawk are.

You are of course free to statically link to a library in many ways 
which do not require vendoring:

- Linking with the system's provided static library (in this case, 
  libxxhash.a)
- Using CMake's FetchContent modules, which nicely integrates with 
  find_package(): https://cmake.org/cmake/help/latest/module/FetchContent.html
- Using the vcpkg package manager, which melonDS already uses.

Are any of the above options applicable for BizHawk?

Reading the previous messages and BizHawk's source code, I gather that 
upstream's build system isn't used at all. Having worked with countless 
upstream projects when packaging stuff, this seems a nightmare to me! 
You'd have to constantly track the CMake build scripts to avoid having 
your makefiles get out of date, right? Moreover, what does this imply 
for melonDS freedom of changing "core" stuff?

If none of the above solutions work for you, and vendoring truly is the 
only solution, than it'd make sense to also document this requirement 
somewhere here, like: "When adding a dependency to the emulation core, 
it MUST be vendored so that certain downstream projects like BizHawk can 
build it without troubles. The vendored dependency though SHOULD also be 
added as a regular imported CMake dependency, and the vendored code 
SHOULD only be used as a fallback".

Does this make sense?

Issue #2264, opened 2025-01-21
Incorrect emulation of SLOT-2 GPIO


edo9300:
Currently melon handles the SLOT-2 GPIO by returning `0` whenever the 3 mapped addresses are attempted to be read if GPIO is disabled, 
https://github.com/melonDS-emu/melonDS/blob/15c3faa26e879bdcff615558ded6dd886681ccae/src/GBACart.cpp#L234-L247
This doesn't match hardware where reads to those addresses correctly return the value stored in the card's rom if the specific card doesn't have the addresses matched to a gpio interface (for example a generic flashable cart or a supercard flashed with a custom firmware running in ds mode).
It should be investigated if the behaviour is entirely on the game pack itself or there's something else being done by the ds itself

PR #2266, opened 2025-01-24
Don't return 0 when attempting to read from the GPA GPIO addresses with GPIO disabled

edo9300:
Currently melon would return `0` if something is attempted to be read from the addresses 0x080000C4-0x080000C8, but gba roms can and do have data in that place, while games using custom gpio hardware have those bytes set to 0 since they're mapped to the gpio lines.
Now it will instead consider those addresses as special only if the game explicitly attempted to enable gpio by writing to the control address.

Fixes https://github.com/melonDS-emu/melonDS/issues/2264


asiekierka:
LGTM

PR #2294, opened 2025-03-25
Change the save type for Kantan Ongaku Ryoku (Japan) to EEPROM 8192

mrhso:
The European version and the Japanese version of the program are completely different (as evidenced, by the ROM size), and their save data structures also differ. The Japanese version uses no more than 8192 bytes of data, it is more likely to be EEPROM 8192.


Arisotura:
I guess we could safely switch it to EEPROM 8K, but it'd be nice if it could be confirmed

PR #2317, opened 2025-05-11
dont waste 12MB of savestate size

SuuperW:
NDS was writing the entire allocated main RAM buffer to savestates. The buffer is always allocated for DSi mode, and this leads to savestates being over twice the size they need to be. Now it writes an amount appropriate for the console type.

### DIFF
--- a/src/NDS.cpp
+++ b/src/NDS.cpp
@@ -630,7 +630,7 @@ bool NDS::DoSavestate(Savestate* file)
         }
     }
 
-    file->VarArray(MainRAM, MainRAMMaxSize);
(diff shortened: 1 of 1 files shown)


Jaklyy:
re-using that variable for this could cause problems in the future when the ability to toggle main ram size is re-enabled in dsi mode.
ofc that assumes that the upper 12MiB persist while in 4MiB mode (they probably aren't immediately wiped at the very least?)

SuuperW:
I didn't know that was a thing. If that causes issues, perhaps there should be a new variable? Or savestates could choose how much to save based on console type? I obviously don't know what the best solution is.

PR #2332, opened 2025-05-21
"Hybrid ratio", new option on View menu for Hybrid screen layout

KDOXG:
## Objective

Adjust how the window works in hybrid screen layout to allow larger screen sizes for the main screen by creating a "Hybrid ratio" setting menu by extending the "Screen gap" setting, and to always display at a pixel-perfect integer scale when the "Force integer scaling" option, located in the "Screen sizing" menu, is enabled.

## Disclaimer

I will refer to the DS top screen as screen 1, the DS bottom screen as screen 2, and the larger screen in hybrid layout as the main screen. This naming convention is used consistently throughout this explanation.  
I will not cover behavior when "Force integer scaling" is disabled. I also will not consider a possible future implementation of flipped hybrid layout, as discussed in PR #1705; whoever implements that should simply mirror how the current hybrid layout works.

As reference, I reviewed how other emulators handle hybrid layout configurations. The libretro port of melonDS, melonDS DS, has a similar setting, but its screen gap only works in vertical mode. Since standalone melonDS offers more flexible hybrid layout options, I won't base this proposal on melonDS DS implementation — although it would be great to bring this to that version as well in the future.

Similarly, Bizhawk’s melonDS core has flexibility comparable to standalone melonDS, but for some reason its screen gap setting distorts the main screen, producing uneven pixels even at what should be a pixel-perfect aspect ratio. I’m not sure why, but since the Bizhawk core doesn’t run well on my machine (at least in the official release build), I’ve chosen to disregard it for now.

Lastly, some feature requests suggest increasing the hybrid layout screen ratio without using screen gaps, centering the smaller screens beside the main screen. That would require additional frontend positioning changes, and is outside the scope of this proposal. However, this PR is a step toward enabling that functionality in the future.

## Problem

Currently, hybrid layout defaults to a 2:1 ratio, where the main screen is twice the size of screens 1 and 2. These two smaller screens are positioned to the right of the main screen in vertical layout.

However, melonDS uses a unified window system. So, a 1x hybrid layout window is treated as a single 768×384 surface — equivalent to a 3:2 ratio in DS screen units.

When the "Screen gap" setting is increased, a gap is introduced between screens 1 and 2, and the main screen is scaled to align:
- its top edge with the top edge of screen 1, and  
- its bottom edge with the bottom edge of screen 2,  
while preserving its 4:3 aspect ratio.

This causes the overall screen height to grow by the gap value. Because of the unified window system, this also increases the width, stretching the main screen and breaking integer scaling.

When "Force integer scaling" is enabled, screens 1 and 2 correctly retain integer scaling. However, the main screen is scaled according to the fixed 4:3 aspect ratio and the screen gap, which causes it to render at non-integer scale values in the current implementation — even with integer scaling enabled.

## Solution

I propose removing the influence of the screen gap setting while in hybrid layout and adding a hybrid ratio setting. This new configuration would appear as a selectable list of screen ratios and can be implemented by cloning the existing "Screen gap" setting, while internally mapping each option to a predefined gap value based on DS screen units (192 pixels tall at 1x).

Specifically:
- Values ≥192 allow ratios between 3:1 and (2 + screen gap / 192):1
- Values <192 allow ratios like (2x + 1):x (e.g., 5:2, 7:3)

### Proposed ratio options:
Ratio | Screen gap | Final resolution
--- | --- | ---
2:1 | 0 | 768x384
3:1 | 192 | 1024x576
4:1 | 384 | 1280x768
5:1 | 576 | 1536x960
6:1 | 768 | 1792x1152
7:1 | 960 | 2048x1344
5:2 | 96 | 1792x960
7:3 | 64 | 2560x1344
9:4 | 48 | 3328x1728

You can test these options right now in the current build. After running the emulator once, open your `melonDS.toml` file (located in the exe folder) and manually set the `ScreenGap` value under the `[Instance.Window]` section. The value will be used regardless of what is displayed in the UI.

Since it's just a modification in the front end, I’ve only edited the code related to the emulator window. The emulator now includes a new menu under **View → Hybrid ratio**, directly below the **Screen gap** menu. As mentioned earlier, these values override the **Screen gap** setting only when the **Screen layout** is set to **Hybrid**.

Note: Due to hardware limitations (I only have a 1080p display), I could not fully test the 6:1, 7:1, 7:3, and 9:4 options on my system.

### DIFF
--- a/src/frontend/qt_sdl/Config.cpp
+++ b/src/frontend/qt_sdl/Config.cpp
@@ -85,6 +85,7 @@ RangeList IntRanges =
     {"Mic.InputType", {0, micInputType_MAX-1}},
     {"Instance*.Window*.ScreenRotation", {0, screenRot_MAX-1}},
     {"Instance*.Window*.ScreenGap", {0, 500}},
+    {"Instance*.Window*.HybridRatio", {0, 960}},
     {"Instance*.Window*.ScreenLayout", {0, screenLayout_MAX-1}},
     {"Instance*.Window*.ScreenSizing", {0, screenSizing_MAX-1}},
     {"Instance*.Window*.ScreenAspectTop", {0, AspectRatiosNum-1}},
--- a/src/frontend/qt_sdl/Screen.cpp
+++ b/src/frontend/qt_sdl/Screen.cpp
@@ -118,13 +118,16 @@ void ScreenPanel::loadConfig()
     auto& cfg = mainWindow->getWindowConfig();
     
     screenRotation = cfg.GetInt("ScreenRotation");
-    screenGap = cfg.GetInt("ScreenGap");
     screenLayout = cfg.GetInt("ScreenLayout");
+    screenGap = cfg.GetInt("ScreenGap");
+	hybridRatio = cfg.GetInt("HybridRatio");
     screenSwap = cfg.GetBool("ScreenSwap");
     screenSizing = cfg.GetInt("ScreenSizing");
     integerScaling = cfg.GetBool("IntegerScaling");
     screenAspectTop = cfg.GetInt("ScreenAspectTop");
     screenAspectBot = cfg.GetInt("ScreenAspectBot");
+
+	currentScreenGap = screenLayout != screenLayout_Hybrid ? screenGap : hybridRatio;
 }
 
 void ScreenPanel::setFilter(bool filter)
@@ -168,7 +171,7 @@ void ScreenPanel::setupScreenLayout()
                 static_cast<ScreenLayoutType>(screenLayout),
                 static_cast<ScreenRotation>(screenRotation),
                 static_cast<ScreenSizing>(sizing),
-                screenGap,
+                currentScreenGap,
                 integerScaling != 0,
                 screenSwap != 0,
                 aspectTop,
@@ -183,7 +186,7 @@ QSize ScreenPanel::screenGetMinSize(int factor = 1)
 {
     bool isHori = (screenRotation == screenRot_90Deg
         || screenRotation == screenRot_270Deg);
-    int gap = screenGap * factor;
+    int gap = currentScreenGap * factor;
 
     int w = 256 * factor;
     int h = 192 * factor;
--- a/src/frontend/qt_sdl/Screen.h
+++ b/src/frontend/qt_sdl/Screen.h
@@ -82,11 +82,14 @@ protected:
 
     int screenRotation;
     int screenGap;
+	int hybridRatio;
     int screenLayout;
     bool screenSwap;
     int screenSizing;
     bool integerScaling;
     int screenAspectTop, screenAspectBot;
+
+	int currentScreenGap;
 
     int autoScreenSizing;
 
--- a/src/frontend/qt_sdl/Window.cpp
+++ b/src/frontend/qt_sdl/Window.cpp
@@ -512,6 +512,24 @@ MainWindow::MainWindow(int id, EmuInstance* inst, QWidget* parent) :
                 connect(grpScreenGap, &QActionGroup::triggered, this, &MainWindow::onChangeScreenGap);
             }
             {
+                QMenu * submenu = menu->addMenu("Hybrid ratio");
+                grpHybridRatio = new QActionGroup(submenu);
+
+				const char *hybridRatio[] = {"2:1", "3:1", "4:1", "5:1", "6:1", "7:1", "5:2", "7:3", "9:4"};
+                const int screengap[] = {0, 192, 384, 576, 768, 960, 96, 64, 48};
+
+                for (int i = 0; i < 9; i++)
+                {
+                    int screenGapData = screengap[i];
+                    actHybridRatio[i] = submenu->addAction(QString(hybridRatio[i]));
+                    actHybridRatio[i]->setActionGroup(grpHybridRatio);
+                    actHybridRatio[i]->setData(QVariant(screenGapData));
+                    actHybridRatio[i]->setCheckable(true);
+                }
+
+                connect(grpHybridRatio, &QActionGroup::triggered, this, &MainWindow::onChangeHybridRatio);
+            }
+            {
                 QMenu * submenu = menu->addMenu("Screen layout");
                 grpScreenLayout = new QActionGroup(submenu);
 
@@ -746,6 +764,16 @@ MainWindow::MainWindow(int id, EmuInstance* inst, QWidget* parent) :
             if (actScreenGap[i]->data().toInt() == screenGap)
             {
                 actScreenGap[i]->setChecked(true);
+                break;
+            }
+        }
+
+        int hybridRatio = windowCfg.GetInt("HybridRatio");
+        for (int i = 0; i < 9; i++)
+        {
+            if (actHybridRatio[i]->data() == hybridRatio)
+            {
+                actHybridRatio[i]->setChecked(true);
                 break;
             }
         }
@@ -2033,6 +2061,14 @@ void MainWindow::onChangeScreenGap(QAction* act)
 {
     int gap = act->data().toInt();
     windowCfg.SetInt("ScreenGap", gap);
+
+    emit screenLayoutChange();
+}
+
+void MainWindow::onChangeHybridRatio(QAction* act)
+{
+    int gap = act->data().toInt();
+    windowCfg.SetInt("HybridRatio", gap);
 
     emit screenLayoutChange();
 }
--- a/src/frontend/qt_sdl/Window.h
+++ b/src/frontend/qt_sdl/Window.h
@@ -217,6 +217,7 @@ private slots:
     void onChangeScreenSize();
     void onChangeScreenRotation(QAction* act);
     void onChangeScreenGap(QAction* act);
+    void onChangeHybridRatio(QAction* act);
     void onChangeScreenLayout(QAction* act);
     void onChangeScreenSwap(bool checked);
     void onChangeScreenSizing(QAction* act);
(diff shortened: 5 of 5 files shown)


crashGG:
Could add a custom mode for screen layout? Sometimes users want to display dual screen content side by side, but want the main screen to be as large as possible while maintaining integer scaling, and the secondary screen to use the remaining space while also maintaining integer scaling?

An example from  Azahar:


In this way, the main screen can be set to 2816x2112 resolution (11x integer scaling), and the secondary screen can be set to 1027x768 (4x integer scaling)

KDOXG:
> Could add a custom mode for screen layout? Sometimes users want to display dual screen content side by side, but want the main screen to be as large as possible while maintaining integer scaling, and the secondary screen to use the remaining space while also maintaining integer scaling?
> 
> An example from Azahar: 
> 
> In this way, the main screen can be set to 2816x2112 resolution (11x integer scaling), and the secondary screen can be set to 1024x768 (4x integer scaling)

To do that, the user would need to set a value through the interface, and with that, the system would need to first apply a scale factor to the main screen, then use the remaining space to calculate and apply a different scale to the smaller screens. Directly setting scale factors for the screen is outside the scope of changes I made, as my focus was specifically on adjusting screen ratios. That said, it’s definitely an idea worth exploring in the future, but considering the current state of the interface, I'm not sure how to design where to put these scale factor values for the user to set. Perhaps an option in the same menu as the hybrid ratios, and then rename it to "Hybrid options".

crashGG:
I tested this PR, it looks good, but the gap is too big. Could you make the gap option in the menu work?

crashGG:
After testing it for a while, I found this feature to be great. Could you port the ratio to 2-screen-emphasize? Currently, if  use 2-screen-emphasize, the ratio of the two windows cannot be controlled.

KDOXG:
> After testing it for a while, I found this feature to be great. Could you port the ratio to 2-screen-emphasize? Currently, if use 2-screen-emphasize, the ratio of the two windows cannot be controlled.

Is `2-screen-emphasize` supposed to be a branch / PR or an actual feature for current melonDS? I searched everywhere and couldn't find anything by that name, but I'm willing to port this implementation if it can help spreading the functionality as a whole.

crashGG:
> > After testing it for a while, I found this feature to be great. Could you port the ratio to 2-screen-emphasize? Currently, if use 2-screen-emphasize, the ratio of the two windows cannot be controlled.
> 
> Is `2-screen-emphasize` supposed to be a branch / PR or an actual feature for current melonDS? I searched everywhere and couldn't find anything by that name, but I'm willing to port this implementation if it can help spreading the functionality as a whole.

2-screen-emphasize，it is not a feature, but an effect. It is actually a combination of two feature.
Firt use Screen layout - horizontal / virtical ,now you have 2 screens (horizontally/vertically depending on the  your physical monitor).
Second ,use Screen sizing - emphasize top/bottom (top/bottom which screen you want to be the primary screen ).
The final effect looks like this:

I personally think this is the most comfortable way to play on modern computers.

KDOXG:
> > > After testing it for a while, I found this feature to be great. Could you port the ratio to 2-screen-emphasize? Currently, if use 2-screen-emphasize, the ratio of the two windows cannot be controlled.
> > 
> > 
> > Is `2-screen-emphasize` supposed to be a branch / PR or an actual feature for current melonDS? I searched everywhere and couldn't find anything by that name, but I'm willing to port this implementation if it can help spreading the functionality as a whole.
> 
> 2-screen-emphasize，it is not a feature, but an effect. It is actually a combination of two feature. Firt use Screen layout - horizontal / virtical ,now you have 2 screens (horizontally/vertically depending on the your physical monitor). Second ,use Screen sizing - emphasize top/bottom (top/bottom which screen you want to be the primary screen ). The final effect looks like this:  I personally think this is the most comfortable way to play on modern computers.

Oh, understood! That would be great to have a ratio list as well. Analyzing here, the difference is that now there are only 2 screens, instead of having a third that acts like a clone, and the gap is added between the main screen (the emphasized one) and the remaining screen. I think I can tweak the math to work here as well, though first I need to understand the scaling behaviour occurring on Emphasize layout mode with integer scaling on. Give me some time and I'll reply here with what I can discover.

KDOXG:
> > > > After testing it for a while, I found this feature to be great. Could you port the ratio to 2-screen-emphasize? Currently, if use 2-screen-emphasize, the ratio of the two windows cannot be controlled.
> > > 
> > > 
> > > Is `2-screen-emphasize` supposed to be a branch / PR or an actual feature for current melonDS? I searched everywhere and couldn't find anything by that name, but I'm willing to port this implementation if it can help spreading the functionality as a whole.
> > 
> > 
> > 2-screen-emphasize，it is not a feature, but an effect. It is actually a combination of two feature. Firt use Screen layout - horizontal / virtical ,now you have 2 screens (horizontally/vertically depending on the your physical monitor). Second ,use Screen sizing - emphasize top/bottom (top/bottom which screen you want to be the primary screen ). The final effect looks like this:  I personally think this is the most comfortable way to play on modern computers.
> 
> Oh, understood! That would be great to have a ratio list as well. Analyzing here, the difference is that now there are only 2 screens, instead of having a third that acts like a clone, and the gap is added between the main screen (the emphasized one) and the remaining screen. I think I can tweak the math to work here as well, though first I need to understand the scaling behaviour occurring on Emphasize layout mode with integer scaling on. Give me some time and I'll reply here with what I can discover.

It's done, here is what I discovered.

**Notes:**

- With *integer scale* enabled, both the main and secondary screens are always guaranteed to be using integer scaling. Unlike what's done for the *hybrid ratio* solution, the tweaks for this feature related to the screen gap are only aimed at allowing the user to choose different scaling options for the screens.
- In emphasize mode, the priority is to keep the main screen as large as possible. As the screen gap increases, if the screens begin to touch the edges of the window, the secondary screen decreases its scale. If the screen gap increases further and the screens are still touching the edges but the secondary screen is already at 1x scale, then the main screen begins to reduce its scale.
- The screen gap is also scaled along with the screens: half of it's value is scaled using the main screen's scale, and the other half with the secondary screen's scale. The final screen gap displayed in the window is the sum of those two scaled values. As such, this final screen gap can also be determined by the window resolution and the screen orientation. Given the behavior described above for emphasize mode, **it's possible to determine a screen gap that achieves the desired scale for the screens**.

---

**Conclusions:**

- A screen gap of 0 already guarantees that both screens will be displayed at the largest possible size. Any change from that will only reduce the scale of at least one of the screens.
- Because the main screen tends to retain the highest possible scale, and since the final screen gap is also defined by the window resolution, it's not possible to have a global configuration that accommodates a list of ratios for this *2-screen-emphasize* mode. Such a list would not work for all window resolutions.

I can think of a few possible solutions that don’t involve creating a global configuration — like having ratio lists with options based on the most common window resolutions — but personally, I don't find the option of reducing scale very appealing, especially since integer scale is already achieved with the mode enabled. As for the example shown in the screenshot, setting a screen gap that reaches a comfortable spacing seems like a highly subjective preference — very different from ensuring integer scale, which is objective: either it is or it isn't.

I think the only interesting solution I can think of would be to have a **toggleable option** that makes the width of the black area between the screens (i.e., the final screen gap) equal to the width of the black area between each screen and the window edges. To implement this, it would be necessary to modify the layout setup method to ignore the input screen gap value and recalculate it based on the window resolution. This would even allow a degree of control over the screens' scaling behavior by resizing the window during runtime. However, the comfort of using this feature would still be subjective. I think more user feedback would be needed to validate adding such an option.

As for myself, personally, I'd prefer something different. Specifically, I’d like an option for *2-screen-emphasize* mode (or even hybrid!) that would center the main screen completely in the window — similar to *top/bottom screen only* mode — but with the secondary screen still appearing somewhere on the side. This would differ from the current default approach, where melonDS centers the union of both screens (plus the screen gap), leaving each screen slightly off-center. Targeting widescreen windows, the way I see, this solution would work very nice. However, implementing it would require deeper changes to the layout setup method, specifically changing how screen position offsets are calculated. I’m not planning to touch those calculations at the moment.

All these ideas are very cool to explore, and I really enjoy the effort to find better ways to present the DS screens on a single screen. Do you happen to know anyone currently researching layouts for DS and 3DS screens (or even Wii U, assuming there are efforts to reproduce the gamepad screen) so we could use those ideas as reference?

JoJoKorok:
Hi @KDOXG,

I tested your hybrid-ratio implementation on the current melonDS `master` and incorporated it into [PR #2705](https://github.com/melonDS-emu/melonDS/pull/2705), alongside an additional screen-layout improvement addressing [issue #2695](https://github.com/melonDS-emu/melonDS/issues/2695).

Your implementation remains a separate commit with your original Git author attribution preserved. The new pull-request description also explicitly credits you as the original author and credits @crashGG for their testing and feedback in this discussion.

The combined PR brings the hybrid-ratio implementation up to the current `master`, adds configuration and naming convention cleanup, and adds fill-window stretching for two-screen layouts.

Please let me know if you have any concerns about this approach or would prefer any changes to the attribution or integration of your work. Thank you for the original implementation.

KDOXG:
> Hi @KDOXG,
> 
> I tested your hybrid-ratio implementation on the current melonDS `master` and incorporated it into [PR #2705](https://github.com/melonDS-emu/melonDS/pull/2705), alongside an additional screen-layout improvement addressing [issue #2695](https://github.com/melonDS-emu/melonDS/issues/2695).
> 
> Your implementation remains a separate commit with your original Git author attribution preserved. The new pull-request description also explicitly credits you as the original author and credits @crashGG for their testing and feedback in this discussion.
> 
> The combined PR brings the hybrid-ratio implementation up to the current `master`, adds configuration and naming convention cleanup, and adds fill-window stretching for two-screen layouts.
> 
> Please let me know if you have any concerns about this approach or would prefer any changes to the attribution or integration of your work. Thank you for the original implementation.

Hello @JoJoKorok,

Thank you for reaching out to inform me. I don't have any concerns regarding your approach. I'm fine with the original commit being preserved and the credits being explicit in the PR. Please keep me informed when your PR gets merged with master so this one can be closed as well.

Good luck!

PR #2381, opened 2025-07-23
Make CameraManager request the camera access permission when using Qt 6

nadiaholmquist:
Camera access seems to have changed to require explicitly asking for permission rather than doing it for you.

Starting the session was moved to the camStart slot so that the permission prompt will show up when the camera is actually used the first time rather than in some cases before the main window even shows up.


asiekierka:
LGTM, though I can't really test it myself, I think?

asiekierka:
"Camera Settings" seems to work no worse than it did on my Linux machine (it can capture the camera), at least.

PR #2387, opened 2025-07-29
Fix SD multiblock writes and FIFO32 writes

shinyquagsire23:
Caused fwrite to hang with BlocksDS homebrew, existing logic seems to work fine afaict

### DIFF
--- a/src/DSi_SD.cpp
+++ b/src/DSi_SD.cpp
@@ -389,7 +389,7 @@ u32 DSi_SDHost::DataTX(u8* data, u32 len)
         {
             if (DataFIFO32.IsEmpty())
             {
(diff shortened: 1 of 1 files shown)


Arisotura:
there has been another PR fixing the DSi SD card stuff, so is this one still required? I think not

(sorry for it taking so long btw... I was wary of anything touching the SD/MMC code, knowing how much trouble I had getting it working in the first place, I was concerned any changes would fix one thing and break another)

shinyquagsire23:
I can check later probably, if it works with anything BlocksDS that's all that I was going for really

Arisotura:
sure, let us know!

PR #2388, opened 2025-07-31
fix: GDB stub reading wrong code for vCont packet

jonko0493:
I was attempting to use the gdb stub with lldb (in its gdb-remote mode) and noticed that the debugging session was always abruptly terminated whenever attempting to step or continue after a breakpoint was hit. It turns out that lldb sends the `vCont` packet for these actions, e.g. `vCont;s:1`. This is the correct format for this command as seen in the gdb documentation [here](https://sourceware.org/gdb/current/onlinedocs/gdb.html/Packets.html#vCont-packet) -- specifically, it indicates continue, step, thread 1.

However, the current gdb stub implementation was reading the `;` as the continue action code, assuming it was invalid, and then throwing an error. To me, this just seems like it should have never worked, but I'm also not super familiar with gdb's internals so I've fixed the issue without introducing a breaking change.

### DIFF
--- a/src/debug/GdbCmds.cpp
+++ b/src/debug/GdbCmds.cpp
@@ -893,7 +893,10 @@ ExecResult GdbStub::Handle_v_Cont(GdbStub* stub, const u8* cmd, ssize_t len)
 		return ExecResult::Ok;
 	}
 
-	switch (cmd[0])
+	// Continue code should always be separated by a ';' (but maybe they aren't sometimes? not sure)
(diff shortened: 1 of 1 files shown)


RSDuck:
The current code does seem to contain a mistake, please handle it so that only the correct behaviour is accepted. It was previously not spotted because gdb uses `c` to continue.

jonko0493:
Sounds great, will change accordingly.

PR #2400, opened 2025-08-10
Fix FF_ReadStorage/FF_WriteStorage callback return value

CasualPokePlayer:
The return value for these callbacks is in sectors, not bytes.

### DIFF
--- a/src/FATStorage.cpp
+++ b/src/FATStorage.cpp
@@ -192,14 +192,16 @@ u64 FATStorage::GetSectorCount() const
 ff_disk_read_cb FATStorage::FF_ReadStorage() const noexcept
 {
     return [this](BYTE* buf, LBA_t sector, UINT num) {
-        return ReadSectorsInternal(File, FileSize, sector, num, buf);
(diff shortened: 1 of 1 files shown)


Arisotura:
I'd have made Read/WriteSectorsInternal() return a number of sectors directly, unless something else relies on them returning bytes

but this works too

edo9300:
This change caused the issue https://github.com/melonDS-emu/melonDS/issues/2416

edo9300:
More specifically, the function already returned the number of sectors read, since down the line it called `fread(data, size, count,...)` with `size` being 0x200 and count being the number of sectors, and fread returns the number of blocks of size `size` read

Arisotura:
well I guess we ought to revert this then

CasualPokePlayer:
Ah, the issue would then just happen in BizHawk's platform function since it wasn't strictly using fread semantics for the return value, but rather just returning the number of bytes always (apparently fread returns the number of objects??? That's a fun footgun)

Arisotura:
indeed... that even confused me, which was why I merged this in the first place

PR #2401, opened 2025-08-10
Windows: Quick and Dirty Build Fix

GABO14231:
GCC has been updated to 15.2.0, yet the build error is still a thing. Not sure if a better long-term solution is desired instead, but for now we can update the conditional and it should build again. This was tested with a local build.


nadiaholmquist:
Maybe we should just make that condition match any GCC 15 version. It can be reverted when the issue is actually fixed in GCC.

GABO1423:
Done, now any GCC 15 version should be covered. I tested this with static Qt5/Qt6 (waiting on the latter was why I took a bit) local builds.

PR #2412, opened 2025-08-21
Add high-quality resampling using blip-buf to the core

nadiaholmquist:
This also gets rid of the need for the frontend to resample the audio, as it can simply ask the core to produce audio at whatever sample rate it desires.

### DIFF
--- a/src/SPU.cpp
+++ b/src/SPU.cpp
@@ -271,8 +271,6 @@ void SPU::DoSavestate(Savestate* file)
     file->Var8(&MasterVolume);
     file->Var16(&Bias);
 
-    file->Var8(&OutputSamplePos);
-    file->Var8(&OutputSampleInc);
     file->VarArray(OutputLastSamples, sizeof(OutputLastSamples));
 
     file->Var32(&MixInterval);
@@ -298,15 +296,12 @@ void SPU::SetSampleRate(AudioSampleRate rate)
     if (rate == AudioSampleRate::_47KHz)
     {
         MixInterval = 704;
-        OutputSampleInc = 16;
     }
     else
     {
         MixInterval = 1024;
-        OutputSampleInc = 11;
     }
 
-    OutputSamplePos = 0;
     memset(OutputLastSamples, 0, sizeof(OutputLastSamples));
 }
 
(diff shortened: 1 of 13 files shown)


src/SPU.cpp
```diff
@@ -993,15 +1003,34 @@ void SPU::Mix(u32 spucycles)
...
-    {
-        u8 fb = OutputSamplePos & 0xF;
-        u8 fa = 0x10 - fb;
-        s16 leftfinal = ((OutputLastSamples[0] * fa) + (output[0] * fb)) >> 4;
-        s16 rightfinal = ((OutputLastSamples[1] * fa) + (output[1] * fb)) >> 4;
+    if (output[0] != OutputLastSamples[0])
+        blip_add_delta(BlipLeft, BlipTimer, (int) output[0] - OutputLastSamples[0]);
+    if (output[1] != OutputLastSamples[1])
+        blip_add_delta(BlipRight, BlipTimer, (int) output[1] - OutputLastSamples[1]);
+    BlipTimer += MixInterval >> 1;
```
CasualPokePlayer, line 1010: You should be using spucycles here, MixInterval is the "new" interval for the next event, spucycles is the "current" interval for this event.

Arisotura:
lemme know when this is ready to go and we can merge it

it seems pretty promising

nadiaholmquist:
I think it should be good to go.

Arisotura:
you're in charge of this PR, you can un-draft it

v-fox:
Unfortunately, master-build with this has completely broke sound for me and made it super-choppy.
I even tried to play with values for `audioBufSize` & `newBufferSize = [256-1152]` and `SDL_AUDIODRIVER=pulseaudio|pipewire` & `PIPEWIRE_QUANTUM=[576-2304]/48000` but it did not make it any better. Tried switching vsync and frame limiting too.
Reverting merged patch fixed it.

I run RT-enabled kernel, KDE on wayland and pipewire & pipewire-pulse with default minimal buffer of 24 ms (`pulse.min.req = 256/48000` & `pulse.min.quantum = 1152/48000`) + easyeffects (bypassing did not change it). Usually 12 ms (`576/48000`) is safe too only on "thin" apps, for emulators and games 24-48 ms is the sweet-spot and with 3 audio-frames per buffer.
But here it's probably problem of timing for swapping/filling buffer rather than its size.

PR #2417, opened 2025-08-23
Implement Screenshot Hotkey

legoabram:
At the request of my brother, I implemented a basic screenshot mechanism into MelonDS. This just registers a new hotkey you can assign which will save the current frame buffer to a file at the native resolution of the renderer when the button press is registered. I've gotten this to functional for all graphic modes, and I've tested it on both Windows and MacOS with a couple of users from a nuzlocke forum who helped verify that it works correctly.

I'm happy to make adjustments as you need to better fit the architecture of the project, though I did my best to try and match it.

The last feature that was requested which I haven't yet added is a way to customize the actual filename. My basic thought process is that I'd add a simple templating structure, (Like %... in python, mmmm-DD-YY with strftime, etc.) and a text box that the user can edit to change this format string. Biggest issue is that I couldn't figure out where I would put such a dialog without adding a whole new window, so I want to ask you where this would make sense to go.

### DIFF
--- a/src/GPU.h
+++ b/src/GPU.h
@@ -861,6 +861,7 @@ public:
     // if the renderer uses RAM buffers, they should be 32-bit BGRA, 256x192 for each screen
     virtual bool GetFramebuffers(void** top, void** bottom) = 0;
     virtual void SwapBuffers() { BackBuffer ^= 1; }
+    virtual int  GetScaleFactor() { return Rend3D->GetScaleFactor(); };
 
     virtual bool NeedsShaderCompile() { return false; }
     virtual void ShaderCompileStep(int& current, int& count) {}
--- a/src/GPU3D.h
+++ b/src/GPU3D.h
@@ -332,6 +332,7 @@ public:
     virtual void RenderFrame() = 0;
     virtual void FinishRendering() {}
     virtual void RestartFrame() {};
+    virtual int GetScaleFactor() const noexcept { return 1; }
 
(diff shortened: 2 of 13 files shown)


Arisotura:
sorry for looking at this so late, but

looks interesting, minus inconsistent code style issues (opening brace should go on a new line...)

I'd merge this, however I'm not sure how well it will mesh with the OpenGL renderer rework I'm doing in the blackmagic3 branch, so I'll wait until that's finished and merged

legoabram:
@Arisotura Worry not. I'm happy to rework/fix it once you've finished the renderer rework.

Though I still have my question regarding if/how to allow customizing the filenames.

Arisotura:
hey hey

blackmagic3 is merged

so if you're able to update this, I'll gladly merge it

legoabram:
@Arisotura Initial rework complete. I'm loving the improvements you've made. Screenshots work for OpenGL and Software modes. I haven't tested the compute version just yet.

I've got two points of concern I'd like your input on:
1. I'm not sure if you'll be happy about how I'm propagating the scale factor from the 3D renderer to the EmuThread.
2. I was using the front buffer lock to synchronize before the rework, but that's gone now so I've skipping locking initially. It seems to work without it, but it gives me the heebie-jeebies to leave it as is. What would make the most sense to sync this with?

legoabram:
@Arisotura You have any thoughts? I'm happy to take another stab at this.

PR #2426, opened 2025-09-10
don't hardcode a version suffix in the libwayland-egl dlopen

Izder456:
This bug was discovered by thfr on the OpenBSD ports mailing list. Hardcoding this is not very portable. Shouldn't break other *NIX platforms like Darwin or Linux.


CasualPokePlayer:
This can (and will) break other *nix platforms. Generally, the non-suffix variant is only available as a dev package (as a symlink for building purposes). That suffix indicates the ABI version, and if it's not 1 in this case, that means the ABI is different, and therefore you cannot assume you can even correctly handle whatever you're trying to link.

Izder456:
Alrighty then. I'll go and close this then.

Issue #2432, opened 2025-09-11
[Feature Request] Allow Arbitrary Pixel Value for Screen Gap


urabeat:
<img width="3072" height="1594" alt="Image" src="https://github.com/user-attachments/assets/0a294849-e405-49ae-8f1e-ac96141be5bf" />

Hello,

Currently, melonDS only allows a few fixed values for the screen gap between the top and bottom screens. However, depending on the game, the ideal gap may vary slightly to match how elements are laid out across both screens.

I’d like to request the ability to set an arbitrary pixel value for the screen gap, so users can fine-tune the layout as needed.

Thank you.

PR #2436, opened 2025-09-13
Add support for wider range of screen gaps

dingdongg:
Addresses https://github.com/melonDS-emu/melonDS/issues/2432

this PR give users the option to specify a custom gap value, ranging from 0 to 128 pixels

image

### DIFF
--- a/src/frontend/qt_sdl/Window.cpp
+++ b/src/frontend/qt_sdl/Window.cpp
@@ -520,7 +520,26 @@ MainWindow::MainWindow(int id, EmuInstance* inst, QWidget* parent) :
                     actScreenGap[i]->setData(QVariant(data));
                     actScreenGap[i]->setCheckable(true);
                 }
+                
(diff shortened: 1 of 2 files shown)


src/frontend/qt_sdl/Window.cpp
```diff
@@ -753,12 +777,16 @@ MainWindow::MainWindow(int id, EmuInstance* inst, QWidget* parent) :
         actScreenRotation[windowCfg.GetInt("ScreenRotation")]->setChecked(true);
 
         int screenGap = windowCfg.GetInt("ScreenGap");
-        for (int i = 0; i < 6; i++)
+
+        if (!actScreenGap[6]->isChecked())
```
asiekierka: When is `actScreenGap[6]` pre-set to checked here? Feels to me like the below `for` loop should instead be adjusted so that if none of the default gap sizes are checked, the "Custom" checkbox is checked.

PR #2438, opened 2025-09-14
SD/MMC controller fixes

asiekierka:
Supersedes https://github.com/melonDS-emu/melonDS/pull/2415 - note that I could not test it at this time, however, so the PR might need some extra work.


edo9300:
If this pr makes the behaviors that my other PR fixed work properly as well (so make profi's tmio driver work for both nand and SD on melon), then I'll close mine.
I couldn't test it yet either.

asiekierka:
Tested (finally), appears to work fine.

PR #2444, opened 2025-09-18
return 0 when reading SCFG_ROM on the arm7 if it is disabled by SCFG_EXT

edo9300:
Currently melonds is returning the SCFG_ROM register on the arm7 on read even if it is disabled via SCFG_EXT


Arisotura:
could also add this check to the ARM9 SCFG registers

edo9300:
handled in https://github.com/melonDS-emu/melonDS/commit/60f2ced9fcc5f306b21078f8fc97a589f70013ae

PR #2449, opened 2025-09-27
progress for netplay

Isaac0-dev:
Hi, I wanted to see netplay working on a DS emulator, so I thought I'd try to build upon the efforts that have been made for melonds.
There is no functioning version available right now so this is only of interest to other devs, sorry!

I'm aiming for the same system that was proposed previously by other folks, which I believe is like this:
Each computer running melonDS emulates every DS that's connected to deal with the strict timing requirements of DS local multiplayer. Then, some sort of deterministic netplay would be implemented, streaming inputs across the network rather than the game's packets. 

So far this is just re-implementing the mirror prototype, to focus on perfecting the input syncing. 
I've hacked together a lot of this, and so I apologise for the mess I have made.
It uses an input history, where we send a short history of the last few frames of input over the network. Only after we've confirmed that every player on the network has seen that input do we stop sending it in each packet.

I'm a little sceptical on whether to use save states or lag to deal with inputs not arriving on time (there is an option to change the input delay of course), but I tried to use save states so far; if we don't get an input for a frame from any player, we save a state and run that frame anyway to avoid lag.
If we get the frame later, load the save state and re-apply all the inputs up to the present frame.

I've never made a netplay system of this sort before, and this is also my first time working on an emulator project.
So please, I'd love feedback!

### DIFF
--- a/src/frontend/qt_sdl/EmuInstance.cpp
+++ b/src/frontend/qt_sdl/EmuInstance.cpp
@@ -65,11 +65,14 @@ const string kWifiSettingsPath = "wfcsettings.bin";
 extern Net net;
 
 
-EmuInstance::EmuInstance(int inst) : deleting(false),
+EmuInstance::EmuInstance(int inst, bool createMainWindow) : deleting(false),
     instanceID(inst),
     globalCfg(Config::GetGlobalTable()),
     localCfg(Config::GetLocalTable(inst))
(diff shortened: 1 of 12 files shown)


Isaac0-dev:
i've ported this to the latest master branch so i'll reopen the pr on that branch

PR #2451, opened 2025-09-28
Quick fix UB in SoftRenderer Interpolator

CasualPokePlayer:
This is probably not correct behavior, but it's enough to avoid UB that ends up causing graphical bugs.


Arisotura:
I did commit a proper fix, so this PR won't be needed I guess...

PR #2453, opened 2025-09-29
netplay

Isaac0-dev:
This is my previous pr #2449 ported to the master branch.
Here's what I said in that pr:

---

Hi, I wanted to see netplay working on a DS emulator, so I thought I'd try to build upon the efforts that have been made for melonds.
There is no functioning version available right now so this is only of interest to other devs, sorry!

I'm aiming for the same system that was proposed previously by other folks, which I believe is like this:
Each computer running melonDS emulates every DS that's connected to deal with the strict timing requirements of DS local multiplayer. Then, some sort of deterministic netplay would be implemented, streaming inputs across the network rather than the game's packets. 

So far this is just re-implementing the mirror prototype, to focus on perfecting the input syncing. 
I've hacked together a lot of this, and so I apologise for the mess I have made.
It uses an input history, where we send a short history of the last few frames of input over the network. Only after we've confirmed that every player on the network has seen that input do we stop sending it in each packet.

I'm a little sceptical on whether to use save states or lag to deal with inputs not arriving on time (there is an option to change the input delay of course), but I tried to use save states so far; if we don't get an input for a frame from any player, we save a state and run that frame anyway to avoid lag.
If we get the frame later, load the save state and re-apply all the inputs up to the present frame.

I've never made a netplay system of this sort before, and this is also my first time working on an emulator project.
So please, I'd love feedback!


Isaac0-dev:
i recorded a quick test video. you can see it have a desync near the end so obviously that needs fixing.
In the video you can see the synced input in play, and also the save state system working (that's what happens when i pressed the Start button).
Obviously it's just a mirror so far, but I've also put in work towards headless instances ready for actual netplay.
I haven't prevented users from starting a game before starting netplay because I think it'd be very useful for players to be able to open netplay after reaching the point in their game that has multiplayer. I haven't thought of any reasons this shouldn't be done.

You can control the game from either window, and i was trying to get to a place to show that touch controls work, but it desynced before i made it.

https://youtu.be/EJDEeSHzpTg

RSDuck:
Please join our Discord or IRC, see the readme.

Isaac0-dev:
> Please join our Discord or IRC, see the readme.

hi, i'm on the discord server. is there something you wanted to discuss there?

lnee94:
I looked at this at this solution is about as carzy as mine I use kasm to host the emulators on the same computer and use streaming to allow players to play with it

Issue #2130, opened 2024-08-09
No$GBA debug registers not visible to the ARM7


pants64DS:
The [No$GBA debug registers](http://problemkaputt.de/gbatek.htm#dsdebugregistersemulatordevkits) starting at `04fffa00` can only be accessed by the ARM9 on melonDS. On No$GBA both CPUs can access them, like GBATEK says.

AntonioND:
I've also noticed this. As far as I can see, the only changes needed are to copy this:

https://github.com/melonDS-emu/melonDS/blob/f143e89c931d12a234851e443b7d85f8017db9a3/src/NDS.cpp#L3660-L3685

Right there:

https://github.com/melonDS-emu/melonDS/blob/f143e89c931d12a234851e443b7d85f8017db9a3/src/NDS.cpp#L4406

PR #2455, opened 2025-10-03
Add support for nocash debug registers on the ARM7

AntonioND:


AntonioND:
Fixes https://github.com/melonDS-emu/melonDS/issues/2130 to some degree, but this should be improved to work with other access sizes.

Arisotura:
sorry for looking at this so late, but

I'm not a big fan of having the wifi IO handlers fall through to the regular IO handlers... maybe we could make this more optimal with a bit of reworking. but maybe I'm just nitpicking.

AntonioND:
Hmmm without that change the accesses don't reach the right switch statement. I will take a look at it this weekend.

Arisotura:
I get the technical reason behind it, that's just my "code efficiency/niceness" part talking :P 

I think there's a way to make this a bit better... I'm thinking about it

AntonioND:
BTW, if someone comes up with a better idea, feel free to do it. I don't mind having my name in the commit, I'd just like this to be fixed.

AntonioND:
@Arisotura have you had time to think about the approach you prefer? I can close this PR if you want to implement it in a different way.

Issue #2493, opened 2025-11-21
The command-line arguments in version 1.1 no longer support East Asian filenames


swanki1998:
The command-line arguments in version 1.1 no longer support East Asian filenames, even though they were supported in version 1.0.

Example:
E:\melonDS\melonDS.exe "E:\ROM\NDS\宝可梦不可思议迷宫 空之探险队.nds"

Is it possible to revert it back?🥺

Issue #1959, opened 2024-01-21
ROMs not loadable from non-ASCII file paths


lesserkuma:
**System used:**
Windows 11 23H2

**Build used:**
https://github.com/melonDS-emu/melonDS/actions/runs/7586469528

**Problem:**
Can't load ROMs from a path including characters that are not native to the operating system language.

**Steps to reproduce:**
1. Rename a confirmed working ROM file to something like `あああ.nds`, or copy it into a directory path with such non-ASCII characters.
2. Try to load it with melonDS via any means (File → Open ROM..., Drag & Drop, or Command Line).
3. Error message "Failed to load the ROM." appears.

RSDuck:
was already fixed a few days after 0.9.5 release #1556

lesserkuma:
Ah thanks! I couldn't find it via search.

I tested it on this build btw., should the fix not be implemented into it?
https://github.com/melonDS-emu/melonDS/actions/runs/7586469528

RSDuck:
yes

lesserkuma:
Hmm, the issue is definitely still there with this build from January 19th, 2024. Or it's a different issue, I'm not sure.

RSDuck:
looks like it is a regression. I just tested it and it used to work.

Issue #1690, opened 2023-05-26
Certain special characters in filenames within archives cause ROM load failure


CasualPokePlayer:
![image](https://github.com/melonDS-emu/melonDS/assets/50538166/ec6da838-edd3-4260-900f-07733a05a03e)

Results in failing to load the ROM as it cannot understand the extension for whatever reason. Both zip/7z seem to have issues with this (haven't tested .rar as 7zip can't create those). Not all special characters seem to have an issue, é seems to be fine, but (for example) 金 is not (maybe all CJK is like this?)

This is tested on Windows, this maybe doesn't affect other OSes?

patataofcourse:
is this from the CLI, the GUI, or both?

CasualPokePlayer:
Any (also, these are from the names inside the archive, not the name of the archive itself)

RupeeClock:
The issue seems to be that melonDS 0.9.5 cannot parse certain characters in filenames on Windows. I'm using Windows 10 Pro version 22H2.
`ポケモン不思議のダンジョン 青の救助隊 [DAXJ01].nds` will fail, same as a `Pokémon Mystery Dungeon Explorers of Sky [DA4P01].nds`, but not a `Pokemon Mystery Dungeon Explorers of Sky [DA4P01].nds`. with the accented é replaced.
This occurs whether you launch the .nds file via Explorer using file association with the melonDS executable, or via CMD with melonDS as a CLI application.

This does not appear to affect the libretro melonDS 0.9.3 core when used with Retroarch 1.19.1, as long as you load the content from within the Retroarch GUI.
If using Ctrl+O to load content via Windows Explorer's Open File dialogue, this can then run into the same issue of being unable to load content from files with filenames containing special characters. This isn't exclusive to the melonDS core, I was able to reproduce this when using the mGBA core.

Curiously, using drag-and-drop from an Explorer window works for both melonDS standalone and as a libretro core in Retroarch. This will allow the ROM to boot from a file using special characters in the filename.
Other Windows applications I've used do not have this filename character issue, such as the standalone mGBA 0.10.0 executable. I wonder if it's a file system API that both Retroarch and melonDS are using that has this issue when using Open Dialogues or Explorer File Association?

RSDuck:
[...]

Issue #1689, opened 2023-05-26
Failed to load rom from command line when file name has special character


Homaxz:
I'm trying to run Pokemon SoulSilver from the command line as such:

`melonds.exe "C:\Users\ADMIN\Desktop\Pokémon - SoulSilver Version.nds"`

melonDS starts and gives the error "Failed to load the ROM."

Starting the rom from the GUI>file option works fine. 
Renaming the rom and to "Pokemon - SoulSilver Version.nds" makes it load through the cmd as well.

RSDuck:
which version are you using, I think we fixed this one already.

Homaxz:
Sorry I forgot to mention. 0.9.5

RSDuck:
then it is a duplicate of #1556

PR #2494, opened 2025-11-21
Fix East Asian char or non ASCII char booting in windows commandline

Wokann:
Thie filename of rom which contains East Asian char or non ASCII char has error when booting from command lin in windows. #1556 #1689 #1690 #1959 #2493
This is an old problem which has had been solved. 
But in v1.1, this problem back again. 
This is beacuse a function in `src\frontend\qt_sdl\main.cpp` was removed.
`int CALLBACK WinMain(HINSTANCE hinst, HINSTANCE hprev, LPSTR cmdline, int cmdshow)`
So I restore it and add comments to remind and prevent accidental deletion.

### DIFF
--- a/src/frontend/qt_sdl/main.cpp
+++ b/src/frontend/qt_sdl/main.cpp
@@ -407,6 +407,9 @@ int main(int argc, char** argv)
 
 #include <windows.h>
 
+// This function is needed for booting filename which contains East Asian char 
+// or non ASCII char from CommandLine in windows.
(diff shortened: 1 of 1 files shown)


RSDuck:
Our WinMain was removed for compability with building with msvc clang, where for some reason SDL's Winmain is required.

https://github.com/melonDS-emu/melonDS/commit/5c45016847b994b6a5f9be503430ec4994eb8edc makes it work as well.

PR #2505, opened 2025-11-30
Write saved state to file after unpausing emulator

anlsh:
I've set up my filesystem such that MelonDS saves to a network filesystem, which is handy for backup purposes. Unfortunately, this causes a noticeable pause while the file is (slowly) written out.

So do the file write in the qt thread instead: this way the emulator can resume ASAP.

I do see that I've violated CONTRIBUTING.md by using `std::ofstream` but... is that a big deal?

### DIFF
--- a/src/frontend/qt_sdl/EmuInstance.cpp
+++ b/src/frontend/qt_sdl/EmuInstance.cpp
@@ -772,19 +772,12 @@ bool EmuInstance::loadState(const std::string& filename)
     return true;
 }
 
-bool EmuInstance::saveState(const std::string& filename)
+bool EmuInstance::saveState(std::string& buffer)
 {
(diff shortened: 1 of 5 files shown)


Arisotura:
I think the idea of saving stuff on a separate thread is worth looking into (for situations like yours), however, not sure about your approach...

anlsh:
> however, not sure about your approach...

Why?

asiekierka:
Wouldn't this lead to a situation where the emulator is writing to the very data being placed in the save state, potentially leading to a corrupted state (say, if the emulation thread changed console RAM or CPU registers while the save method is running in the Qt thread)?

anlsh:
The emulator is still paused paused while the state is serialized to a buffer- it's just that the serialization of the buffer to the file no longer blocks the emulator from resuming. I don't believe that the emulator reuses the buffer in question for anything else, so there's no data race to speak of

Issue #1548, opened 2022-10-29
DSi mode Compatibility list


KostaSaizo7:
~Tested melonDS 0.9.5 DSi mode~
Re-tested with melonDS 1.1, direct mode now works

**Broken: Hangs when booting (black or white screen)**

- Again - Interactive Crime Novel
- Dragon Quest Monsters - Joker 2
- Dragon Quest IX - Sentinels of the Starry Skies
- Mario & Luigi - Bowser's Inside Story
- Mega Man Zero Collection
- Nine Hours, Nine Persons, Nine Doors
- Pokemon Ranger: Guardian Signs
- Phantasy Star 0 [phantasy star 0 dsi log.txt](https://github.com/user-attachments/files/21135300/phantasy.star.0.dsi.log.txt)
- Professor Layton and the Unwound Future
- Rooms - The Main Building

**Intro/Menu: Hangs somewhere between booting and starting**

- Blue Dragon Awakened Shadow - Hangs after logos
- Dragon Ball Origins 2 - Hangs after logos
- Final Fantasy The 4 Heroes of Light [ff4 log](https://github.com/melonDS-emu/melonDS/files/9741962/Final.Fantasy.-.The.4.Heroes.of.Light.USA.txt) - Hangs after creating the characters / loading a save
- Pokemon HeartGold/SoulSilver - Hangs after naming the player / loading a save

KostaSaizo7:
A log file where I launched  Xtransceiver in DSi mode
[log.txt](https://github.com/melonDS-emu/melonDS/files/9895560/log.txt)

poudink:
do the games run better (or run at all) with builds from before DSP emulation was added?

KostaSaizo7:
With 0.9.4 stable, James Cameron's Avatar: The Game and WarioWare - Snapped! don't run at all. The other games crash when they enable the dsi camera.

KostaSaizo7:
opened new issue for the dsp https://github.com/melonDS-emu/melonDS/issues/2367

huyhoang160593:
> Splitting the dsp issue
> 
> **Low FPS possibly due to DSP, right from booting the game**
> 
> * Hidden Photo (DSi exclusive game card)
> * James Cameron's Avatar: The Game
> * My Cooking Coach - Prepare Healthy Recipes
> * My Healthy Cooking Coach - Easy Way to Cook Healthy
> * WarioWare - Snapped! (DSiWare)
> 
[...]

PR #2523, opened 2025-12-27
Allow booting in DSi direct mode without BIOS/firmware/NAND dumps.

asiekierka:
This is an initial patch which enables booting programs in DSi direct mode without requiring BIOS/firmware/NAND dumps from increasingly expensive consoles:

- The open DSi BIOS is derived from a modified build of FreeBIOS; minor patches have been done (such as moving SWI 0x12 to 0x02), but no new SWI functions have been implemented.
- The DSi firmware stub is generated in the same way as the NDS firmware stub; this appears to be good enough for BlocksDS's dswifi to pick up the new AP definitions.
- The DSi NAND is currently not stubbed at all; rather, the code has been edited to allow bringup without a NAND image. This is sufficient for DSi mode homebrew; programs running from the cartridge slot are likewise not expected to require NAND access.

More work has to be done before it's useful for non-homebrew scenarios.

### DIFF
--- a/src/SPI_Firmware.h
+++ b/src/SPI_Firmware.h
@@ -164,7 +164,7 @@ public:
     static_assert(sizeof(ExtendedWifiAccessPoint) == 512, "WifiAccessPoint should be 512 bytes");
 
 
-    enum class FirmwareConsoleType : u8
+    enum FirmwareConsoleType : u8
     {
         DS = 0xFF,
         DSLite = 0x20,
@@ -266,7 +266,7 @@ public:
             u8 BuildMonth;
             u8 BuildYear;
 
-            FirmwareConsoleType ConsoleType;
+            u8 ConsoleType;
 
             u8 Unused0[2];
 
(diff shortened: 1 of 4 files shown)


KostaSaizo7:
I have tested this pr and most DS games work in DSi mode without BIOS/firmware/NAND dumps.

Games that do not boot:
- DSiware
- DSi-exclusive game cards
- [DS games that are DSi enhanced](https://www.rfgeneration.com/blogs/pdrydia/DSi-enhanced-titles-an-incomplete-list-1969.php)
- [DS games that do not boot in DSi mode even with BIOS/firmware/NAND dumps](https://github.com/melonDS-emu/melonDS/issues/1548)

asiekierka:
Yeah, so there's no regression. Of course, the BIOS needs work before it can run non-homebrew DSi programs.

Arisotura:
looks good to me, just gotta fix that one conflict that popped up...

asiekierka:
Done.

Arisotura:
okie dokie

PR #2531, opened 2026-01-02
Use intrinsics to optimize bit counting operations

asiekierka:
x64 features a "count bits" intrinsic, while on other platforms it can typically be performed faster than through a naive for loop.

This should provide a very small performance improvement; it also simplifies the code a little.

### DIFF
--- a/src/ARMInterpreter_LoadStore.cpp
+++ b/src/ARMInterpreter_LoadStore.cpp
@@ -400,11 +400,7 @@ void A_LDM(ARM* cpu)
 
     if (!(cpu->CurInstr & (1<<23)))
     {
-        for (int i = 0; i < 16; i++)
-        {
-            if (cpu->CurInstr & (1<<i))
-                base -= 4;
-        }
+        base -= 4 * __builtin_popcount(cpu->CurInstr & 0xFFFF);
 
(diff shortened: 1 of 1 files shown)


RSDuck:
Within some bounds the naive loop can actually be faster. I did some benchmarking for the scheduler already a few years ago and it was actually slower.

It turns out that at least in that case the compiler unrolls that loop making and then the branch predictor probably does its magic. But even without it the naive loop has fewer dependencies from one iteration to another which is also very beneficial.

asiekierka:
Fair; though I'd guess in that case the popcounts should still be faster, as they can be optimized down to either an intrinsic (x64) or a more efficient bit twiddling operation than a naive or unrolled loop?

RSDuck:
Yeah, that's probably helpful. Population count is a relatively recent instruction on x64, so it's going to be bit twiddling unless we kick out pre Nehalem cpus.

RSDuck:
What about https://github.com/melonDS-emu/melonDS/blob/master/src/Wifi.cpp#L571?

asiekierka:
I don't think it is used anywhere near often enough to justify it there, plus the WiFi code is more of a moving target.

PR #2541, opened 2026-01-14
add hotkey to cycle layout

### LINKED ISSUE #1868: Feature request : Hotkey to change the screen layout.
Could be nice to have.
Since i'm in a hurry, i did a diry mod where i just recyled the onScreenEmphasis keybind to do the job, but that's something worth to have his own keybind
void MainWindow::onScreenEmphasisToggled()
{
	int layout = Config::ScreenLayout;
	layout++;
	if(layout>4) layout = 0;
	Config::ScreenLayout = layout;
	emit screenLayoutChange();
}

Also, i can figure what is the difference beetween Natural and Vertical ?

guidocella:
This actually switches back to the natural layout when possible depending on the current rotation.

Closes #1868.
Supersedes #1020.


Arisotura:
melonDS team, your thoughts on this?

RSDuck:
Looks ok to me.

EDIT: building it right now
EDIT2: actually there is a bit of a hole in the state machine. If you start from hybrid layout you never go back to it?

guidocella:
I did not see the edit. It now goes Natural →Horizontal → Hybrid →Natural or Natural →Vertical → Hybrid → Natural.

PR #2542, opened 2026-01-15
Fix C++17 compatibility by removing designated initializers

JoJoKorok:
### Summary
This PR replaces C++20-style designated initializers in `.cpp` files with
C++17-compatible initialization patterns.

### Motivation
When I initially was setting up my fork on my local machine, I noticed that melonDS targets C++17. While some compilers accept designated initializers
as extensions, they are not portable or guaranteed across all supported
tool-chains. This change improves portability without changing behavior.

### Scope
- No functional changes
- No formatting-only changes
- No build system changes

### Notes
This PR keeps initialization semantics identical by using value-initialization
followed by explicit member assignment.

### DIFF
--- a/src/ARCodeFile.cpp
+++ b/src/ARCodeFile.cpp
@@ -129,7 +129,7 @@ bool ARCodeFile::Load()
             if (ret < retchk)
             {
                 Log(LogLevel::Error, "AR: malformed CAT line: %s\n", start);
-                CloseFile(f);
+                Platform::CloseFile(f);
                 return false;
             }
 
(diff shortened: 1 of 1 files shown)


Arisotura:
debating whether we should do this, or bump the requirements to C++20

(sorry tho, I did initializers this way because I like it, wasn't aware of the requirements for it)

JoJoKorok:
> debating whether we should do this, or bump the requirements to C++20
> 
> (sorry tho, I did initializers this way because I like it, wasn't aware of the requirements for it)

If it were bumped up to C++ 20 I feel like it would make it easier for if not only me, just others depending on what their IDE is working on for other projects. It was my first time going through the main and I just noticed that was all so I thought I would do a quick refactor.

asiekierka:
The vast majority of C++20 features were supported [by GCC 11](https://en.cppreference.com/w/cpp/compiler_support/20.html), which is now almost 5 years old. Given the longest LTS cycles in consumer Linux distributions converge around 2 years, I think it's a valid point in time to just bump it up...

PR #2504, opened 2025-11-29
Audio mute options

nico-abram:
This implements 2 things:

1. A mute toggle hotkey. This is something I personally want, since I often listen to external music while using melonDs and only want to unmute the game for specific sections.
2. A checkbox option in the Interface Menu to mute the game while fastforwarding (Either via fastforward toggle or the held down fastforward hotkey)

I put the checkbox in interface for now but I can move it to Audio if that seems more fitting.

imagen
imagen

This is my first PR to this project, apologies in advance if I'm missing something.

Implements https://github.com/melonDS-emu/melonDS/issues/1543 , https://github.com/melonDS-emu/melonDS/issues/2056, https://github.com/melonDS-emu/melonDS/issues/1605

PR #2546, opened 2026-01-16
Audio mute hotkey

JoJoKorok:
### Summary
Adds a configurable hotkey to toggle game audio mute in the Qt frontend.

### Motivation
When I was trying to stream both Tekken and Pokemon Black at the same time, I found myself constantly going to mute melonDS when I found an opponent. Currently, muting audio requires interacting with the volume slider or menu.
A dedicated hotkey is useful for gameplay, testing, and streaming scenarios.

### Implementation
- Adds a new hotkey action: "Toggle audio mute"
- Stores the last non-zero volume to restore on unmute
- Reuses existing audio volume handling
- Qt frontend only; no core emulation changes

### Testing
- Built and tested on Windows (MSYS2, Qt5 static)
- Verified:
  - audio toggles immediately on hotkey press
  - previous volume is restored correctly
  - volume slider behavior unchanged

### DIFF
--- a/src/frontend/qt_sdl/Config.cpp
+++ b/src/frontend/qt_sdl/Config.cpp
@@ -55,7 +55,7 @@ DefaultList<int> DefaultInts =
     {"Screen.VSyncInterval", 1},
     {"3D.Renderer", renderer3D_Software},
     {"3D.GL.ScaleFactor", 1},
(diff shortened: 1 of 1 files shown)


Arisotura:
I merged #2504 which does the same thing but with an extra bit of functionality, so now this is redundant...

I appreciate the PR, tho :)

Issue #2386, opened 2025-07-28
Implement icon through QWindow::setIcon


danielfrrrr:
Currently, the AppImage release does not have an icon in the taskbar and Alt + Tab. This can be solved by implementing Qt `QWindow::setIcon`, which will allow multi-platform systems to have an icon, such as Linux x11/wayland and Windows. Apparently melonDS does not support Linux wayland, but if it does in the future, `QWindow::setIcon` will allow icons on Linux wayland without relying on a desktop entry file (as is customary in Linux wayland).

nadiaholmquist:
The window icon not being set is actually intentional, that way it can be picked up from the desktop file so if the user has set a custom icon for melonDS (or icon theme, or whatever) then it will be picked up by the program itself.

I don't know if there's anything we can do to make it work properly with the AppImage, and also, keep in mind that windows setting their own icons is not universally supported in Wayland as far as I know.

danielfikko:
> I don't know if there's anything we can do to make it work properly with the AppImage, and also, keep in mind that windows setting their own icons is not universally supported in Wayland as far as I know.

Fortunately, this has changed in recent months with the [xdg-toplevel-icon](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/blob/main/staging/xdg-toplevel-icon/xdg-toplevel-icon-v1.xml) protocol, allowing wayland clients to change their window icons. All you need is Qt 6.10 and to implement the icon through `QWindow::setIcon`.
https://codereview.qt-project.org/c/qt/qtwayland/+/558537

> The window icon not being set is actually intentional, that way it can be picked up from the desktop file so if the user has set a custom icon for melonDS (or icon theme, or whatever) then it will be picked up by the program itself.

Isn't that the compositor's job? For example, labwc gives priority to desktop entry files, and if that fails, labwc then uses the icon provided by the `xdg-toplevel-icon` protocol.
https://github.com/labwc/labwc/pull/2839

EDIT: I'm not a programmer, so I can't say for sure, but reading the description of https://codereview.qt-project.org/c/qt/qtwayland/+/558537, it seems that even Qt prioritizes icons defined by the theme first, then tries the `xdg-toplevel-icon` protocol.

danielfikko:
[...]

PR #2547, opened 2026-01-18
ci: add wayland support to AppImage releases

danielfrrrr:
Currently, when you run melonDS with `QT_QPA_PLATFORM=wayland ./melonDS-x86_64.AppImage`, you get this error:

```
qt.qpa.plugin: Could not find the Qt platform plugin "wayland" in ""
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

Available platform plugins are: xcb.
```

This is because `linuxdeploy-plugin-qt` needs to explicitly configure wayland support at build time. That is what this PR does.

Reason for adding the new dependency:

`$EXTRA_PLATFORM_PLUGINS` needs the files `/usr/lib/{aarch64,x86_64}-linux-gnu/qt6/plugins/platforms/{libqwayland-egl.so,libqwayland-generic.so}`, found in the `qt6-wayland` package.

References for my implementation:
https://github.com/linuxdeploy/linuxdeploy-plugin-qt#environment-variables
[PCSX2](https://github.com/PCSX2/pcsx2/blob/edb2b37a925635150d3217cdd4bccff3de05b3f7/.github/workflows/scripts/linux/appimage-qt.sh#L111-#L112)

With this PR, this issue will eventually be closed: https://github.com/melonDS-emu/melonDS/issues/2386. In Qt 6.9, support was added for wayland clients to set their window icon using `setWindowIcon`, which in turn uses the wayland `xdg-toplevel-icon` protocol. However, AppImage is currently being built and packaged against a version lower than 6.9, due to the fact that the Ubuntu image of the GitHub Actions runner is using version 24.04 LTS, where the `qt6-wayland` package is still at version **_6.4_**! Therefore, this PR is the first step, then it is just a matter of waiting for GitHub Actions to update its system to a version that has Qt 6.9, and the icon will likely be available in wayland.

### DIFF
--- a/.github/workflows/build-ubuntu.yml
+++ b/.github/workflows/build-ubuntu.yml
@@ -36,7 +36,7 @@ jobs:
       run: |
         sudo apt update
         sudo apt install --allow-downgrades cmake ninja-build extra-cmake-modules libpcap0.8-dev libsdl2-dev libenet-dev \
-          qt6-{base,base-private,multimedia}-dev libqt6svg6-dev libarchive-dev libzstd-dev libfuse2 libfaad-dev
+          qt6-{base,base-private,multimedia}-dev qt6-wayland libqt6svg6-dev libarchive-dev libzstd-dev libfuse2 libfaad-dev
     - name: Configure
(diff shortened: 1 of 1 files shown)


Arisotura:
gonna see if this builds fine - if so, I'll merge it

PR #2564, opened 2026-02-01
Fix BIOSPROT initialization in a full-BIOS-boot scenario

PoroCYon:


Arisotura:
ARM7BIOSProt should be initialized to 0 - setting it to 0xFFFF will make it impossible to write to it again (the write handler checks for value 0)

PR #2567, opened 2026-02-03
feat: add weighted screen sizing

matthr0w:
This PR introduces a new weighted screen sizing option that allows users to independently select top and bottom screen weights. It integrates cleanly with the existing sizing modes and provides more layout flexibility without adding unnecessary complexity.

### DIFF
--- a/src/frontend/ScreenLayout.cpp
+++ b/src/frontend/ScreenLayout.cpp
@@ -127,6 +127,7 @@ void ScreenLayout::Setup(int screenWidth, int screenHeight,
     int screenGap,
     bool integerScale,
     bool swapScreens,
+    int topWeight, int botWeight,
     float topAspect, float botAspect)
 {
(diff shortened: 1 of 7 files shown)


Mambouna:
Does this provide the option to change by how much one screen is emphasized, when that layout option is picked? I just started using melonDS and am very happy, but would prefer to have the top screen emphasized just a little, instead of so much. Your PR sounds like it fixes that, right?

Matthr0w:
Yes, it does. You have a new screen sizing option called _Weighted_, where you can assign a weight to each screen. For example, if you want a slightly larger top screen, you can set the top screen to 3x and the bottom screen to 2x.

Mambouna:
Nice, I hope it'll be merged.

matthr0w:
Closed, as #2634 handles this better.

PR #2569, opened 2026-02-06
You can keep multiple mappings, for different joystick models, and they won't overwrite each other

vitor251093:
As the title implies, with those changes, melonDS now stores the joystick mappings by model, so you can use different controller models without their mappings overwriting each other.

As of this PR, I'm grouping the joystick mappings by a combination of the Vendor ID and the Device ID, which I refer in the code as the "Unique ID". The combination of Vendor ID and Device ID is fine grained enough to differ between a DualShock 4 v1 (0x054c05c4) and a DualShock 4 v2 (0x054c09cc).

I've removed the legacy imports of joystick mappings since the OG `Instance*.Joystick` isn't used after this PR.

This is still being tested, which is why it's, for now, a draft.


vitor251093:
Tested, and now ready for validation

PR #2570, opened 2026-02-07
Start again with screens turned on

edo9300:
Reverts https://github.com/melonDS-emu/melonDS/commit/2e9434f7ba91c3c864fc005f65f407049fdda039, after hardware testing, the console does boot up with the powcnt1 register being set to 1 (tested when booting straight from gcdfirm)
immagine

### DIFF
--- a/src/DSi.cpp
+++ b/src/DSi.cpp
@@ -184,6 +184,9 @@ void DSi::Reset()
     GPU.DispStat[0] |= (1<<6);
     GPU.DispStat[1] |= (1<<6);
 
+    // On DSi the LCDs are already powered on at boot time
+    PowerControl9 = 0x0001;
+
     UpdateVRAMTimings();
 }
 
--- a/src/NDS.cpp
+++ b/src/NDS.cpp
@@ -492,7 +492,7 @@ void NDS::Reset()
 
     PostFlag9 = 0x00;
     PostFlag7 = 0x00;
-    PowerControl9 = 0x0001;
+    PowerControl9 = 0x0000;
     PowerControl7 = 0x0000;
 
     WifiWaitCnt = 0xFFFF; // temp


Arisotura:
When in the boot chain is your code running?

My testing shows that the BIOS bootloader sets POWCNT to 0, then the firmware initializes it.

edo9300:
> When in the boot chain is your code running?

Right after stage1 (since it is a stage2 firm).


> My testing shows that the BIOS bootloader sets POWCNT to 0, then the firmware initializes it.

By doing a full bios boot, you can see yourself that the first write to powcnt is done by the stage2, and in my gcdfirm payload I'm not setting that bit explicitly at all (and it works on hw)

Arisotura:
that is indeed true for the DSi, the BIOS doesn't touch POWCNT1 at all...

however, there's no way to verify it for the DS, since the BIOS sets POWCNT1 to 0 there.

I guess we can explicitly change the reset value in DSi::Reset().

edo9300:
Right, that's shared for the ds as well :sweat_smile:, I guess it could also be confirmed there by flashing a custom firmware

Arisotura:
Unless you're able to somehow replace the BIOS (which is built into the SoC), you can't make that claim about the DS. It _might_ be true, but there's no way to verify it. Keep in mind that LCD stuff works somewhat differently in the DSi.

edo9300:
> Unless you're able to somehow replace the BIOS (which is built into the SoC), you can't make that claim about the DS. It _might_ be true, but there's no way to verify it. Keep in mind that LCD stuff works somewhat differently in the DSi.

Ah, so it's not done by the firmware, right, well, applying your suggestion and setting it to 1 only on dsi

Arisotura:
the rationale behind that change was to mirror what I could observe - the screens start powered off. if POWCNT1 is inited to 1 in DS mode, you get a spurious white flash on boot...

however, that seems fine for DSi mode, so let's do this.

on the DSi, you also have that "LCD init" flag in DISPSTAT, so the LCDs take some time to actually power on. the firmware checks the flag (on the ARM7), so it'd be interesting to see how long it takes to get set. (would also be interesting to see if that flag is affected by POWCNT1 at all)

PR #2583, opened 2026-02-21
Fix @loader_path find and replace for static builds

vnescape:
I followed the instructions in BUILD.md on macOS 15.7.4 to build melonDS statically, but it failed. I resolved the issue by respecting the @loader_path that some libraries use for their includes.

### DIFF
--- a/tools/mac-libs.rb
+++ b/tools/mac-libs.rb
@@ -120,9 +120,12 @@ end
 def fixup_libs(prog, orig_path)
   throw "fixup_libs: #{prog} doesn't exist" unless File.exist? prog
 
-  libs = get_load_libs(prog)
-    .map { |it| expand_load_path(orig_path, it) }
-    .select { |it| not system_lib? it[0] }
+  raw_libs = get_load_libs(prog)
+  
+  libs = raw_libs.map do |raw|
+    expanded = expand_load_path(orig_path, raw)
+    expanded ? [raw, expanded[0], expanded[1]] : nil
+  end.compact.select { |it| not system_lib? it[1] }
 
   FileUtils.chmod("u+w", prog)
   strip prog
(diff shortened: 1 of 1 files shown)


nadiaholmquist:
You didn't actually build a static build if it depends on any non-system dylibs, if you used homebrew for the dependencies then I don't think that will work.

vnescape:
> You didn't actually build a static build if it depends on any non-system dylibs, if you used homebrew for the dependencies then I don't think that will work.

I compiled it in a VM to isolate my dev environment and then copied it over to another machine to see if it worked. Before that I would need to have the QT libs installed on the machine. Now I can run the application without the QT libs being installed on the target machine.

nadiaholmquist:
There shouldn't be any dylibs in the first place in the expected setup for static builds.

If you look inside the app bundle you will notice that the Qt libraries are in there, and if that's fine (we build statically linked builds to save file size) you can just build a dynamically linked build and the script should work fine as-is.

vnescape:
I wanted to build a macOS build of melonDS that does not require any external lib to be installed on the target machine.
To do this I executed the following commands to build it on a fresh copy of macOS 15 in a VM.

1. `brew install git pkg-config cmake sdl2 qt@6 libarchive enet zstd faad2`
2. `git clone https://github.com/melonDS-emu/melonDS`
3. `cd melonDS`
4. `cmake -B build -DMACOS_BUNDLE_LIBS=ON -DCMAKE_PREFIX_PATH="$(brew --prefix qt@6);$(brew --prefix libarchive)"`
5. `cmake --build build -j$(sysctl -n hw.logicalcpu)`

This will lead to the following error message:

```
[100%] Linking CXX executable ../../../melonDS.app/Contents/MacOS/melonDS
Bundling macOS libraries...
/Users/user/Downloads/melonDS/tools/mac-libs.rb:125:in `block in fixup_libs': undefined method `[]' for nil:NilClass (NoMethodError)
	from /Users/user/Downloads/melonDS/tools/mac-libs.rb:125:in `select'
	from /Users/user/Downloads/melonDS/tools/mac-libs.rb:125:in `fixup_libs'
	from /Users/user/Downloads/melonDS/tools/mac-libs.rb:267:in `block in <main>'
	from /Users/user/Downloads/melonDS/tools/mac-libs.rb:261:in `each'
	from /Users/user/Downloads/melonDS/tools/mac-libs.rb:261:in `<main>'
make[2]: *** [melonDS.app/Contents/MacOS/melonDS] Error 1
make[2]: *** Deleting file `melonDS.app/Contents/MacOS/melonDS'
make[1]: *** [src/frontend/qt_sdl/CMakeFiles/melonDS.dir/all] Error 2
make: *** [all] Error 2
```

This PR tries to fix this problem and be able to build melonDS on macOS using the `-DMACOS_BUNDLE_LIBS=ON ` flag successfully.

PR #2595, opened 2026-03-08
Cartridge emulation improvements

### LINKED ISSUE #380: Game crashes after second cutscene in The World Ends With You

### LINKED ISSUE #2586: Rabbids Go Home works in DSi mode but not in DS mode
First reported here https://github.com/melonDS-emu/melonDS/issues/1831#issuecomment-1723997635

With DS mode after the ubisoft logo it just shows blue and black screens

Image

Interestingly it also happens with desmume https://github.com/TASEmulators/desmume/issues/483

Booting from a DSi firmware, the games works perfectly fine.

Image

Arisotura:
* refine cart interface implementation to match hardware, including timing fixes
* add second cart slot for DSi
* fully implement SCFG_MC
* move cart classes to subdirectory

also, I introduced a new naming convention (kind of following up, I really started it in blackmagic3)

struct = sThing
class = cThing
etc

I kind of want to extend it to all of the melonDS core, to avoid class names conflicting with variable names. thoughts?

fixes #2586 
fixes #380 (might need extensive testing)
fixes Surviving High School (DSiWare)

### DIFF
--- a/src/DSi_NDMA.cpp
+++ b/src/DSi_NDMA.cpp
@@ -134,8 +134,8 @@ void DSi_NDMA::WriteCnt(u32 val)
         // * timers (00-03)
         // * NDS-wifi?? (ARM7 07, likely not working)
 
-        if (StartMode <= 0x03 || StartMode == 0x05 || (StartMode >= 0x0C && StartMode <= 0x0F) ||
-            (StartMode >= 0x20 && StartMode <= 0x23) || StartMode == 0x25 || StartMode == 0x27 || (StartMode >= 0x2D && StartMode <= 0x2F))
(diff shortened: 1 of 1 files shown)


KostaSaizo7:
fixes #380 probably the most notorious melonDS bug

Arisotura:
will clean it up tomorrow... the code is still kinda messy/inconsistent, so gotta fix that up. then we'll merge.

PR #2612, opened 2026-03-27
CI : update actions/checkout to latest version

Mrlinkwii:
updates the actions/checkout , their is various other  out of date 

would recommend using like https://docs.github.com/en/code-security/tutorials/secure-your-dependencies/dependabot-quickstart-guide for automatic updating of versions


RayyanAnsari:
thanks!

PR #2629, opened 2026-04-17
Fix portable mode not working when running as AppImage

CyrixJD115:
Hey! So here's the deal.

If you run melonDS as an AppImage, the `portable` folder you placed next to it just gets ignored. That's because Qt thinks the app is running from some random `/tmp/.mount_...` path (the internal FUSE mount point) instead of where you actually put your `.AppImage` file. So it never finds the `portable` marker.
### What this does
- Checks the `$APPIMAGE` environment variable, which the AppImage runtime sets to the real path of the file
- Uses that to look for the `portable` folder in the right place instead of the temp mount path
### What if I'm not using an AppImage?
This changes nothing for you. Completely harmless.
### How to test
1. Put a `portable` folder next to your AppImage
2. Run it
3. Check that melonDS uses it as the config directory

Cheers!

### DIFF
--- a/src/frontend/qt_sdl/main.cpp
+++ b/src/frontend/qt_sdl/main.cpp
@@ -193,6 +193,15 @@ void pathInit()
     }
 #endif
 
+#if defined(__linux__)
+    QByteArray appimageEnv = qgetenv("APPIMAGE");
+    if (!appimageEnv.isEmpty())
+    {
+        QFileInfo appimageInfo(QString::fromUtf8(appimageEnv));
+        portablepath = appimageInfo.absolutePath() + QDir::separator() + "portable";
+    }
+#endif
+
     QDir portabledir(portablepath);
     if (portabledir.exists())
     {


poudink:
The ROM library stuff seem unrelated to the changes advertised in the PR. Was it added by mistake?

CyrixJD115:
@poudink Yea they were added by mistake, let me fix that

nadiaholmquist:
You can just use AppImage's standard mechanism for portable apps, if you create a directory with the same name as the AppImage ending in .config (like `melonDS-x86_64.AppImage.config`), it will store the configuration files there instead of your standard user config directory.

CyrixJD115:
> You can just use AppImage's standard mechanism for portable apps, if you create a directory with the same name as the AppImage ending in .config (like `melonDS-x86_64.AppImage.config`), it will store the configuration files there instead of your standard user config directory.

@nadiaholmquist , I mean, most people probably aren’t gonna know about this, especially since it’s not really mentioned in the docs anywhere. It’s honestly a pretty simple thing to implement too. Even I don’t really like having to rely on AppImage-specific workarounds when it could just check for a `portable` folder instead and make the whole portable setup way more straightforward. It’s literally like 9 lines of code, and then it’s just set it and forget it.

PR #2634, opened 2026-04-21
Redesigned and added more screen layouts

KojoZero:
### Changes
  - The screen layouts have been changed to closer resemble the style of Azahar since it's much easier to use than the current system, has all the same layouts + more, and is immediately recognizable to anyone who has used Azahar.
  - Large Screen Scaling which lets you choose the ratio between the larger and smaller screens when in Large Screen layout. "Auto" acts the same as "Emphasize Top/Bottom".
  - Small Screen Position lets you move the smaller screen to the various positions when in Large Screen layout.
  - Swap Screen now switches the position of the two screens as well as the emphasis as this is the behavior most users expect.
  - Removed "Screen Sizing" from the GUI. 
Screenshot 2026-04-16 094551
Screenshot 2026-04-16 094712
Screenshot 2026-04-16 094611


serenity4:
Looks like a great addition, I tested it and it is much more ergonomic than the original. Would be great if a maintainer could review it and get it merged.

PR #2642, opened 2026-04-27
Make the OpenGL (Classic) renderer work on OpenGL 3.1

FireNX70:
All shaders were already using version 140, which corresponds to OGL 3.1. None of them seem to be using any extensions introduced into the core spec in 3.2, as far as I can tell.

My first go at checking whether the CPU side was calling any functions introduced into the core standard in 3.2 turned up none. Testing showed glFramebufferTexture (which neither Mesamatrix nor [this](https://wikis.khronos.org/opengl/History_of_OpenGL#OpenGL_3.2_(2009)) mention) was a problem. Using glFramebufferTexture2D with a GL_TEXTURE_2D textarget seems to work just as well, difference being glFramebufferTexture2D does exist in OGL 3.1. Note that glFramebufferTexture**Layer** DOES exist in 3.1 (in fact, it's existed since 3.0), so those calls were fine as they were.

With this, MelonDS can use OpenGL 3.1 instead of 3.2; so Panfrost will now work.
This only affects the conventional OpenGL renderer; the compute renderer is, of course, a whole 'nother story.

Testing on radeonsi turned up no regressions (the conventional OpenGL renderer does show some artifacting in some games, but it's already there in master, I'll open an issue for that later).

### DIFF
--- a/src/GPU2D_OpenGL.cpp
+++ b/src/GPU2D_OpenGL.cpp
@@ -284,7 +284,7 @@ bool GLRenderer2D::Init()
             glTexImage2D(GL_TEXTURE_2D, 0, GL_RGBA, sz[0], sz[1], 0, GL_RGBA, GL_UNSIGNED_BYTE, nullptr);
 
             glBindFramebuffer(GL_FRAMEBUFFER, AllBGLayerFB[l]);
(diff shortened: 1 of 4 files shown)


Arisotura:
sure, why not!

we'll see what the CI says, what the team says, etc

PR #2655, opened 2026-05-09
Soul Link Milestone 2: add hardcoded TCP read endpoint

aryashah5000:
Adds SoulLinkServer, a QTcpServer that listens on localhost:8765. When a client connects, it reads 16 bytes from offset 0x02000000 (start of main RAM) and returns them as a hex line, then closes.

This proves the socket→MainRAM pipeline end-to-end with no game logic yet. Verify with: nc 127.0.0.1 8765 while a ROM is running.

The address is a placeholder; Milestone 3 will replace it with a generic WATCH/EVENT protocol.


Arisotura:
is this an accident?

aryashah5000:
Yeah sorry i meant to commit this to a fork

PR #2661, opened 2026-05-16
Improved filtering and antialiasing

### LINKED ISSUE #848: [Feature Request] Downsampling (filtering the higher resolutions to prevent aliasing)
As you can have higher resolutions with opengl, but the aliasing is still quite obvious i am thinking it has to do with how it's displayed.
I know on Ishiiruka (Dolphin) you could use a shader to simply filter it when downsampling to the window resolution,
instead of the basic nearest neighbour, and that helps a ton.

I might be mistaken though but i think it should help a lot and probably be fairly "simple" i hope.

Love the emulator, great job!

### LINKED ISSUE #1100: [Enhancement] Improved program window scaling via super-sampled integer nearest neighbor
The current "screen filtering" option seems to simply change the scaling algorithm from nearest neighbor to bilinear.

So my proposal is doing what I believe mGBA does for its "bilinear (pixelated)" scaling mode on the 3DS - doing integer nearest neighbor scaling to whatever would be the next largest ratio is that's bigger than the current window size, and then downscaling to the program window via bilinear.

For example, on a 1360x768 display with the DS screens arranged horizontally with the program window maximized, 1360px is 2.65625x that of the 512px of both DS screens side-by-side.  That is...not a particularly high ratio, meaning non-integer nearest neighbor looks pretty bad on anything that isn't an already blocky or pixelated game.

With my proposal, you'd instead scale that 512px to the next highest integer ratio which, in the case of a 1360px-wide display, would be 1536px wide - an exact 3x scale.  From there you would then downscale back to the display's 1360px width via a simple bilinear or the like.

(now to clarify this doesn't have to be exclusive to having the screens in a landscape orientation or even with the two screens at the same size; this was just simpler for me to explain my proposal using the situation I have found myself in)

Here's a visual example of what I mean, all scaled to 1360px wide:

**Nearest neighbor** (various diagonal slopes are ugly)
nearest_opt

**Bilinear** (blurry, 'nuff said)
bilinear_opt

**Super-sampled integer nearest neighbor**
supersampled_opt

KojoZero:
This PR adds more image filtering options and anti-aliasing settings such as FXAA and SMAA.
### Changes
- Added "Output Resampling" which has image filtering options such as: Nearest, Bilinear, Adaptive, Sharp Bilinear.
- Adaptive uses bilinear filtering when the internal resolution is below the output resolution and area sampling when it is above. With this change, anytime the internal resolution is higher than the output resolution, the image gets properly downscaled without introducing aliasing (it essentially acts like SSAA).
- Sharp Bilinear integer scales the image until it's greater than or equal to the output resolution using nearest neighbor filtering, and then downscales with bilinear filtering. The shader I use basically does this in a single pass. 
- All the filtering options have been gamma corrected
- Added Anti-Aliasing options such as FXAA and SMAA  

Note: Area sampling is paired with bilinear filtering because it's only meant for downscaling. I just used the name "Adaptive" because it uses different algorithms based on whether it's upscaling or downscaling, though I'm open to changing it to a better name or reworking it.

Screenshot 2026-05-15 182320
Screenshot 2026-05-15 182335

Closes #1100
Closes #848


nadiaholmquist:
Generally looks good to me. It seems like (on Linux with an AMD GPU) there's a high chance of things breaking when changing the display options, swapping the 3D renderer between software/GL seems to fix it though, but that's not ideal.

Capitalization of the includes in Screen.cpp need to be changed as they won't work on case sensitive file systems, and a bit of the formatting needs to be made in line with the rest of the project.

Felipefpl:
All those explanations of what the filters do will come as tooltips in the emulator? This must be done so the users will know what each of them will do.

72374:
I tested this with 2 games, and almost everything[^1] seems to work properly, but when downsampling with Adaptive, which should use area-sampling, it does look like Nearest, with some sampling-points offset by a pixel.

GPU: Radeon RX 6750 XT
GPU-Driver: Mesa 26.1.6

[^1]: SMAA, FXAA, Sharp Bilinear, Nearest, Bilinear, upscaling with Adaptive.

Issue #2555, opened 2026-01-28
blackmagic3 - FMV tearing


FireNX70:
Pokémon Black's opening FMV has some sort of tearing problem. This wasn't my display, V-Sync was enabled. This also happens with Ghost Trick's rewind FMV. Both OpenGL renderers are affected at any internal resolution. The software renderer's fine. Tested on radeonsi (Mesa 25.2.6) and nVidia's Windows drivers (561.09), using commit https://github.com/melonDS-emu/melonDS/pull/2553/commits/73859c836a8245f9e90fbe190339403f63c50c55. This was also unwittingly tested on Intel's horrible Windows drivers; version 23.20.16.4973, according to Windows' device manager.
If I didn't mess anything up when testing, the FMV bug already behaved the same way on commit https://github.com/melonDS-emu/melonDS/pull/2553/commits/a83df1f4eb11ad22c9945c4da79d51a21e820996.
Retested on radeonsi after commits https://github.com/melonDS-emu/melonDS/pull/2553/changes/99363f6ff93f4def5b5a1924c8c87e150f23de0d and https://github.com/melonDS-emu/melonDS/pull/2553/changes/654dd1a7045a96637792093a727f389cd92fc293, no change.

<img width="1228" height="941" alt="Image" src="https://github.com/user-attachments/assets/74b67d9f-295d-420f-a8a3-76d59ca3d8dd" />

<img width="1165" height="782" alt="Image" src="https://github.com/user-attachments/assets/cc6b1e62-c145-4b18-8c3f-a9ac6784da1d" />

<img width="1165" height="870" alt="Image" src="https://github.com/user-attachments/assets/da6d8d4c-6744-4c07-bfa4-ceac44c760c2" />

<img width="1165" height="870" alt="Image" src="https://github.com/user-attachments/assets/ffe3715c-00bb-4c29-9261-4696aea2fa97" />

KostaSaizo7:
Can reproduce this with other games FMVs too
tested on windows with amd gpu https://github.com/melonDS-emu/melonDS/pull/2553/changes/5a4594861ea25a32265f4b452ff54088a8c946f2

Pokemon White 2
<img width="960" height="726" alt="Image" src="https://github.com/user-attachments/assets/cf3945af-f7a5-492e-bb14-edb28d30dcb8" />

Inazuma Eleven
[...]

PR #2668, opened 2026-05-20
Fix tearing in hidden BG FMV playback

ZironZ:
This fixes FMV tearing in games that stream video data into an inactive BG layer, then later switch that BG on-screen.

The OpenGL renderer could prerender a hidden BG layer while the game was still updating it, leaving the cached BG texture with a mix of old and new frame data. This change still uploads VRAM changes normally, but defers rebuilding the BG texture until the layer is actually visible. That avoids building an FMV's BG texture from a half-updated video frame.

This partially addresses #2555. It affects several games that use Mobiclip FMVs. I tested it with Kingdom Hearts 358/2 Days and Ni no Kuni.

I also looked into Pokémon Black, but that seems to be a different case where it updates BG layers that are already visible. Fixing that safely would need a larger change, since games can also update visible BGs mid-frame for other effects.

### DIFF
--- a/src/GPU2D_OpenGL.cpp
+++ b/src/GPU2D_OpenGL.cpp
@@ -42,6 +42,7 @@ GLRenderer2D::GLRenderer2D(melonDS::GPU2D& gpu2D, GLRenderer& parent)
     : Renderer2D(gpu2D), Parent(parent)
 {
     ScaleFactor = 0;
+    DeferredLayerPrerenderDirty = 0;
 }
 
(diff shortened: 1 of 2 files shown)


KostaSaizo7:
I can confirm that this fixes the tearing in Ghost Trick, Inazuma Eleven and Kingdom Hearts

Arisotura:
thanks, I'll review this when I find some time

things been busy for me lately

PR #2675, opened 2026-05-27
Rich Presence support

ShiralynDev:
Adds rich presence support (RPC), current system adds support for discord but could be adapted for other social platforms like fluxer (when they get rich presence support that is). This feature is requested in #318. 
image

I'm not super happy with my code in EmuThread.cpp but I was hoping it could be discussed to lead to a better implementation that keeps it clean. 
The code also adds settings for rich presence both to respect the users privacy but also to allow for future customisation of RPC.
This does unfortunately add [discord-rpc](https://github.com/discord/discord-rpc) as a dependency. I added it in a similar way to blip-buf since it also looked like an external dependency to me. I did strip the discord-rpc down to avoid unnecessary bloat. It has also been slightly modified to make it compile. I went for this implementation instead of the new discord social sdk due to the MIT license discord-rpc uses aswell as the "chill vibe" of discord-rpc compared to the social sdk. 

In the case that this feature would be appreciated and merged, the ApplicationID in `src/RichPresence.h` should be changed to something that an admin has control over.
The current system is not a full implementation of what is possible. I believe that joining games and such could also be enabled trough the rich presence.


snowboyz0825:
You could probably make the icon for rich presence the rom's embedded icon, just setup melonDS and was curious about this and I find it hilarious this was posted 2 days ago

ShiralynDev:
> You could probably make the icon for rich presence the rom's embedded icon, just setup melonDS and was curious about this and I find it hilarious this was posted 2 days ago

I believe icons are set in the developer portal so it would have to be set by the person actually maintaining the discord RPC in the case of a merge.

PR #2681, opened 2026-06-04
wifi: don't auto power down the transceiver when power save is disabled

### LINKED ISSUE #2680: DS wifi silently dies ~67 seconds after init (dswifi / infra. mode)
Any homebrew using dswifi in infrastructure mode loses all wifi ~67s after `Wifi_InitDefault()`. The emulated transceiver is powered off and never comes back. The guest gets no error: `sendto()` keeps returning success, `Wifi_AssocStatus()` keeps reporting associated — RX/TX just stop. The same ROM runs indefinitely on real DS hardware.

Reproduced on melonDS 1.1 (macOS) and current master (c69c1ce). Easy to spot in the log (Debug level): a final `wifi: TRX power OFF` with no matching power-on.

## Repro

Attached: `wifi-death-repro.nds` (source below, BlocksDS). It connects via firmware wifi settings, sends a small UDP packet to the gateway every 100ms, and prints a status line every 5s — `rx` is `WSTAT_RXPACKETS`, which advances continuously (beacons) while the radio is alive.

1. Run the ROM with any working wifi setup (I used local-NAT/libslirp)
2. Watch it connect and the `rx` counter climb
3. ~67s after boot, `rx` freezes; 15s later the ROM prints `RADIO DEAD`

On real hardware (DS Lite, WEP AP) it runs forever.

<details>
<summary>Repro ROM source (BlocksDS, single file)</summary>

```c
#include <arpa/inet.h>
#include <netinet/in.h>
#include <stdio.h>
#include <sys/socket.h>

#include <dswifi9.h>
#include <nds.h>

#define SEND_INTERVAL_VBLS   6    // ~100ms @ 60 Hz
#define STATUS_INTERVAL_VBLS 300  // 5s
#define DEAD_AFTER_STALE     3    // rx frozen for 3 status intervals = dead
#define DEST_PORT            9999 // nothing needs to listen; TX is what matters

static volatile uint32_t vbl_cnt = 0;
static void on_vblank(void) { vbl_cnt++; }

int main(void)
{
    consoleDemoInit();
    irqSet(IRQ_VBLANK, on_vblank);
    irqEnable(IRQ_VBLANK);

    printf("wifi-death-repro\n");
    printf("on melonDS the radio dies ~67s\nafter boot; on real HW it runs\nforever\n\n");
    printf("Connecting to WiFi...\n");

    if (!Wifi_InitDefault(WFC_CONNECT)) {
        printf("\x1b[31mWiFi init failed!\x1b[0m\n");
        while (1) swiWaitForVBlank();
    }

    struct in_addr gw = {0}, mask = {0}, dns1 = {0}, dns2 = {0};
    struct in_addr ip = Wifi_GetIPInfo(&gw, &mask, &dns1, &dns2);
    printf("\x1b[32mconnected\x1b[0m ip=%s\n\n", inet_ntoa(ip));

    int sock = socket(AF_INET, SOCK_DGRAM, 0);
    if (sock < 0) {
        printf("\x1b[31msocket() failed\x1b[0m\n");
        while (1) swiWaitForVBlank();
    }

    struct sockaddr_in dst = {0};
    dst.sin_family = AF_INET;
    dst.sin_port   = htons(DEST_PORT);
    dst.sin_addr   = gw;

    const char payload[] = "ping";
    uint32_t sent_ok = 0, last_send = vbl_cnt, last_status = vbl_cnt;
    uint32_t last_rx = 0, stale = 0;
    bool announced = false;

    while (1) {
        swiWaitForVBlank();

        if (vbl_cnt - last_send >= SEND_INTERVAL_VBLS) {
            last_send = vbl_cnt;
            if (sendto(sock, payload, sizeof(payload), 0,
                       (struct sockaddr *)&dst, sizeof(dst)) >= 0)
                sent_ok++;
        }

        if (vbl_cnt - last_status >= STATUS_INTERVAL_VBLS) {
            last_status = vbl_cnt;
            uint32_t secs = vbl_cnt / 60;
            uint32_t rx = Wifi_GetStats(WSTAT_RXPACKETS);

            printf("t+%3lus ok=%-5lu rx=%lu\n",
                   (unsigned long)secs, (unsigned long)sent_ok, (unsigned long)rx);

            stale = (rx == last_rx) ? stale + 1 : 0;
            last_rx = rx;

            if (stale >= DEAD_AFTER_STALE && !announced) {
                announced = true;
                printf("\n\x1b[31mRADIO DEAD since ~t+%lus\x1b[0m\n",
                       (unsigned long)(secs - stale * (STATUS_INTERVAL_VBLS / 60)));
                printf("\x1b[31mrx frozen; sendto() still\x1b[0m\n");
                printf("\x1b[31mreturns success\x1b[0m\n\n");
            }
            if (announced && stale == 0) {
                announced = false;
                printf("\x1b[32mradio recovered?!\x1b[0m\n");
            }
        }
    }
}
```

Makefile is the BlocksDS `rom_arm9` template with `ARM7ELF := arm7_dswifi_maxmod.elf` and `LIBS := -ldswifi9 -lnds9`.

</details>

[wifi-death-repro.nds.zip](https://github.com/user-attachments/files/28602834/wifi-death-repro.nds.zip)

dannyow:
Fixes #2680

dswifi in infrastructure mode runs with power saving disabled (`W_MODE_WEP`sleep mode 2), but arms the one-shot `W_BeaconCount2` countdown at 0xFFFF TUs (≈67s) during init. `SetIRQ13()` exempted only sleep mode 3 from auto power-down, so when the countdown ran out the transceiver was powered off and since stations with power saving disabled don't service IRQ13/IRQ15, nothing ever woke it back up. All wifi traffic silently died ~67s after init.

This restricts the auto power-down to automatic power saving mode (0). Real hardware does not power down in mode 2 (verified with the repro ROM from the issue: dies at ~67s on melonDS, runs indefinitely on a real DS).

I wasn't sure whether mode 1 should also auto-sleep happy to adjust if the intended semantics are different.

Tested: repro ROM from #2680 on macOS, master build radio stays alive well past the 67s mark (>4 min soak); unpatched build dies on schedule.

### DIFF
--- a/src/Wifi.cpp
+++ b/src/Wifi.cpp
@@ -393,7 +393,13 @@ void Wifi::SetIRQ13()
 {
     SetIRQ(13);
 
-    if ((IOPORT(W_ModeWEP) & 0x7) != 3)
+    // auto power-down should only happen in automatic power saving mode (0).
+    // it must not happen when power saving is disabled (mode 2): stations
+    // running in that mode (eg. dswifi in infrastructure mode) don't service
+    // IRQ13/IRQ15, so a power-down triggered by the one-shot W_BeaconCount2
+    // countdown (armed at 0xFFFF = ~67s) would turn the transceiver off
+    // with nothing ever waking it up again
+    if ((IOPORT(W_ModeWEP) & 0x7) == 0)
     {
         if (!(IOPORT(W_PowerTX) & (1<<1)))
         {


AntonioND:
I've created an issue in BlocksDS to investigate switching to automatic power saving mode: https://codeberg.org/blocksds/sdk/issues/379

Arisotura:
oh, good catch

PR #2715, opened 2026-07-30
NocashPrint: Fix can't print from ITCM/DTCM

Prof9:
No$gba defines a printing mechanism using pseudo IO ports, which melonDS also supports. In No$gba, printing a string in TCM works fine, but in melonDS it doesn't (yields an unknown arm9 read warning).

This happens because melon effectively emulates the read as a bus read, calling NDS::ArmXRead8 which skips the TCM check (which is implemented in the ARM module). This PR changes it to a read via CPU, which defers to a bus read if the address is not in TCM.

This does have a side effect of updating DataCycles in the ARM instance, but I don't think this should be a problem here? It will also cause a data abort if the address being printed is not mapped , but I don't think that's a use case. Debug printing is a development feature anyway.

Reference: https://problemkaputt.de/gbatek.htm#dsdebugregistersemulatordevkits

### DIFF
--- a/src/NDS.cpp
+++ b/src/NDS.cpp
@@ -1555,14 +1555,14 @@ void NDS::NocashPrint(u32 ncpu, u32 addr, bool appendNewline)
     // addr: debug string
 
     ARM* cpu = ncpu ? (ARM*)&ARM7 : (ARM*)&ARM9;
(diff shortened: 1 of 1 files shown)


RSDuck:
we probably need a separate debug memory read/write function at some point for the GDB interface as well.

PR #2720, opened 2026-08-01
fix implementation of sqrt in freebios

bigfarts:
The sqrt SWI takes an unsigned integer, but subge/orrge perform signed comparisons. This means for any number >= 0x80000000, we perform incorrect arithmetic and the root gets forced to 0. This switches the instructions to perform unsigned comparisons.

An example of this bug can be observed in Rockman.EXE Operate Shooting Star, where if Rockman is on the furthest row from a Beetank, the Beetank's projectile will be aimed at itself rather than Rockman.

Before: 
image

After:
image
image

### DIFF
--- a/freebios/src/bios_common.s
+++ b/freebios/src/bios_common.s
@@ -541,9 +541,9 @@ swi_sqrt:
  0:
   orr root_check, square_index, root
   cmp remainder, root_check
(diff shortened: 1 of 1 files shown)


Jaklyy:
skimmed through the retail ntr arm9 bios and this looks like a correct fix

PR #2747, opened 2026-09-08
JIT: fix Thumb literal reloads and selective invalidation

Umberto-DEV:
A Thumb PC-relative load can return a rotated or stale value after memory changes. Separately, a write can discard unrelated cached blocks. This PR combines #2745 and #2746 with the same production changes and one consolidated explanation.

- Preserve the write mask in `InvalidateByAddr`, so only blocks overlapping the written 16-byte granule are invalidated.
- Align the second Thumb literal-folding lookup and declare PC as a source register for the desktop runtime fallback.

The alignment at [line 454](https://github.com/melonDS-emu/melonDS/blob/906e9ebb27da8c6a715cd7abab4abfe8a8d29427/src/ARMJIT_A64/ARMJIT_LoadStore.cpp#L451-L457) is correct, but it only covers the first lookup. Its local aligned address does not change `R15`. After that lookup rejects an invalidated literal at `0x02002640`, [`Comp_MemAccess` retries at `0x02002642`](https://github.com/melonDS-emu/melonDS/blob/906e9ebb27da8c6a715cd7abab4abfe8a8d29427/src/ARMJIT_A64/ARMJIT_LoadStore.cpp#L115-L120). This misses the invalidation entry and can fold `0x77885566` after a write of `0x55667788`, returning before runtime alignment. **The existing alignment alone does not fix this second path; this patch makes both attempts check the same address.**

Validation on macOS AArch64 and an Android 13 ARM64 AVD:

- Dedicated [invalidation tests](https://github.com/Umberto-DEV/melonDS/blob/8cc7cdb26880bfc0bfc4649de22a50d402137a99/jit-validation-20260907/invalidation-mask-desktop/README.md): baseline fails 25 of 37 cases; the mask fix passes all 37.
- The 26-case literal guest: folding-on failures fall from 22 to zero on the desktop core and from 12 to zero on the Android core with their corresponding Thumb fixes. Interpreter and folding-off controls pass.
- The full comparison covers 104 processes and 2,380 case executions. Five further runs with loads replaced by NOPs reject every case.

[Source analysis, reproduction commands and complete results](https://github.com/Umberto-DEV/melonDS/tree/6933a60719f3d9dacbc22c0c4a667930faeaf652/jit-pipeline-review-20260908).

Three separate failures remain outside these fixes: same-block store/load behavior on ARM9 and ARM7, and ARM9 Shared WRAM remapping. These tests do not establish an FPS improvement; local validation is separate from upstream CI.


Umberto-DEV (commented):
The final diff contains the two tested fixes. One avoids unnecessary recompilation of cached code; the other corrects a memory lookup that can still return the wrong value despite the existing alignment. Before/after results and reproduction steps are linked in the PR description.

Issue #1187, opened 2021-08-23
Maxmod software channels glitchy with JIT


nadiaholmquist:
In the "Giana 2001 Remix" track in the [maxmod demo](https://maxmod.devkitpro.org/demo/maxmod_ds.zip), occasional wrong notes of what I assume is a square wave sample from the module can be heard in the background. The easiest way to hear it is if you let the track play through fully and loop. From what I can tell this affects only the tracks that use Maxmod's software mixing.

Only seems to happen with branch and literal optimizations enabled, and a sufficiently large block size.

PR #2749, opened 2026-09-09
Fix SPU bit depth and stale JIT state

Umberto-DEV:
## Problem

The core can keep using settings or compiled code after they are no longer valid:

- SPU startup overwrites the selected audio bit depth.
- The JIT can reuse code for the wrong CPU, keep old memory links, discard unrelated code, or reload a Thumb value from the wrong address.
- A compiled block can hard-code a memory value, change that memory, and continue using the old value.
- Remapping Shared WRAM can leave compiled readers and lookup tables connected to the previous layout.

## Why the existing fixes are not enough

[PR #2745](https://github.com/melonDS-emu/melonDS/pull/2745) and [PR #2746](https://github.com/melonDS-emu/melonDS/pull/2746) cover only the unrelated-code and Thumb-address errors. [PR #2747](https://github.com/melonDS-emu/melonDS/pull/2747) only combines those two changes. All three are closed and unmerged, and none covers the other stale-state paths.

The safeguard in [`55ec724f`](https://github.com/melonDS-emu/melonDS/commit/55ec724fee5cdc5ed91226f3e4f1aad11a3bcfb2) only sees exact write addresses observed while compiling. It misses changing or conditional addresses, overlapping byte writes, and exchange instructions. The WVRAM fix in [`32609bbc`](https://github.com/melonDS-emu/melonDS/commit/32609bbc980d2a496b40f29c41ac2e39f7651230) only covers ARM7 video WRAM, not Shared WRAM, its ARM7 alias, or both CPU lookup tables.

## Fix

- Apply the existing audio bit-depth policy during SPU startup.
- Reuse compiled code only for the same CPU and remove all old memory links when mappings change.
- Select affected code using the address actually written and use the correct aligned address for Thumb reloads.
- Read values at runtime whenever a block may write memory; read-only blocks keep the existing optimization.
- Invalidate Shared WRAM users and clear both CPU lookup tables before changing the mapping.

JIT and no-JIT Release builds and all 104 focused test processes pass.

### DIFF
--- a/src/ARMJIT.cpp
+++ b/src/ARMJIT.cpp
@@ -941,7 +941,7 @@ void ARMJIT::InvalidateByAddr(u32 localAddr) noexcept
 
     AddressRange* region = CodeMemRegions[localAddr >> 27];
     AddressRange* range = &region[(localAddr & 0x7FFFFFF) / 512];
-    u32 mask = 1 << ((localAddr & 0x1FF) / 16);
+    u32 invalidationMask = 1 << ((localAddr & 0x1FF) / 16);
 
     range->Code = 0;
     for (int i = 0; i < range->Blocks.Length;)
@@ -955,7 +955,7 @@ void ARMJIT::InvalidateByAddr(u32 localAddr) noexcept
             if (block->AddressRanges()[j] == (localAddr & ~0x1FF))
             {
                 mask = block->AddressMasks()[j];
-                invalidated = block->AddressMasks()[j] & mask;
+                invalidated = mask & invalidationMask;
                 assert(mask);
                 break;
             }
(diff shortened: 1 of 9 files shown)


Umberto-DEV:
The day-to-day effect is better correctness in a few specific cases. The selected DS audio bit depth now takes effect at startup, and the JIT avoids using old values after memory changes or Shared WRAM is remapped. Most games should behave as before. The available evidence does not prove that #1187 or any specific commercial game is fixed.

Compatibility: there are no new settings or file-format changes. Both JIT and non-JIT builds were verified, including DS and DSi behavior.

Performance: no FPS improvement has been measured. More selective invalidation preserves unrelated compiled code and avoids unnecessary recompilation. The tradeoff is that literal folding is disabled in compiled blocks that may write memory, so those blocks may be slightly slower because they read values at runtime. No net performance claim is intended.

Umberto-DEV:
Closing this PR: the fork is being consolidated on a single development branch, and these changes will be resubmitted later in a consolidated form. Sorry for the noise.

PR #2751, opened 2026-09-11
Three small fixes: MAC address, redundant shader rebuild, unused edge list

Umberto-DEV:
Three unrelated fixes, one commit each.

### 1. Only two of the six bytes of the MAC address are written — `src/SPI.cpp`

When a DSi game boots directly, melonDS writes the address to the same spot three times instead of moving along. Four of the six bytes never arrive, so the emulated console ends up with an address its firmware never gave it.

*What changes for you:* fixes a bug. I don't know of a game that misbehaves because of it, and I'm not claiming there is one.

### 2. Changing a video setting rebuilds every shader, even when nothing changed — `src/GPU3D_Compute.cpp`

Every time the renderer is handed its settings it throws away all its shaders and rebuilds them, even when the values are the ones it is already using. The OpenGL renderer already checks for this; the compute one doesn't.

*What changes for you:* faster. On a test device, changing video settings rebuilt 33 shaders ten times in one session. After the fix: none. It still rebuilds on startup, where it has to.

### 3. The renderer prepares data for a drawing step that is switched off — `src/GPU3D_OpenGL.cpp`

The edge outlines step has been commented out for years, but the list of edges it would need is still built for every polygon and sent to the graphics card every frame. This doesn't turn the step back on — it just stops paying for it while it's off.

*What changes for you:* removes work that has no effect. I haven't measured how much.

### What this doesn't prove

Fixes 1 and 3 aren't backed by a measurement — I can show the bug in the code, not its effect on screen. Fix 2 was measured on Android only.

### DIFF
--- a/src/GPU3D_Compute.cpp
+++ b/src/GPU3D_Compute.cpp
@@ -324,6 +324,13 @@ void ComputeRenderer3D::SetRenderSettings(int scale, bool highResolutionCoordina
 {
     u8 TileScale;
 
+    // Nothing below depends on anything but these two, and rebuilding means
+    // recompiling every shader and reallocating the buffers. The GL renderer
+    // already bails out the same way. ScaleFactor starts at 0, which is not a
+    // valid scale, so the first call always goes through.
+    if (scale == ScaleFactor && highResolutionCoordinates == HiresCoordinates)
+        return;
+
     if (ScaleFactor != -1)
     {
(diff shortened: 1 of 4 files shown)


Umberto-DEV:
Closing this PR: the fork is being consolidated on a single development branch, and these changes will be resubmitted later in a consolidated form. Sorry for the noise.

PR #2755, opened 2026-09-15
Learn with Pokemon: Typing Adventure keyboard implementation

Davey-Hughes:
This PR implements keyboard input for the game "Learn with Pokemon: Typing Adventure". There were a few things that had to be reverse engineered to get this to work, as well as making a menu to support remapping keys.

image
image

A few things to note about the input settings:
- on a fresh save the game asks you to type the `Fn` key on the keyboard to ensure it's connected. I think it's unlikely this is actually used in the game though I haven't played the game extensively (I needed to build this code to try it out), so one of the options is to just automatically send `Fn` on a fresh game start/save. You can disable this and map a custom `Fn` key to your keyboard, however
- the experimental dpad input for the arrow keys is because the game actually asks you to use the arrow keys on the keyboard to use some stuff in the menus, however the game doesn't actually seem to implement this in any way and you need to use the touch screen or DS dpad to perform the action. From my testing this experimental enhancement works, but it's good to make it an option in case it collides somewhere else I think.

#### The Bluetooth emulation (the reverse-engineered part)

Before this branch: upstream master's CartRetailBT was a stub whose SPITransmitReceive returned 0. The cart holds a Broadcom BCM2070 Bluetooth chip on the same SPI bus as the save chip.

The layers, from the bottom up:
1. SPI framing (src/NDSCart/CartRetailBT.cpp:238)
   - The first byte of each session sets its type: 0x01 = game sends, 0xFF = sync byte, 0x02 = game reads.
   - A frame is type | length (big-endian) | HCI packet.
   - The cart signals "I have a reply" by raising IREQ_MC.
   - It shares the bus with the save chip, so ClassifySPISession works out where each session goes.
2. HCI (src/NDSCart/BTKeyboard.cpp)
   - This is a scripted controller, not a real Bluetooth stack: about 30 commands, no radio, no link-timing emulation.
   - Values the game insists on, all measured from traces:
     - Device class 0x000540. The game's inquiry filter drops anything else.
     - The authentication replies are 0x040B/0x040C/0x040D/0x040E. Swapping 0x040B and 0x040C silently breaks registration.
     - The game supplies the PIN itself (8574 in the European build, 7068 in the Japanese one), and we accept whatever it sends.
     - Read_RSSI gets polled about every 11 s.
3. L2CAP + SDP: channel setup plus one SDP record. That record declares HIDReconnectInitiate, which matters for reconnecting (below).
4. HIDP
   - Input reports need report ID 0x01: A1 01 <8 bytes>.
   - Battery is a GET_REPORT for ID 0x03, answered A1 03 FF (full). Any other answer shows empty batteries.

#### Game saving
The game never uses the cart's save chip. It stores its save in the Bluetooth controller's own flash, through Broadcom vendor commands:
- 0xFC4D: read
- 0xFC4C: write
- 0xFF5E: erase a 4 KB sector

CartRetailBT maps controller address 0xFF008000 onto the normal 64 KB .sav file. The game recognises its save by the signature RAXT.

#### Key injection
On real hardware: a HID report arrives → the game's handler at 0x0205BBBC (European build) → it decodes the key through the game's own tables → it adds one event to a 16-slot ring buffer in RAM → the game reads one event per frame.

melonDS skips straight to the last step. PokeTypeKeyboard::PushKeyID (src/PokeTypeKeyboard.cpp:276):

1. Safety check: IsReady() compares the bytes of the game's input-fetch routine against a stored signature. An unknown build fails closed instead of corrupting memory.
2. Read the indexes: the read index is at queue +0x40, the write index at +0x44. A full queue drops the key, as the game's own code does.
3. Write the event in the game's own format: (mods << 24) | (HID usage << 16) | char. The Japanese build stores the bare character.
4. Look up addresses per build. There are six builds (UZPP/F/D/I/S/J). For example UZPP has the queue at 0x020C88B8 and the driver state at 0x020C87D0.

Choosing the character happens in the frontend (src/frontend/qt_sdl/EmuInstanceInput.cpp:360):
- Three modes:
  - Layout: your host keyboard layout decides the character.
  - Overrides: the same, but bound keys win.
  - Positional: bindings only.
- The layout tables were pulled from the game ROM by tools/poketype/extract_layout.py. CharForKey copies the game's own decoder.
- Keys are queued under a mutex on the UI thread and drained on the emulator thread.
- Fn is special: until the game has connected, it makes the emulated keyboard discoverable.
- Arrows → D-pad is opt-in. The game's menus only read the D-pad, and they ignore keyboard arrows whether they arrive by injection or by HID.

It is possible to do typing entirely over HID. The way I built this was the injection method first, and it seems to work very well, so while I did some preliminary testing of typing over HID it's not built out and I don't really think it's needed unless some other bugs come up.


JesseTG:
It looks like this just injects the emulated keyboard's keypresses directly into the game's RAM?

Davey-Hughes:
> It looks like this just injects the emulated keyboard's keypresses directly into the game's RAM?

The typing yes. This was the fastest way to get some initial progress before any of the other HID or Bluetooth reverse engineering and it worked well so I just kept it.

